PR #3, opened 2021-08-20
Simple IPC between DJL serving and Python server

sindhuvahinis:
Simple communication between DJL serving and python server with byte array. 
Should manually start the python in order to test it.


sindhuvahinis:
I did not include the domain socket changes yet. Will add those changes in another commit or next PR.

PR #4, opened 2021-08-22
UDS changes 

sindhuvahinis:
Changes to include UDS as well while connecting to python server.


serving/src/main/java/ai/djl/serving/pyclient/PythonConnector.java
```diff
@@ -13,53 +13,55 @@
...
 
 /** This class creates a netty client. */
-public class SocketConnector {
+public class PythonConnector {
 
-    private static final Logger logger = LoggerFactory.getLogger(SocketConnector.class);
-    private static final SocketConnector SOCKET_CONNECTOR = newInstance();
+    private static final Logger logger = LoggerFactory.getLogger(PythonConnector.class);
+    private static final PythonConnector PYTHON_CONNECTOR = newInstance();
+    private static final String DEFAULT_SOCKET_PATH = "/tmp/uds_sock";
```
sindhuvahinis, line 34: Is this okay, to hardcode the default socket path or shall we get it from config?

serving/src/main/java/ai/djl/serving/pyclient/PythonConnector.java
```diff
@@ -81,7 +83,18 @@ public Channel getChannel() {
...
+
+    private boolean isUDS() {
+        ConfigManager configManager = ConfigManager.getInstance();
+        return configManager.useNativeIo();
+    }
+
+    private SocketAddress getSocketAddress() {
+        return isUDS()
+                ? new DomainSocketAddress(DEFAULT_SOCKET_PATH)
+                : InetSocketAddress.createUnresolved("127.0.0.1", 9000);
```
sindhuvahinis, line 98: Is it okay to have these private methods? I think these methods can be removed, if we include to ConnectorType as Python.

PR #6, opened 2021-08-24
Input and response changes 

sindhuvahinis:
Design:
1. User's preprocess function is expected to take an input object as func param and return an list of numpy array 
2. User's postprocess function is expected to take an list of numpy array and return list of numpy array
3. Postprocess is considered optional

Changes in this PR:
1. Added example pre_processing.py and post_processing.py files in resources for testing
2. serializing input and sending it to python server.
3. changes to include postprocess as well and test cases. 

In the next PR:
1. start python server from DJL
2. multiple python workers


serving/src/test/java/ai/djl/serving/PythonTranslatorTest.java
```diff
@@ -84,12 +79,22 @@ public void setup() throws ModelException, IOException, ParseException {
                 Files.copy(is, paramFile, StandardCopyOption.REPLACE_EXISTING);
             }
 
-            Path libsDir = modelDir.resolve("bin");
+            Path libsDir = modelDir.resolve("libs");
```
sindhuvahinis, line 82: Code to include an empty lib for now, for testing. 

Because [ServingTranslatorFactory](https://github.com/deepjavalibrary/djl/blob/master/api/src/main/java/ai/djl/translate/ServingTranslatorFactory.java#L79), if lib/libs folder is not found, then RawTranslator is loaded.

serving/src/main/python/protocol/input.py
```diff
@@ -0,0 +1,94 @@
+import numpy as np
```
frankfliu, line 1: Add license header

serving/src/main/python/protocol/response.py
```diff
@@ -0,0 +1,16 @@
+class Response(object):
```
frankfliu, line 1: License header

serving/src/main/python/protocol/input.py
sindhuvahinis: Okay

frankfliu (approved):

PR #15, opened 2021-10-01
[python] remove type hint to be compatible with lower version of python

frankfliu:
## Description ##


zachgk:
It may work if you add `from __future__ import annotations` - https://stackoverflow.com/questions/63939138/is-there-a-way-to-use-python-3-9-type-hinting-in-its-previous-versions

PR #45, opened 2021-12-01
[serving] Install engine dependency ondemand

frankfliu:
## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
+++ b/serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
@@ -1,0 +1,235 @@
+/*
+ * Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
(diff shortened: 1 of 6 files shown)


serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
```diff
@@ -0,0 +1,235 @@
...
+                Utils.moveQuietly(tmp, file);
+            } finally {
+                Files.deleteIfExists(tmp);
+            }
+        }
+        MutableClassLoader mcl = MutableClassLoader.getInstance();
+        mcl.addURL(file.toUri().toURL());
+    }
+
+    private static String resolveDjlVersion() {
```
zachgk, line 153: This queries for the latest version, but don't we need the version that is currently being used?

frankfliu: The active DJL version is SNAPSHOT version, which we cannot be downloaded from maven central. To download SNAPSHOT version is complicated. The logic here is a workaround for development.

zachgk (approved):

PR #52, opened 2022-01-14
Workflow file parsing

zachgk:
This adds support for parsing a workflow from a json or yaml file.

In addition, it refactors the workflow to support the WorkflowFunctions. This
includes creating the executor, helping constructs, and supporting model
execution as a WorkflowFunction.

It also makes the WorkloadManager autoCloseable for easier testing.


frankfliu (approved):
Looks good to me in general.

I will take a look closely for the syntax next week.

PR #57, opened 2022-01-21
Load all model by default when model store is specified

frankfliu:
## Description ##


serving/src/main/java/ai/djl/serving/ModelServer.java
```diff
@@ -265,14 +265,17 @@ private void initModelStore() throws IOException {
         Set<String> startupModels = ModelManager.getInstance().getStartupModels();
 
         String loadModels = configManager.getLoadModels();
+        Path modelStore = configManager.getModelStore();
         if (loadModels == null || loadModels.isEmpty()) {
-            return;
+            if (modelStore != null) {
```
zachgk, line 270: There should be a way to disable this too. Maybe null for "all" and empty string for loading no models

zachgk (approved):

PR #59, opened 2022-01-28
Multi-device model support

zachgk:
This modifies server to support models with workers on multiple devices rather
than using the previous system of having it as multiple models with versions
indicating the device.

As part of this, it also tries to replace ad-hoc device usages with actual
device. This includes systematically converting device names into Device and
storing device rather than device name or gpu boolean.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ systemProp.org.gradle.internal.http.connectionTimeout=60000
 # FIXME: Workaround gradle publish issue: https://github.com/gradle/gradle/issues/11308
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
-djl_version=0.15.0
+djl_version=0.16.0
 commons_cli_version=1.4
 netty_version=4.1.72.Final
 slf4j_version=1.7.32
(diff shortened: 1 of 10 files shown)


serving/src/main/java/ai/djl/serving/util/DeviceName.java
```diff
@@ -0,0 +1,52 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.serving.util;
+
+import ai.djl.Device;
+
+/** Utility for handling device name strings. */
+public final class DeviceName {
```
frankfliu, line 18: Seems more fit better in Device class:

```
public static Device fromDeviceName(String deviceName)

```

frankfliu (approved):

zachgk:
This PR depends on DJL `model.newPredictor(device)`. However, that function works correctly with DJL defined models, but not imported models (SymbolBlocks). For now, I am marking this PR as draft until the dependency is fixed.

PR #66, opened 2022-03-03
error message when model path is invalid

sindhuvahinis:
Throws error when model directory path is invalid.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/ModelServer.java
+++ b/serving/src/main/java/ai/djl/serving/ModelServer.java
@@ -416,7 +416,7 @@ public class ModelServer {
             Repository repository = Repository.newInstance("modelStore", modelUrl);
             List<MRL> mrls = repository.getResources();
             if (mrls.isEmpty()) {
-                throw new AssertionError("Model directory path does not exist : " + modelUrl);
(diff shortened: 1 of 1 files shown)


serving/src/main/java/ai/djl/serving/ModelServer.java
```diff
@@ -415,6 +415,10 @@ private String inferEngineFromUrl(String modelUrl) {
         try {
             Repository repository = Repository.newInstance("modelStore", modelUrl);
             List<MRL> mrls = repository.getResources();
+            if (mrls.isEmpty()) {
+                throw new AssertionError("Model directory path does not exist : " + modelUrl);
```
frankfliu, line 419:
```suggestion
                throw new IllegalArgumentException("Invalid model url: " + modelUrl);
```

frankfliu: AssertionError should only be used for the case that it should never happen.

sindhuvahinis: I see. Okay!

frankfliu (approved):

PR #75, opened 2022-03-24
Create control plane for workflows

zachgk:
This updates the management request handler to add a control plane for
workflows. It shares a number of routes between the old models and workflows,
while some routes have different behaviors for each.

Specifically, there is a separate register and list command for both models and
workflows, while all other commands including describe, unregister, and scale
only work on workflows but will still accept "/models/..." routes. The precise
behavior is documented in management_api.md.

There are also a few refactors for ModelServerTest including cleaning up test
client connections, a shared list reader, a shared model registration checking function.

Finally, it also fixes a bug with the WorkLoadManager close not closing the
threadpool. This can result in the tests hanging.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/http/ListWorkflowsResponse.java
+++ b/serving/src/main/java/ai/djl/serving/http/ListWorkflowsResponse.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
  * with the License. A copy of the License is located at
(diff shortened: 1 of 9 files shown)


serving/src/main/java/ai/djl/serving/http/ListWorkflowsResponse.java
```diff
@@ -0,0 +1,104 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
frankfliu, line 2:
```suggestion
 * Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

serving/src/main/java/ai/djl/serving/http/ManagementRequestHandler.java
```diff
@@ -64,7 +69,8 @@
     /** HTTP Parameter "min_worker". */
     private static final String MIN_WORKER_PARAMETER = "min_worker";
 
-    private static final Pattern PATTERN = Pattern.compile("^/models([/?].*)?");
+    private static final Pattern PATTERN = Pattern.compile("^/(models|workflows)([/?].*)?");
```
frankfliu, line 72: better call it WORKFLOW_PATTERN

frankfliu (approved):

PR #101, opened 2022-06-02
add Java engine for Java client purpose

lanking520:
## Description ##

Create a Java engine (without model) to support Java client on data conversion from/to NDList


settings.gradle
```diff
@@ -3,5 +3,6 @@ include ':central'
 include ':plugins:plugin-management-plugin'
 include ':plugins:static-file-plugin'
 include ':engines:python'
+include 'engines:java'
```
frankfliu, line 6:
```suggestion
include ':engines:java'
```

engines/java/src/main/java/ai/djl/java/engine/JavaEngine.java
```diff
@@ -0,0 +1,112 @@
...
+import ai.djl.training.GradientCollector;
+
+/**
+ * The {@code JavaEngine} is an implementation of the {@link Engine} that runs Java applications.
+ */
+public class JavaEngine extends Engine {
+
+    public static final String ENGINE_NAME = "Java";
+    static final int RANK = 10;
+    static final String VERSION = "0.18.0";
```
frankfliu, line 29: If we want to match DJL version, we return:
```
Engine.class.getPackage().getSpecificationVersion()
```

frankfliu (approved):

PR #106, opened 2022-06-09
add DeepSpeed docker image

lanking520:
## Description ##

Add DeepSpeed docker image for large model supporting

### DIFF
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -1,0 +1,39 @@
+FROM nvidia/cuda:11.3.0-cudnn8-devel-ubuntu20.04
+ARG djl_version=0.17.0
+ARG torch_version=1.11.0
+ARG deepspeed_version=0.6.5
+ARG transformers_version=4.19.2
+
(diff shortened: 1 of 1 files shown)


serving/docker/deepspeed.Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM nvidia/cuda:11.3.0-cudnn8-devel-ubuntu20.04
```
frankfliu, line 1: Let name it cu113.Dockerfile, this docker can be a generic GPU docker

frankfliu: And the base docker image should be parameterized as well

lanking520: This is specialized for pytorch, transformers and deepspeed. For GPU generic docker, we should not install anything

frankfliu, line 1:
```suggestion
ARG version=11.3.0-cudnn8-devel-ubuntu20.04
FROM nvidia/cuda:$version
```

PR #111, opened 2022-06-15
[Action] add two new instances

lanking520:
## Description ##

Add two new instances for CI on GitHub Actions. Graviton and Inferentia

### DIFF
--- a/.github/workflows/nightly-publish.yml
+++ b/.github/workflows/nightly-publish.yml
@@ -16,10 +16,30 @@ jobs:
           --fail \
           | jq '.token' | tr -d '"' )
           ./start_instance.sh action_gpu $token
+      - name: Create new Graviton instance
+        id: create_aarch64
+        run: |
+          cd /home/ubuntu/djl_benchmark_script/scripts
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-publish.yml
```diff
@@ -16,26 +16,64 @@ jobs:
           --fail \
           | jq '.token' | tr -d '"' )
           ./start_instance.sh action_gpu $token
+      - name: Create new Graviton instance
+        id: create_arm64
```
frankfliu, line 20: Can we use aarch64 instead?

.github/workflows/nightly-publish.yml
```diff
@@ -16,26 +16,64 @@ jobs:
...
+  execute-gpu-runner:
     runs-on: [self-hosted, gpu]
     needs: create-runners
     steps:
       - name: Test GPU running
         run: |
           nvidia-smi
 
+  execute-arm64-runner:
+    runs-on: [ self-hosted, arm64 ]
```
frankfliu, line 51: Here are another 3 place using arm64

PR #114, opened 2022-06-17
[Docker] fix download issue

lanking520:
## Description ##

fix ARM build download issue


zachgk (approved):

serving/docker/aarch64.Dockerfile
```diff
@@ -27,12 +27,12 @@ CMD ["serve"]
 
 COPY scripts scripts/
 RUN mkdir -p /opt/djl/conf
```
frankfliu, line 29: did you removed:
`ENV MODEL_SERVER_HOME=/opt/djl` ?

lanking520: It is in line 23

PR #122, opened 2022-06-21
[docker] Reduce docker layer to reduce size

frankfliu:
## Description ##

### DIFF
--- a/serving/docker/Dockerfile
+++ b/serving/docker/Dockerfile
@@ -1,18 +1,18 @@
 FROM ubuntu:20.04
 ARG djl_version=0.17.0
 
-COPY scripts scripts/
 RUN mkdir -p /opt/djl/conf
+COPY scripts scripts/
 COPY config.properties /opt/djl/conf/
-RUN scripts/install_djl_serving.sh $djl_version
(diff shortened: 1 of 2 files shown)


serving/docker/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
 COPY config.properties /opt/djl/conf/
-RUN scripts/install_djl_serving.sh $djl_version
-RUN rm -rf scripts
-
 COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
-RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
 
-EXPOSE 8080
+RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh && \
+    scripts/install_djl_serving.sh $djl_version && \
```
lanking520, line 10: How do we run the scripts without copying them to docker container first?

lanking520 (changes requested):

serving/docker/Dockerfile
frankfliu: the script folder is copied on line 5

PR #130, opened 2022-06-22
ModelInfo from Criteria

zachgk:
Adds a new constructor for ModelInfo based on a Criteria. This can be used when
using the WLM as a standalone tool to create jobs to run.

### DIFF
--- a/wlm/src/test/java/ai/djl/serving/wlm/ModelInfoTest.java
+++ b/wlm/src/test/java/ai/djl/serving/wlm/ModelInfoTest.java
@@ -12,17 +12,55 @@
  */
 package ai.djl.serving.wlm;
 
+import ai.djl.Device;
+import ai.djl.ModelException;
+import ai.djl.inference.Predictor;
+import ai.djl.modality.Input;
+import ai.djl.modality.Output;
+import ai.djl.repository.zoo.Criteria;
+import ai.djl.repository.zoo.ZooModel;
+import ai.djl.translate.TranslateException;
+import ai.djl.util.Utils;
+
 import org.testng.Assert;
 import org.testng.annotations.Test;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
 
 public class ModelInfoTest {
 
     @Test
     public void testQueueSizeIsSet() {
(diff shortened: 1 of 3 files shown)


zachgk:
@frankfliu @lanking520

wlm/src/test/java/ai/djl/serving/wlm/ModelInfoTest.java
```diff
@@ -12,17 +12,55 @@
...
+
+    @Test
+    public void testCriteriaModelInfo() throws ModelException, IOException, TranslateException {
+        String modelUrl = "djl://ai.djl.zoo/mlp/0.0.3/mlp";
+        Criteria<Input, Output> criteria =
+                Criteria.builder()
+                        .setTypes(Input.class, Output.class)
+                        .optModelUrls(modelUrl)
+                        .build();
+        try (ModelInfo modelInfo = new ModelInfo(criteria)) {
```
lanking520, line 52: I am getting confused why we need to introduce an extra class to load the model on different device. Previously we made Criteria class to be read only. But I think we can simply allow user to change a criteria maybe like:

```
criteria.toBuilder()
```

or 

```
Criteria<Input, Output> newCriteria = Criteria.builder(oldCriteria)
```
to achieve the same goal

frankfliu (approved):
either criteria.toBuilder() or Criteria.build(Criteria criteria) works for me.

wlm/src/test/java/ai/djl/serving/wlm/ModelInfoTest.java
zachgk: We have been using the ModelInfo for a few things including storing model configuration (queue size, batch size, etc.) and as a spawner that can spawn multiple workers. I actually did create the `criteria.toBuilder()` in a previous PR to this. This PR just enhances the ModelInfo to take input as a Criteria in addition to a modelUrl. 

If we want to remove the ModelInfo, it would be a separate process in a different PR. We would have to fix model.newPredictor(device) to use for worker spawning, migrate out model configuration, and possibly a few other  changes. For now, I think this change is consistent with the current pattern.

PR #149, opened 2022-06-29
add change docker option

lanking520:
## Description ##

Change docker permission to avoid:

```
Got permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/version": dial unix /var/run/docker.sock: connect: permission denied
```

### DIFF
--- a/serving/docker/scripts/install_docker.sh
+++ b/serving/docker/scripts/install_docker.sh
@@ -7,4 +7,5 @@ fi
 
 curl -fsSL https://get.docker.com -o get-docker.sh
 sudo sh get-docker.sh
-sudo chmod 666 /var/run/docker.sock
+# grant docker permission
+sudo usermod -a -G docker ubuntu


serving/docker/scripts/install_docker.sh
```diff
@@ -7,4 +7,5 @@ fi
 
 curl -fsSL https://get.docker.com -o get-docker.sh
 sudo sh get-docker.sh
-sudo chmod 666 /var/run/docker.sock
+# grant docker permission
+sudo usermod -a -G docker ubuntu
```
frankfliu, line 11: this wont' work, you need restart ssh session to activate group permission

lanking520: This will be helpful for AARCH JNI build. All future usage of docker would work

PR #162, opened 2022-07-08
[docker] numpy is required by djl python engine by default

frankfliu:
## Description ##


serving/docker/scripts/install_python.sh
```diff
@@ -6,4 +6,5 @@ set -e
 apt-get update
 DEBIAN_FRONTEND=noninteractive apt-get install -yq python3-dev python3-pip
 python3 -m pip --no-cache-dir install -U pip
+python3 -m pip --no-cache-dir install -U numpy
```
lanking520, line 9: Which part do we use numpy?

lanking520 (approved):

PR #167, opened 2022-07-12
Fix OnnxRuntime Benchmark on GPU

zachgk:
This fixes the gradle detection of hardware to enable onnxruntime to run on GPU
that was lost in the move of djl-bench from DJL to DJL Serving.

### DIFF
--- a/benchmark/build.gradle
+++ b/benchmark/build.gradle
@@ -14,7 +14,6 @@ dependencies {
     runtimeOnly "ai.djl.tensorflow:tensorflow-model-zoo"
     runtimeOnly "ai.djl.mxnet:mxnet-model-zoo"
     runtimeOnly "ai.djl.paddlepaddle:paddlepaddle-model-zoo"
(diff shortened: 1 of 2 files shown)


benchmark/build.gradle
```diff
@@ -23,6 +22,23 @@ dependencies {
...
+        Process process = pb.start()
+        hasGPU = process.waitFor() == 0
+    } catch (IOException ignore) {
+    }
+
+    if (hasGPU) {
+        runtimeOnly("ai.djl.onnxruntime:onnxruntime-engine") {
+            exclude group: "com.microsoft.onnxruntime", module: "onnxruntime"
+        }
+        runtimeOnly "com.microsoft.onnxruntime:onnxruntime_gpu:${onnxruntime_version}"
```
frankfliu, line 38: I don't think we have onnxruntime_version defined in gradle.properties in this repo

frankfliu (approved):

PR #170, opened 2022-07-14
bump to latest djl snapshot version

siddvenk:
## Description ##

bump djl version so nightly image builds use latest snapshot. 

It's expected that some of the checks fail due to not having the 0.19.0-SNAPSHOT serving artifacts in maven. We need to run the "serving publish" action from djl after this has been merged, and then run the actions in this repo to verify. 

Alternatively I could push this change to a different branch, modify the workflow in djl on a separate branch to checkout the corresponding serving branch and run the action that way to validate before merging this to master.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,7 +7,7 @@ systemProp.org.gradle.internal.http.connectionTimeout=60000
 # FIXME: Workaround gradle publish issue: https://github.com/gradle/gradle/issues/11308
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
-djl_version=0.18.0
+djl_version=0.19.0
 commons_cli_version=1.5.0
 netty_version=4.1.79.Final
 slf4j_version=1.7.36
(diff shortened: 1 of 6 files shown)


serving/docker/Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN scripts/install_python.sh && \
     djl-serving -i ai.djl.mxnet:mxnet-native-mkl:1.9.0:linux-x86_64 && \
     djl-serving -i ai.djl.pytorch:pytorch-native-cpu:1.11.0:linux-x86_64 && \
     djl-serving -i ai.djl.tensorflow:tensorflow-native-cpu:2.7.0:linux-x86_64 && \
-    djl-serving -i ai.djl.onnxruntime:onnxruntime-engine:0.18.0 && \
+    djl-serving -i ai.djl.onnxruntime:onnxruntime-engine:$djl_version && \
```
frankfliu, line 48: "djl-serving -i" doesn't work for SNAPSHOT version

frankfliu (approved):

gradle.properties
```diff
@@ -7,7 +7,7 @@ systemProp.org.gradle.internal.http.connectionTimeout=60000
 # FIXME: Workaround gradle publish issue: https://github.com/gradle/gradle/issues/11308
 systemProp.org.gradle.internal.publish.checksums.insecure=true
 
-djl_version=0.18.0
+djl_version=0.19.0
```
lanking520, line 10: need to go with ~SNAPSHOT as well

frankfliu: No, we add SNAPSHOT in build.gradle

PR #177, opened 2022-07-26
[KServe] KServe v2 Model metadata API

sindhuvahinis:
## Description ##

https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#model-metadata

`GET v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]`

```
$metadata_model_response =
{
  "name" : $string,
  "versions" : [ $string, ... ] #optional,
  "platform" : $string,
  "inputs" : [ $metadata_tensor, ... ],
  "outputs" : [ $metadata_tensor, ... ]
}
```


frankfliu:
We should have a separate KServeRequestHandler, this should be registered before InvalidRequestHandler

sindhuvahinis:
> We should have a separate KServeRequestHandler, this should be registered before InvalidRequestHandler

OKay, Sure. Changed it.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        if (segments.length > 4) {
+            modelVersion = segments[5];
+        }
+
+        ModelManager modelManager = ModelManager.getInstance();
+        Map<String, Endpoint> endpoints = modelManager.getEndpoints();
+
+        Endpoint endpoint = endpoints.get(modelName);
+        if (endpoint == null) {
+            sendModelNotFoundError(modelName, modelVersion, ctx);
```
frankfliu, line 89: Just throw ModelNotFoundException

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            return;
+        }
+
+        List<Workflow> workflows = endpoint.getWorkflows();
+        if (workflows.isEmpty()) {
+            sendModelNotFoundError(modelName, modelVersion, ctx);
```
frankfliu, line 95: Throw new 
```suggestion
            throw new ModelNotFoundException("Model not found for the given model: "
                        + modelName
                        + " and model version "
                        + modelVersion);
```

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        KServeDescribeModelResponse response = new KServeDescribeModelResponse();
+        response.setVersions(versions);
+        response.setPlatformForEngineName(modelInfo.getEngineName());
+
+        ZooModel<Input, Output> model = modelInfo.getModel(modelInfo.withDefaultDevice(null));
+        response.setName(model.getName());
+        DataType dataType = model.getDataType();
+
+        if (model.describeInput() == null || model.describeOutput() == null) {
+            String errorMsg =
```
frankfliu, line 121: Is input/output shape required for KServe?

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+                    "Input/Output shapes are unknown, "
+                            + "please run predict or forward once and call describe model again";
+
+            Map<String, String> content = new ConcurrentHashMap<>();
+            content.put("error", errorMsg);
+
+            NettyUtils.sendJsonResponse(ctx, content, HttpResponseStatus.NOT_FOUND);
```
frankfliu, line 128: What error code should we return if Input/Ouput is required? Not all the model has that information in DJL?

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/** A class handling inbound HTTP requests for the KServe API. */
+public class KServeRequestHandler extends HttpRequestHandler {
+
+    private static final Pattern PATTERN = Pattern.compile("^/v2([/?].*)?");
```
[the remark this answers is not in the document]
sindhuvahinis: Yes, you are right. Will change this.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            ChannelHandlerContext ctx,
+            FullHttpRequest req,
+            QueryStringDecoder decoder,
+            String[] segments)
+            throws ModelException {
+        HttpMethod method = req.method();
+
+        if (HttpMethod.GET.equals(method) && isKServeDescribeModelReq(segments, method)) {
+            handleKServeDescribeModel(ctx, segments);
+        }
```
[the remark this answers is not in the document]
sindhuvahinis: Okay.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
sindhuvahinis: Oh, Their protocol wants to return a json for error response. 

```
metadata_model_error_response =
{
  "error": $string
}
```

https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#model-metadata-response-json-error-object

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
sindhuvahinis: Yes, It is required.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
sindhuvahinis: They did not seem to specify any specific error code. 

I saw ONNX returns an empty object when input/output is not available. So I thought null check means, then initialization through prediction is required. 

If this is not supported by all models, then Maybe we can return  input/output, if it is available, and return empty otherwise.

serving/src/main/java/ai/djl/serving/ServerInitializer.java
```diff
@@ -81,6 +82,7 @@ public void initChannel(Channel ch) {
                 pipeline.addLast("management", new ManagementRequestHandler());
                 break;
         }
+        pipeline.addLast("kserve", new KServeRequestHandler());
```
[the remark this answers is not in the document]
sindhuvahinis: So, Kserve includes a inference request too. Something like this. 

`POST v2/models/${MODEL_NAME}[/versions/${MODEL_VERSION}]/infer`  . So I thought it requries maybe both?

serving/src/main/java/ai/djl/serving/http/KServeDescribeModelResponse.java
```diff
@@ -0,0 +1,253 @@
...
+     * Adds the output to the list of outputs.
+     *
+     * @param name name of the output.
+     * @param dataType datatype of the output.
+     * @param shape shape of the output.
+     */
+    public void addOutput(String name, DataType dataType, Shape shape) {
+        List<Long> shapeList = Arrays.stream(shape.getShape()).boxed().collect(Collectors.toList());
+        KServeDataType kdt = KServeDataType.getKServeDtForNDArrayDT(dataType);
+        this.inputs.add(new KServeIO(name, kdt.name(), shapeList));
```
oyy2000, line 97: I tried these codes and found out that this 'inputs' should be replaced by outputs otherwise there will be a nullPointer error.

oyy2000: I think the root problem is that sometimes the  model.describeOutput() will return null not the above one.

sindhuvahinis: Oh good catch

sindhuvahinis: But the null should be fixed now. I added null check in handler.

oyy2000: ok

frankfliu (approved):

PR #178, opened 2022-07-26
[serving] Fixes insecure Xml parsing

frankfliu:
## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
+++ b/serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
@@ -199,6 +199,9 @@ public class DependencyManager {
                         + ".pom";
         try (InputStream is = new URL(maven).openStream()) {
             DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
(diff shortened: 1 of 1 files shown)


serving/src/main/java/ai/djl/serving/plugins/DependencyManager.java
```diff
@@ -199,6 +199,8 @@ private String getOrtVersion(String djlVersion) throws IOException {
                         + ".pom";
         try (InputStream is = new URL(maven).openStream()) {
             DocumentBuilderFactory dbf = DocumentBuilderFactory.newInstance();
+            dbf.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            dbf.setXIncludeAware(false);
             DocumentBuilder db = dbf.newDocumentBuilder();
```
siddvenk, line 204: recommendation is to also set the following on the factory `dbf.setExpandEntityReferences(false);`

frankfliu: > recommendation is to also set the following on the factory `dbf.setExpandEntityReferences(false);`

Fixed

siddvenk (approved):

PR #198, opened 2022-08-08
Write infer protocol for Kserve

oyy2000:
## Description ##

Implements APIs below:
- v2/models/resnet18_v1/infer

WIP: no testing samples yet but test locally in a right way, and the DJL output still can not convert in a easy way to be  compatible for the Kserve in link https://github.com/kserve/kserve/blob/master/docs/predict-api/v2/required_api.md#response-output


serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -174,4 +392,124 @@ private void onException(Exception ex, ChannelHandlerContext ctx) {
...
+            status = HttpResponseStatus.SERVICE_UNAVAILABLE;
+        } else {
+            logger.warn("Unexpected error", t);
+            status = HttpResponseStatus.INTERNAL_SERVER_ERROR;
+        }
+        if (ctx != null) {
+            NettyUtils.sendError(ctx, status, t);
+        }
+    }
+}
```
sindhuvahinis, line 411: Merge this method with `OnException` method. Since the error format is same for all the APIs in v2.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -174,4 +392,124 @@ private void onException(Exception ex, ChannelHandlerContext ctx) {
...
+
+class RequestInput {
+    private String name;
```
sindhuvahinis, line 414: Will it be possible to reuse KServeIO class?

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -174,4 +392,124 @@ private void onException(Exception ex, ChannelHandlerContext ctx) {
...
+    public void setId(String id) {
+        this.id = id;
+    }
+
+    public void setList(ArrayList<ResponseOutput> responseOutput) {
+        this.outputs = responseOutput;
+    }
+}
+
+class ResponseOutput {
```
sindhuvahinis, line 494: ResponseOutput and RequestInput have the same format. You can use single class for this.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -160,6 +185,199 @@ private void handleKServeDescribeModel(ChannelHandlerContext ctx, String[] segme
...
+
+    private ArrayList<RequestOutput> getOutputListFromString(String requestString) {
+        GsonBuilder builder = new GsonBuilder();
+        builder.setPrettyPrinting();
+        Gson gson = builder.create();
+        ArrayList<RequestOutput> arrayList =
+                gson.fromJson(
+                        requestString, new TypeToken<ArrayList<RequestOutput>>() {}.getType());
+        return arrayList;
+    }
```
sindhuvahinis, line 217: Instead of parsing it as a string and then converting to RequestInput/RequestOutput, you can directly parse the bytes to the respective class. And if the user given input/output format is wrong, throw the exception.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -174,4 +392,124 @@ private void onException(Exception ex, ChannelHandlerContext ctx) {
 
         NettyUtils.sendJsonResponse(ctx, content, status);
     }
+
+    void onModelException(Throwable t, ChannelHandlerContext ctx) {
```
frankfliu, line 396: This should be merged into onException()

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
sindhuvahinis: See whether this design would be helpful for you. 

You can move that KServeIO class from inner class to http package. 
Make KServeIO to extend Input.java
And then you can add any other variables that are needed. 
You can methods which would directly convert bytes to the request input or output class you needed.

frankfliu:
I rebased your code from master. Please use rebase going forward.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -242,6 +267,199 @@ private ModelInfo<Input, Output> getModelInfo(String modelName, String modelVers
...
+
+        FullHttpResponse resp = new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, status, false);
+        for (Map.Entry<String, String> entry : output.getProperties().entrySet()) {
+            resp.headers().set(entry.getKey(), entry.getValue());
+        }
+        // construct the output
+        BytesSupplier data = output.getData();
+        String datatype =
+                output.get("datatype") != null ? output.get("datatype").getAsString() : "null";
+//        String shape = output.get("shape") != null ? output.get("shape").getAsString() : "null";
```
frankfliu, line 392: Please format your code before commit:

```
./gradlew fJ
```

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -258,4 +476,125 @@ private void onException(Exception ex, ChannelHandlerContext ctx) {
...
+            logger.warn("Unexpected error", t);
+            status = HttpResponseStatus.INTERNAL_SERVER_ERROR;
+        }
+        if (ctx != null) {
+            NettyUtils.sendError(ctx, status, t);
+        }
+    }
+
+
+    public static final class KserveIO {
```
frankfliu, line 497: We should never write two top level class in single file.

All other classes must be inner class.

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -242,6 +267,199 @@ private ModelInfo<Input, Output> getModelInfo(String modelName, String modelVers
...
+        }
+        Input inferenceRequest = requestParser.parseRequest(req, decoder);
+        infer(ctx, inferenceRequest, modelName, modelVersion);
+    }
+
+    private ArrayList<KserveIO> getInputListFromString(String requestString) {
+        GsonBuilder builder = new GsonBuilder();
+        builder.setPrettyPrinting();
+        Gson gson = builder.create();
+        return gson.fromJson(requestString, new TypeToken<ArrayList<KserveIO>>() {}.getType());
```
frankfliu, line 288: I think it's simpler use `KserveIO[]`

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -242,6 +267,199 @@ private ModelInfo<Input, Output> getModelInfo(String modelName, String modelVers
...
+        String modelVersion = null;
+        if (segments.length > 5) {
+            modelVersion = segments[5];
+        }
+        Input inferenceRequest = requestParser.parseRequest(req, decoder);
+        infer(ctx, inferenceRequest, modelName, modelVersion);
+    }
+
+    private ArrayList<KserveIO> getInputListFromString(String requestString) {
+        GsonBuilder builder = new GsonBuilder();
```
frankfliu, line 285: Use `JsonUtils.GSON`

serving/src/main/java/ai/djl/serving/http/KServeRequestHandler.java
```diff
@@ -242,6 +267,199 @@ private ModelInfo<Input, Output> getModelInfo(String modelName, String modelVers
...
+
+        Input input = new Input();
+        // construct the  input
+        try (NDManager manager = NDManager.newBaseManager()) {
+            NDList list = new NDList();
+            for (KserveIO requestInput : inputsArrayList) {
+                List<Double> dataList = requestInput.getData();
+                double[] dataArray = dataList.stream().mapToDouble(j -> j).toArray();
+
+                Shape shape = new Shape(requestInput.getShape());
```
frankfliu, line 328: I don't think we need shape as NDArray to pass down to model

frankfliu (approved):

PR #203, opened 2022-08-09
[Docker] add DLC license builder for packages we use

lanking520:
## Description ##

In compliance of all necessary licenses, we equipped the license builder created by DLC team to patch the licences after build


serving/docker/Dockerfile
```diff
@@ -48,3 +48,13 @@ RUN scripts/install_python.sh && \
     djl-serving -i com.microsoft.onnxruntime:onnxruntime:1.12.1 && \
     rm -rf scripts && pip3 cache purge && \
     apt-get clean -y && rm -rf /var/lib/apt/lists/*
+
+RUN HOME_DIR=/root && PYTHON=python \
```
frankfliu, line 52: Can we make this into shell script?

frankfliu (approved):

Issue #211, opened 2022-08-11
[Docker] pip install issue


lanking520:
## Description

the pip command in the 0.18.0 docker is broken for some reason. Always have error code.

lanking520:
it turns out 0.18.0 shipped 22.1.2 is broken. The same issue was not seen in previous version or latter version

PR #212, opened 2022-08-11
[Docker] Fix pip package issue

lanking520:
## Description ##

pip is not always following the latest available

Fixes https://github.com/deepjavalibrary/djl-serving/issues/211

### DIFF
--- a/serving/docker/scripts/install_python.sh
+++ b/serving/docker/scripts/install_python.sh
@@ -5,6 +5,6 @@ set -e
 # Ubuntu 20.04 ships python3.8 by default
 apt-get update
 DEBIAN_FRONTEND=noninteractive apt-get install -yq python3-dev python3-pip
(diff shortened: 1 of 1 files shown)


serving/docker/scripts/install_python.sh
```diff
@@ -5,6 +5,6 @@ set -e
 # Ubuntu 20.04 ships python3.8 by default
 apt-get update
 DEBIAN_FRONTEND=noninteractive apt-get install -yq python3-dev python3-pip
-python3 -m pip --no-cache-dir install -U pip
+python3 -m pip --no-cache-dir install -U pip --upgrade
```
frankfliu, line 8: Why we need this?

-U == --upgrade
```
-U, --upgrade               Upgrade all specified packages to the newest available version. The handling of dependencies depends on the
                              upgrade-strategy used.
```

PR #217, opened 2022-08-15
[Frontend] fix CVE and add frontend test

lanking520:
## Description ##

ejs dpendencies is causing critical CVE and patch with latest

### DIFF
--- a/plugins/management-console/webapp/package.json
+++ b/plugins/management-console/webapp/package.json
@@ -20,9 +20,9 @@
   },
   "devDependencies": {
     "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@vue/cli-plugin-babel": "~4.5.0",
-    "@vue/cli-plugin-router": "~4.5.0",
-    "@vue/cli-service": "~4.5.0",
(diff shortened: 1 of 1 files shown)


lanking520:
@c007456

.github/workflows/frontend.yml
```diff
@@ -0,0 +1,32 @@
...
+      - uses: actions/cache@v2
+        with:
+          path: ~/.gradle/caches
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: run npm build
+        run: |
+          cd plugins/management-console
+          ./gradlew bCA
```
frankfliu, line 32: This is part of gradle build already, we don't need a separate workflow

lanking520: How do you run this without setting up npm?

frankfliu: gradle node plugin download the node and npm

frankfliu (approved):

PR #282, opened 2022-09-30
Create an ensemble workflow

zachgk:
This creates a test for an ensemble workflow. As part of it, it also modifies the workflow expression evaluation to return an Item instead of an Input. This is primarily so that a list can be returned. It is useful for more logical or control flow functions like map, filter, functionsApply, etc.

It also creates two new built-ins: ensemble and functionsApply.

Lastly, the dependency on Ensembleable means that the current version of DJL Serving must depend on DJL 0.20.0 (including as snapshot).

### DIFF
--- a/.github/workflows/client-test.yml
+++ b/.github/workflows/client-test.yml
@@ -51,7 +51,6 @@ jobs:
           kill %1
       - name: Run Python binary mode tests
         run: |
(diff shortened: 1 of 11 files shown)


lanking520:
please feel free to rebase and we can merge

frankfliu (approved):

PR #292, opened 2022-10-16
[python] Adds built-in DeepSpeed handler

frankfliu:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -1,0 +1,119 @@
+#!/usr/bin/env python
+#
+# Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+import json
+import logging
+import os
+from typing import Optional
+
+from djl_python.inputs import Input
+from djl_python.outputs import Output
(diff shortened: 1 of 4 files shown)


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -0,0 +1,118 @@
...
+
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        if data_type == "fp16":
+            model.half()
+
+        model = deepspeed.init_inference(model,
+                                         mp_size=mp_size,
+                                         dtype=model.dtype,
+                                         replace_method='auto',
```
lanking520, line 68: This line is not always used. Sometimes loading checkpoint will need to remove this line

engines/python/src/test/resources/gpt2/serving.properties
```diff
@@ -0,0 +1,10 @@
+engine=DeepSpeed
+option.entryPoint=djl_python.deepspeed
+option.parallel_loading=true
+option.tensor_parallel_degree=2
+option.model_loading_timeout=600
```
lanking520, line 5: This time is too short?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -0,0 +1,118 @@
...
+                # TODO: check all supported architectures
+
+        logging.info(
+            f"Init: {model_id}, tensor_parallel_degree={mp_size}, data_type={data_type}, "
+            f"device={device}, max_new_tokenx={self.max_new_tokens}")
+
+        model = AutoModelForCausalLM.from_pretrained(model_id)
+        tokenizer = AutoTokenizer.from_pretrained(model_id)
+        if data_type == "fp16":
+            model.half()
```
lanking520, line 63: It may be more efficient to allow HuggingFace do the FP16 conversion during loading

lanking520 (commented):
I would suggest we look back on DeepSpeed config to make it more generic. This solution may only applies to small and simple model. For large model requires CPU memory load saving, DeepSpeed checkpoint loading may not work

lanking520 (approved):

PR #295, opened 2022-10-17
print full logging message on exception

lanking520:
https://stackoverflow.com/questions/3702675/how-to-catch-and-print-the-full-exception-traceback-without-halting-exiting-the

Print full error message


engines/python/setup/djl_python_engine.py
```diff
@@ -117,6 +117,7 @@ def run_server(self):
                     outputs = Output(code=204, message="No content")
             except Exception as e:
                 logging.exception("Failed invoke service.invoke_handler()")
+                logging.exception(e)
```
frankfliu, line 120: logging.exception suppose print out the full stack trace, did you observe more information in model exception unit test?

PR #322, opened 2022-11-17
support download model from s3

lanking520:
## Description ##

Add S3 download capabiltiy to Python engine

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -27,6 +27,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -280,6 +288,32 @@ private void createAllPyProcesses(int mpiWorkers) {
...
+                        };
+            } else {
+                logger.info("s3cmd is not installed, using aws cli");
+                commands =
+                        new String[] {
+                            "aws", "s3", "sync", url, modelDir.toAbsolutePath().toString()
+                        };
+            }
+            Process exec = Runtime.getRuntime().exec(commands);
+            exec.waitFor();
```
frankfliu, line 311: We need read the process output, if there is a lot of output, waitFor may hang.

frankfliu (approved):

PR #335, opened 2022-11-28
fix the telemetry inaccessible issues

lanking520:
## Description ##

By default, IMDSV2 is not accessible from the network

### DIFF
--- a/tests/integration/launch_container.sh
+++ b/tests/integration/launch_container.sh
@@ -22,7 +22,7 @@ set -x
 docker run \
   -itd \
   --rm \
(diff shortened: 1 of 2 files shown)


tests/integration/launch_container.sh
```diff
@@ -22,7 +22,7 @@ set -x
 docker run \
   -itd \
   --rm \
-  -p 8080:8080 \
+  --network="host" \
```
frankfliu, line 25:
```suggestion
  --network=host \
```

frankfliu (approved):

PR #339, opened 2022-11-29
[G5][Docker] add gptj model

lanking520:
## Description ##

Test GPTJ model for inference


engines/python/setup/djl_python/inputs.py
```diff
@@ -14,6 +14,7 @@
 import ast
```
frankfliu, line 14: remove ast

frankfliu (approved):

PR #349, opened 2022-12-01
[Docker][G5] add huggingface tests

lanking520:
## Description ##

Add LLM huggingface handler tests

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -51,22 +51,26 @@ class HuggingFaceService(object):
         device_id = int(properties.get("device_id", "-1"))
         model_id = properties.get("model_id")
         task = properties.get("task")
+        tp_degree = int(properties["tensor_parallel_degree", "-1"])
         # HF Acc handling
         kwargs = {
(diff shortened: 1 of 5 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -58,6 +59,11 @@ def initialize(self, properties: dict):
         # https://huggingface.co/docs/accelerate/usage_guides/big_modeling#designing-a-device-map
         if "device_map" in properties:
             kwargs["device_map"] = properties.get("device_map")
+            logging.info(f"Using device map {kwargs['device_map']}")
+        elif "tensor_parallel_degree" in properties:
+            kwargs["device_map"] = "auto"
+            world_size = torch.cuda.device_count()
+            logging.info(f"Using {world_size} gpus")
```
frankfliu, line 66: We should use tensor_paralel_detree here. or we should add a check to verify world_size and tensor_paralel_detree are equal?

lanking520: added a check here

frankfliu (approved):

PR #359, opened 2022-12-05
[WIP] Updating deepspeed handler for more models

siddvenk:
## Description ##

This PR adds support for built in handlers to serve models using the DeepSpeed engine.

2 handlers are added/modified:
- Stable Diffusion handler
- Transformer/LLM handler

Testing for Stable Diffusion and most language models except bloom have been tested. Things left to be done:
- Stable Diffusion is a separate handler as I felt it was cleaner this way. It is different enough that implementing it within a generic deepspeed handler felt messy. But, right now it's tied to DS so we may want to modify it so it can work with accelerate or deepspeed
  - Currently I have only tested the text2image task from HF. Others should be easy to add, but might require additional input/output processing
- In the deepspeed handler
  - Support user supplied injection policy 
  - Support user provided MPU
  - Better support for batch inference (right now it works, but the usage/interface could probably be improved)
  - DS config is written for 0.7.6. It can be modified to be compatible with 0.7.5 if that's the version we plan on including in the next release
  - Adding tests

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -95,13 +95,16 @@ MODEL_TYPE_TO_INJECTION_POLICY = {
 
 
 def get_torch_dtype_from_str(dtype: str):
+    if dtype == "fp32":
+        return torch.float32
     if dtype == "fp16":
         return torch.float16
-    if dtype == "bf16":
+    elif dtype == "bf16":
         return torch.bfloat16
-    if dtype == "int8":
+    elif dtype == "int8":
         return torch.int8
-    return torch.float32
+    else:
+        raise ValueError(f"Invalid data type: {dtype}")
 
 
 class DeepSpeedService(object):
@@ -133,7 +136,7 @@ class DeepSpeedService(object):
                          f"tensor_parallel_degree: {self.tensor_parallel_degree}")
         self.initialized = True
 
-    def parse_properties(self, properties):
+    def _parse_properties(self, properties):
         self.model_dir = properties.get("model_dir")
         self.model_id = properties.get("model_id")
         self.task = properties.get("task")
@@ -159,7 +162,7 @@ class DeepSpeedService(object):
             "max_tokens": self.max_tokens,
         }
 
-    def validate_model_type_and_task(self):
+    def _validate_model_type_and_task(self):
         if not self.model_id:
             self.model_id = self.model_dir
             config_file = os.path.join(self.model_id, "config.json")
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -10,98 +10,246 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-
-import json
 import logging
 import os
-from typing import Optional
-
+import torch
+from transformers import (
+    AutoConfig,
+    PretrainedConfig,
+    AutoTokenizer,
+    AutoModelForCausalLM,
+    AutoModelForSequenceClassification,
+    AutoModelForQuestionAnswering,
+    AutoModelForMaskedLM,
+    AutoModelForTokenClassification,
+    pipeline,
+    Conversation,
+    SquadExample,
+)
+import transformers
+from deepspeed.module_inject.replace_policy import (
+    HFBertLayerPolicy,
+    HFGPTNEOLayerPolicy,
+    GPTNEOXLayerPolicy,
+    HFGPTJLayerPolicy,
+    MegatronLayerPolicy,
+    HFGPT2LayerPolicy,
+    BLOOMLayerPolicy,
+    HFOPTLayerPolicy,
+    HFCLIPLayerPolicy,
+)
+import deepspeed
 from djl_python.inputs import Input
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
         self.initialized = False
+        self.ds_config = None
+        self.task = None
+        self.logger = logging.getLogger()
+        self.model_dir = None
+        self.model_id = None
+        self.data_type = None
+        self.max_tokens = None
+        self.device = None
+        self.world_size = None
```
frankfliu, line 109: world_size should be identical to tensor_parallel_degree

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
+        self.validate_model_type_and_task()
+        self.create_model_pipeline()
+        self.logger.info(f"Initialized DeepSpeed model with the following configurations"
+                         f"model: {self.model_id}"
+                         f"task: {self.task}"
+                         f"data_type: {self.data_type}"
+                         f"tensor_parallel_degree: {self.tensor_parallel_degree}")
+        self.initialized = True
+
+    def parse_properties(self, properties):
```
frankfliu, line 124:
```suggestion
    def _parse_properties(self, properties):
```

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
+            "checkpoint": properties.get("checkpoint"),
+            "base_dir": properties.get("base_dir"),
+            "return_tuple": properties.get("return_tuple", "true").lower() == "true",
+            "training_mp_size": int(properties.get("training_mp_size", 1)),
+            "replace_method": "auto",
+            "injection_policy": None,
+            "max_out_tokens": self.max_tokens,
+        }
+
+    def validate_model_type_and_task(self):
```
frankfliu, line 152:
```suggestion
    def _validate_model_type_and_task(self):
```

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
+                         f"data_type: {self.data_type}"
+                         f"tensor_parallel_degree: {self.tensor_parallel_degree}")
+        self.initialized = True
+
+    def parse_properties(self, properties):
+        self.model_dir = properties.get("model_dir")
+        self.model_id = properties.get("model_id")
+        self.task = properties.get("task")
+        self.data_type = properties.get("data_type", "fp32")
+        self.max_tokens = int(properties.get("max_tokens", 1024))
```
frankfliu, line 129: what should be a proper default to max_tokens?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
+            "dtype": self.data_type,
+            "tensor_parallel": {
+                "enabled": True,
+                "tp_size": self.tensor_parallel_degree,
+                "mpu": None,
+            },
+            "enable_cuda_graph": properties.get("enable_cuda_graph", "false").lower() == "true",
+            "triangular_masking": properties.get("triangular_masking", "true").lower() == "true",
+            "checkpoint": properties.get("checkpoint"),
+            "base_dir": properties.get("base_dir"),
```
frankfliu, line 144: Why user need to set `base_dir`?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -10,98 +10,246 @@
...
-        self.max_new_tokens = 0
+        self.pipeline = None
         self.initialized = False
+        self.ds_config = None
+        self.task = None
+        self.logger = logging.getLogger()
+        self.model_dir = None
+        self.model_id = None
+        self.data_type = None
+        self.max_tokens = None
```
frankfliu, line 107: We might need `min_tokens`

engines/python/setup/djl_python/deepspeed.py
siddvenk: 👍

engines/python/setup/djl_python/deepspeed.py
siddvenk: 👍

engines/python/setup/djl_python/deepspeed.py
siddvenk: If base_dir is not set, DS will look for checkpoint in current directory. If checkpoint is in some child dir then base_dir helps to identify it (though it works if you specify the full path to checkpoint and leave base_dir empty). Not sure if we need to expose this, but if user is coming from DS world they might expect this.

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -93,6 +94,16 @@
 }
 
 
+def get_torch_dtype_from_str(dtype: str):
+    if dtype == "fp16":
```
frankfliu, line 98: if dtype == "fp32":
        return torch.float32
    if dtype == "fp16": 
        return torch.float16
    elif dtype == "bf16":
        return torch.bfloat16
    elif dtype == "int8":
        return torch.int8
    else:
        raise ValueError(f"Invalid data type: {dtype}")

engines/python/setup/djl_python/deepspeed.py
siddvenk: Deepspeed uses 1024 as the default, so I think keeping it the same as that makes sense. 

Though, i think task specific pipelines on HF side might override that and set other defaults.

lanking520 (approved):

PR #366, opened 2022-12-07
[DeepSpeed] Fix datatype and allow checkpoint loading

lanking520:
## Description ##

Add a few fixes on data type and checkpoint loading settings


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -95,11 +95,11 @@
 
 
 def get_torch_dtype_from_str(dtype: str):
```
siddvenk, line 97: Do we want to include something like this https://github.com/microsoft/DeepSpeed-MII/blob/main/mii/config.py#L7?

Or do we need to add documentation about what the supported dtype strings are in serving.properties?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -95,11 +95,11 @@
...
-    if dtype == "fp32":
+    if dtype == "float32":
```
frankfliu, line 98: I'm not sure if full name is better.
seems fp32, fp16, bf16 are well accepted name already

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -140,11 +139,10 @@ def _parse_properties(self, properties):
         self.model_dir = properties.get("model_dir")
         self.model_id = properties.get("model_id")
         self.task = properties.get("task")
-        self.data_type = get_torch_dtype_from_str(properties.get("data_type", "fp32"))
+        self.data_type = get_torch_dtype_from_str(properties.get("dtype", "float16"))
```
siddvenk, line 142: float16 ->fp16

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -140,11 +139,10 @@ def _parse_properties(self, properties):
         self.model_dir = properties.get("model_dir")
         self.model_id = properties.get("model_id")
         self.task = properties.get("task")
-        self.data_type = get_torch_dtype_from_str(properties.get("data_type", "fp32"))
+        self.data_type = get_torch_dtype_from_str(properties.get("dtype", "fp16"))
```
frankfliu, line 142: Why use fp16 as default?

lanking520: fp16 works better and better range of supported model on DeepSpeed. model like Bloom is not working on fp32

frankfliu: I mean use a normal user, they will assume fp32 is default. They only override it to fp16 intentionally

lanking520: most LLM saved on huggingface is under fp16

frankfliu (approved):

PR #371, opened 2022-12-08
[python] Add dynamic batching feature to python engine

frankfliu:
## Description ##


lanking520:
Can you add a test case in client test or integration test?

lanking520 (approved):

PR #377, opened 2022-12-10
Update sd handler to return image in request, fix some default values…

siddvenk:
## Description ##
- Updates stable diffusion handler to return content-type image/png to user so client can receive image
- Change some default values from None to empty version of corresponding data type (causes issues in generate if None)
- Add test for stable-diffusion-v1-4 model

### DIFF
--- a/engines/python/setup/djl_python/stable-diffusion.py
+++ b/engines/python/setup/djl_python/stable-diffusion.py
@@ -54,21 +54,25 @@ class StableDiffusionService(object):
         self.data_type = get_torch_dtype_from_str(properties.get("dtype"))
         self.max_tokens = int(properties.get("max_tokens", "1024"))
         self.device = int(os.getenv("LOCAL_RANK", "0"))
-        self.tensor_parallel_degree = int(properties.get("tensor_parallel_degree", 1))
+        self.tensor_parallel_degree = int(
+            properties.get("tensor_parallel_degree", 1))
         self.ds_config = {
             "replace_with_kernel_inject": True,
             # TODO: Figure out why cuda graph doesn't work for stable diffusion via DS
             "enable_cuda_graph": False,
             "replace_method": "auto",
-            "dtype": torch.float16 if self.data_type == "fp16" else torch.float32,
+            "dtype":
+            torch.float16 if self.data_type == "fp16" else torch.float32,
             "mp_size": self.tensor_parallel_degree
         }
 
         if not self.model_id:
             config_file = os.path.join(self.model_dir, "model_index.json")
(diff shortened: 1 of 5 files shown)


engines/python/setup/djl_python/stable-diffusion.py
```diff
@@ -39,12 +40,10 @@ def __init__(self):
     def initialize(self, properties: dict):
         self.model_dir = properties.get("model_dir")
         self.model_id = properties.get("model_id")
-        self.data_type = properties.get("data_type", "fp32")
+        self.data_type = properties.get("data_type", "fp16")
```
frankfliu, line 43: We should make it consistent with other handler:
1. data_type vs dtype, or we support both
2. default value should be None, we cast the data type only if user explicitly set the value

frankfliu (approved):

engines/python/setup/djl_python/stable-diffusion.py
siddvenk: good catch!

tests/integration/llm/prepare.py
```diff
@@ -33,6 +33,11 @@
                 "option.task": "text-generation"}
 }
 
+sd_handler_list = {
+    "stable-diffusion-v1-4": {"option.s3url": "s3://djl-llm/stable-diffusion-v1-4", "option.tensor_parallel_degree": 2,
```
lanking520, line 37:
```suggestion
    "stable-diffusion-v1-4": {"option.s3url": "s3://djl-llm/stable-diffusion-v1-4/", "option.tensor_parallel_degree": 2,
```

tests/integration/llm/client.py
```diff
@@ -114,6 +118,24 @@ def test_ds_raw_model(model):
                 assert float(memory) / 1024.0 < spec["max_memory_per_gpu"]
 
 
+def test_sd_handler(model, model_spec):
+    if model not in model_spec:
+        raise ValueError(f"{model} is not one of the supporting models {list(sd_model_spec.keys())}")
+    spec = sd_model_spec[model]
```
lanking520, line 124: please also add a worker number check in here

lanking520 (approved):

tests/integration/llm/prepare.py
siddvenk: done

tests/integration/llm/client.py
siddvenk: good catch

lanking520:
Merge it for now, let's fix it in the next PR

PR #378, opened 2022-12-10
[docker] Adds non-root user to docker

frankfliu:
Allows user run DJL docker image as non-root user

## Description ##


lanking520:
Is that okay if the work directory is different? WORKDIR /opt/djl

frankfliu:
User's home directory has nothing to do with working directory

lanking520 (approved):

PR #438, opened 2023-01-22
[test] Fixes prepare.py typo

frankfliu:
## Description ##

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -77,7 +77,7 @@ sd_handler_list = {
 }
 
 
-def write_prperties(properties):
+def write_properties(properties):
     model_path = "models/test"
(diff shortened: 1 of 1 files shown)


tests/integration/llm/prepare.py
```diff
@@ -60,6 +60,11 @@
         "option.s3url": "s3://djl-llm/opt-13b/",
         "option.tensor_parallel_degree": 2,
         "option.task": "text-generation"
+    },
+    "opt-30b": {
```
lanking520, line 64: we cannot add this since our testing is on g5 12x. Memory is not enough

frankfliu: removed

lanking520 (approved):

PR #448, opened 2023-01-28
Remove unecessary java engine adapter

frankfliu:
## Description ##


tools/gradle/publish.gradle
```diff
@@ -23,12 +22,6 @@ configure([
         options.header "<script type='text/javascript' src='https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.1/MathJax.js?config=TeX-AMS-MML_HTMLorMML'></script>"
     }
 
-    task uploadJavadoc(type: Exec) {
```
zachgk, line 26: Don't we still want this for uploading Javadoc for python engine, wlm, and serving?

frankfliu: We use javadoc.io for our javadoc, I stopped upload generated javadoc to our s3 for a long time.

zachgk (approved):

PR #451, opened 2023-02-02
Fix issues with model_dir and model_id usage that occur when s3url is…

siddvenk:
## Description ##

There are some issues in our python handlers with the usage of model_dir and model_id. When s3url is provided in serving.properties, we override the model_id to be the path of the tmp directory where artifacts are stored (workaround implemented to work with sagemaker). This led to some incorrect usage of model_dir when model_id should have been used, and vis versa. Specifically, if user provides s3 url but not the task, the model fails to load because it was looking in the wrong place for model config file.

This change fixes those issues by only using 1 variable as the model artifact identifier.


sindhuvahinis:
Just curious, Will model_dir be set when model is downloaded through s3_url?

frankfliu (commented):
In HuggingFace pipeline API, the parameter "model" can be model_id or model_dir,
We explicitly change it to `model_dir` can cause confusion in future. we either use general name like "model" or find some else

We can keep the key name in serving.properties "model_id" and "model_dir"

siddvenk:
> Just curious, Will model_dir be set when model is downloaded through s3_url?
 
There is no additional logic to change the model_dir when using s3_url. We instead set the model_id key, which will point to the tmp dir where the model has been downloaded.

>We explicitly change it to model_dir can cause confusion in future. we either use general name like "model" or find some else

I agree, the naming here is confusing. Keeping in line with huggingface api, I think the name `pretrained_model_name_or_path` should work well here in our handler. We still let the user set either `s3url`, `model_id`, or `model_dir`, but internally we only need to worry about 1 identifier/location to do the loading. It becomes confusing checking which key to use in multiple places in the handler currently.

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -128,8 +128,7 @@ def __init__(self):
         self.ds_config = None
         self.task = None
         self.logger = logging.getLogger()
-        self.model_dir = None
-        self.model_id = None
+        self.pretrained_model_name_or_path = None
```
frankfliu, line 131: I feel `model_id_or_path` is better, we exposed `model_id` key in the serving.properties to match huggingface, using model_name here is not consistent.

siddvenk: Good point, will update.

frankfliu (approved):

lanking520:
Did we cover these changes in test? Also by default we are still overriding model_id in Java end. Did you make the change accordingly?

PR #460, opened 2023-02-06
add python3.9 for fastertransformer

lanking520:
## Description ##

Add python 3.9 since DLC by default install python 3.9


rohithkrn:
without 3.9, is it causing discrepancy with pytorch installation?

lanking520:
> without 3.9, is it causing discrepancy with pytorch installation?

Yes. python3.8 cannot install 3.9 wheel

frankfliu (approved):

PR #465, opened 2023-02-08
Update python handlers for deepspeed 0.8.0

siddvenk:
## Description ##

Updating the deepspeed handlers for 0.8.0. Minimal changes are required to support the current functionality.

So far I have tested gptj-6b with fp16, fp32, bf16 with no issues.

In 0.8.0, Deepspeed has changed how the replacement of layers works. Only 1 of replace_method and injection_policy will be honored. Auto replacement seems to be working as expected now (previously it was broken for some model classes like opt), so we defer to that.

Remaining work pending wheel updates:
- int8 integration and corresponding testing
- Model loading from partitioned checkpoints
- Adding support for new ds configs related to quantization

Remaining enhancements:
- Support user provided injection policy
- Dynamic batching support (applies to all python handlers)

### DIFF
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -14,8 +14,7 @@ FROM nvidia/cuda:$version
 ARG djl_version=0.21.0~SNAPSHOT
 ARG torch_version=1.13.1
 ARG accelerate_version=0.16.0
-# This needs to be added once available in s3
-ARG deepspeed_wheel=""
+ARG deepspeed_wheel="https://publish.djl.ai/deepspeed/deepspeed-0.8.0-py2.py3-none-any.whl"
 ARG transformers_version=4.26.0
 ARG diffusers_version=0.12.0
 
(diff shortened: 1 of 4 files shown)


lanking520:
let's stick with Ubuntu 20.04 at the moment to align with standard dlc. We can switch to 22.04 when 18.04 on deprecation path

serving/docker/deepspeed.Dockerfile
```diff
@@ -45,7 +45,7 @@ RUN apt-get update && \
     scripts/install_s5cmd.sh x64 && \
     DEBIAN_FRONTEND=noninteractive apt-get install -yq libaio-dev libopenmpi-dev && \
     pip3 install torch==${torch_version} --extra-index-url https://download.pytorch.org/whl/cu116 && \
```
lanking520, line 47: Torch url need to change to cu117

siddvenk: done

lanking520 (approved):

PR #469, opened 2023-02-09
Add fastertransformer t5 integration test

rohithkrn:
## Description ##

Add an intial fastertransformer t5 integration test to test the workflow. Will add more tests after this.

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -74,7 +74,8 @@ ds_model_spec = {
 
 ft_raw_model_spec = {
     "t5-small": {
-        "batch_size": [1, 2]
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 2
     }
 }
 
@@ -137,7 +138,8 @@ def batch_generation(batch_size):
 def t5_batch_generation(batch_size):
     input_sentences = [
         "translate English to German: The house is wonderful.",
-        "translate English to German: My name is AWS",
+        "summarize: state authorities dispatched emergency crews tuesday to survey the damage after an onslaught \
+             of severe weather in mississippi…",
     ]
     if batch_size > len(input_sentences):
         input_sentences *= math.ceil(batch_size / len(input_sentences))
@@ -230,7 +232,12 @@ def test_ft_raw_handler(model, model_spec):
         req = {"inputs" : t5_batch_generation(batch_size)}
         res = send_json(req)
         res = res.json()
+        print("inference output: ", res)
         assert len(res) == batch_size
+        memory_usage = get_gpu_memory()
+        logging.info(memory_usage)
+        for memory in memory_usage:
+            assert float(memory) / 1024.0 < spec["max_memory_per_gpu"]
 
 if __name__ == '__main__':
     args = parser.parse_args()
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -72,6 +72,12 @@ ds_model_spec = {
     }
 }
 
+ft_raw_model_spec = {
+    "t5-small": {
+        "batch_size": [1, 2]
+    }
+}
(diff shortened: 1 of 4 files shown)


tests/integration/llm/prepare.py
```diff
@@ -129,11 +135,23 @@ def build_sd_handler_model(model):
     write_properties(options)
 
 
+def build_ft_raw_model(model):
+    if model not in ft_model_list:
+        raise ValueError(
+            f"{model} is not one of the supporting handler {list(ft_model_list.keys())}"
+        )
+    options = ft_model_list[model]
+    options["engine"] = "DeepSpeed"
```
lanking520, line 144: This can now be `FasterTransformer` if you rebase from master

rohithkrn:
Testing: job on personal devo branch - https://github.com/deepjavalibrary/djl-serving/actions/runs/4138893628/jobs/7155865025

tests/integration/llm/prepare.py
rohithkrn: looks like this change is not complete, ran into this error - 
```
Model Store: /opt/ml/model
Initial Models: test=file:/opt/ml/model/test/
Initial Workflows: N/A
Netty threads: 0
Maximum Request Size: 67108864
[INFO ] 2023-02-10 00:01:12 ModelServer - Initializing model: test=file:/opt/ml/model/test/
[INFO ] 2023-02-10 00:01:15 ModelServer - Model server stopped.
Error:  2023-02-10 00:01:15 ModelServer - Invalid configuration: Deep learning engine not found: FasterTransformer
```

so keeping it DeepSpeed for now.

rohithkrn: docker container is not updated with latest build of djl-serving i guess, Will change it in a subsequent PR.

tests/integration/llm/client.py
```diff
@@ -205,6 +219,18 @@ def test_sd_handler(model, model_spec):
...
+        raise ValueError(
+            f"{model} is not one of the supporting models {list(ft_raw_model_spec.keys())}"
+        )
+    spec = model_spec[model]
+    for batch_size in spec['batch_size']:
+        print(f"testing ft_handler with model: {model}, batch_size: {batch_size} ")
+        req = {"inputs" : t5_batch_generation(batch_size)}
+        res = send_json(req)
+        res = res.json()
+        assert len(res) == batch_size
```
lanking520, line 233: Could we also print the result and gpu memory usages? For different TP degree, we should also limit the memory per gpu to avoid some (load on single GPU) cases.

tests/integration/llm/client.py
```diff
@@ -128,6 +134,14 @@ def batch_generation(batch_size):
         input_sentences *= math.ceil(batch_size / len(input_sentences))
     return input_sentences[:batch_size]
 
+def t5_batch_generation(batch_size):
+    input_sentences = [
+        "translate English to German: The house is wonderful.",
+        "translate English to German: My name is AWS",
```
lanking520, line 140: Could we also add some other task other than translation?

rohithkrn: yes, plan to add more tests. Want to test the entire workflow in this PR first.

tests/integration/llm/client.py
rohithkrn: I will gather the data for this and add.

rohithkrn: Added. logs here - https://github.com/deepjavalibrary/djl-serving/actions/runs/4140117446/jobs/7158467593

tests/integration/llm/client.py
rohithkrn: replaced one case with summarization.

lanking520 (approved):

PR #475, opened 2023-02-10
[FT][LLM tests]Add FT integration tests for gpt2, opt and bloom models

rohithkrn:
## Description ##

Add FT integration tests for gpt2, opt and bloom models

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -76,7 +76,19 @@ ft_raw_model_spec = {
     "t5-small": {
         "batch_size": [1, 2],
         "max_memory_per_gpu": 2
-    }
+    },
+    "gpt2-xl": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 7.0
+    },
+    "facebook/opt-13b": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 15.0
+    },
+    "bigscience/bloomz-3b": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 8.0
+    },
 }
 
 sd_model_spec = {
@@ -228,11 +240,14 @@ def test_ft_raw_handler(model, model_spec):
         )
     spec = model_spec[model]
     for batch_size in spec['batch_size']:
-        print(f"testing ft_handler with model: {model}, batch_size: {batch_size} ")
-        req = {"inputs" : t5_batch_generation(batch_size)}
+        logging.info(f"testing ft_handler with model: {model}, batch_size: {batch_size} ")
+        if "t5" in model:
+            req = {"inputs" : t5_batch_generation(batch_size)}
+        else:
+            req = {"inputs": batch_generation(batch_size)}
         res = send_json(req)
         res = res.json()
-        print("inference output: ", res)
+        logging.info("inference output: ", res)
         assert len(res) == batch_size
         memory_usage = get_gpu_memory()
         logging.info(memory_usage)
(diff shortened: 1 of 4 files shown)


tests/integration/llm/prepare.py
```diff
@@ -81,7 +81,19 @@
...
         "option.model_id": "t5-small",
         "option.tensor_parallel_degree": 4,
-    }
+    },
+    "gpt2-xl": {
+        "option.model_id": "gpt2-xl",
+        "option.tensor_parallel_degree": 1,
+    },
+    "facebook/opt-13b": {
+        "option.model_id": "facebook/opt-13b",
```
lanking520, line 90: This model is in our S3 bucket, using s3 bucket downloading will speed up

tests/integration/llm/prepare.py
```diff
@@ -81,7 +81,19 @@
...
+        "option.tensor_parallel_degree": 4,
+    },
+    "bigscience/bloomz-3b": {
+        "option.model_id": "bigscience/bloomz-3b",
```
lanking520, line 94: bloom 7b is in the bucket

rohithkrn:
the test job failed for gpt2-xl with error: `RuntimeError: context has already been set`. Let's wait for the fix to merge in.
https://github.com/deepjavalibrary/djl-serving/actions/runs/4148142635/jobs/7175851331

tests/integration/llm/prepare.py
rohithkrn: updated

tests/integration/llm/prepare.py
rohithkrn: updated

rohithkrn:
Passing test logs - https://github.com/deepjavalibrary/djl-serving/actions/runs/4167577542/jobs/7213363647.
It's taking several tries to connect djl-serving for opt and bloom models.

tests/integration/llm/client.py
```diff
@@ -76,7 +76,19 @@
...
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 7.0
+    },
+    "facebook/opt-13b": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 15.0
+    },
+    "bigscience/bloom-7b1": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 17.0
```
lanking520, line 90: please use fp16 for testing. Also the artifact saving in this as well. Ideally, we are trying to keep a GPU memory usage under 50% (12GB).

tests/integration/llm/client.py
```diff
@@ -76,7 +76,19 @@
...
         "max_memory_per_gpu": 2
-    }
+    },
+    "gpt2-xl": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 7.0
+    },
+    "facebook/opt-13b": {
+        "batch_size": [1, 2],
+        "max_memory_per_gpu": 15.0
```
lanking520, line 86: same in here. the model is saved in fp16, not making sense to test in fp32

tests/integration/llm/client.py
rohithkrn: makes sense, these tests are written when only fp32 support was tested/present. Before updating these tests, we need to update the wheel with latest changes I guess.

tests/integration/llm/client.py
rohithkrn: updated.

tests/integration/llm/client.py
rohithkrn: updated.

rohithkrn:
Passing test logs - https://github.com/deepjavalibrary/djl-serving/actions/runs/4179684457/jobs/7239925371

lanking520 (approved):

PR #479, opened 2023-02-15
[serving] Adds SageMaker MME support

frankfliu:
## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
+++ b/serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
@@ -55,7 +55,7 @@ public class InferenceRequestHandler extends HttpRequestHandler {
     private RequestParser requestParser;
 
     private static final Pattern PATTERN =
-            Pattern.compile("^/(ping|invocations|predictions)([/?].*)?");
+            Pattern.compile("/(ping|invocations|predictions)([/?].*)?|/models/.+/invoke");
 
     /** default constructor. */
     public InferenceRequestHandler() {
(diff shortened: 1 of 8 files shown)


serving/src/main/java/ai/djl/serving/http/ManagementRequestHandler.java
```diff
@@ -242,40 +213,43 @@ private void handleRegisterModel(final ChannelHandlerContext ctx, QueryStringDec
...
                                     return null;
                                 });
-        if (synchronous) {
-            final String msg = "Model \"" + modelName + "\" registered.";
+        if (req.isSynchronous()) {
+            final String msg = "Model \"" + req.getModelName() + "\" registered.";
             f.thenAccept(v -> NettyUtils.sendJsonResponse(ctx, new StatusResponse(msg)));
         } else {
-            String msg = "Model \"" + modelName + "\" registration scheduled.";
+            String msg = "Model \"" + req.getModelName() + "\" registration scheduled.";
```
siddvenk, line 235: Is msg supposed to be final in one block and not the other?

serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
```diff
@@ -56,6 +56,7 @@ public class InferenceRequestHandler extends HttpRequestHandler {
 
     private static final Pattern PATTERN =
```
siddvenk, line 57: Can we update this variable name to be more clear now that we're introducing INVOKE_PATTERN? Or is it possible to combine these into a single pattern?

serving/src/main/java/ai/djl/serving/http/ManagementRequestHandler.java
frankfliu: When a local variable used in inner class (Lambda), it must be final (or effective final). We explicitly add final to indicate it's being used in inner class

serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
frankfliu: done

siddvenk (approved):

PR #485, opened 2023-02-17
add AI template package

lanking520:
## Description ##

Build AI template package into the container

### DIFF
--- a/serving/docker/scripts/install_aitemplate.sh
+++ b/serving/docker/scripts/install_aitemplate.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+apt-get install -y git
+
+# TODO: change the hack on the code to support A10G
+git clone https://github.com/lanking520/AITemplate --recursive
+cd AITemplate/python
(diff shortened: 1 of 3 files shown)


serving/docker/scripts/install_aitemplate.sh
```diff
@@ -0,0 +1,11 @@
...
+
+apt-get install -y git
+
+# TODO: change the hack on the code to support A10G
+git clone https://github.com/lanking520/AITemplate --recursive
+cd AITemplate/python
+python setup.py bdist_wheel
+pip install dist/*.whl --force-reinstall
+cd ../../
+rm -rf AITemplate
```
frankfliu, line 11: apt-get remove git

lanking520: Why removing git?

frankfliu: We are not using git at runtime.

frankfliu (approved):

PR #490, opened 2023-02-17
Read dtype from hf model when not supplied in config

siddvenk:
## Description ##

Tested the handler with opt-2.7b using both user supplied dtype in serving properties, and inferred dtype from model (not specified in serving.properties)

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -131,8 +131,6 @@ class DeepSpeedService(object):
         ds_config = {
             "replace_with_kernel_inject":
             True,
-            "dtype":
-            self.data_type,
             "tensor_parallel":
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -210,6 +208,10 @@ def create_model_pipeline(self):
                 self.model_id_or_path,
                 low_cpu_mem_usage=self.low_cpu_mem_usage,
                 **kwargs)
+        if self.data_type:
+            self.ds_config["dtype"] = self.data_type
+        elif model.dtype:
```
lanking520, line 213: just do else. There should be no case that model doesn't have a dtype.

siddvenk: oops 😆

lanking520 (approved):

PR #496, opened 2023-02-20
[serving] Check available memory before loading model

frankfliu:
## Description ##

### DIFF
--- a/serving/docs/modes.md
+++ b/serving/docs/modes.md
@@ -65,6 +65,10 @@ and python handler functions.
 - `gpu.maxWorkers`: Maximum number of workers for GPU.
 - `cpu.minWorkers`: Minimum number of workers for CPU.
 - `cpu.maxWorkers`: Maximum number of workers for CPU.
+- `required_memory_mb`: Specify the required memory (CPU and GPU) in MB to load the model.
+- `gpu.required_memory_mb`: Specify the required GPU memory in MB to load the model.
+- `reserved_memory_mb`: Reserve memory in MB to avoid system out of memory.
+- `gpu.reserved_memory_mb`: Reserve GPU memory in MB to avoid system out of memory.
 
 
 For example, set minimum workers and maximum workers for your model:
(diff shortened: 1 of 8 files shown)


serving/docs/modes.md
```diff
@@ -65,6 +65,10 @@ and python handler functions.
 - `gpu.maxWorkers`: Maximum number of workers for GPU.
 - `cpu.minWorkers`: Minimum number of workers for CPU.
 - `cpu.maxWorkers`: Maximum number of workers for CPU.
+- `required_memory_mb`: Specify the required memory (CPU and GPU) in MB to load the model.
```
siddvenk, line 68: For Python based engines, are we assuming that this accounts for the number of mpi workers? Or should we account for mpi workers and parallel_loading in the estimate calculation?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -621,6 +632,80 @@ private void configPerModelSettings(Path modelDir) throws IOException {
...
+        long requiredMemory = intValue(prop, "required_memory_mb", -1) * 1024L * 1024;
+        WlmConfigManager wlmc = WlmConfigManager.getInstance();
+        int defMemory = wlmc.getReservedMemoryMb();
+        long reservedMemory = intValue(prop, "reserved_memory_mb", defMemory) * 1024L * 1024;
+        if (requiredMemory <= 0) {
+            try (Stream<Path> walk = Files.walk(modelDir)) {
+                // estimate the memory to be 1.2x of file size
+                requiredMemory =
+                        walk.filter(Files::isRegularFile).mapToLong(ModelInfo::getFileSize).sum();
+                requiredMemory = requiredMemory * 12 / 10;
```
siddvenk, line 646: Is 1.2x a valid multiplier for all models? I'm ok with this for now, but I think we determined this from our experiments with DS based models, so not sure how accurate it is for other models (as long as it doesn't severely underestimate the required memory it should be ok).

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -206,7 +217,7 @@ public void load(Device device) throws ModelException, IOException {
 
             ZooModel<I, O> m = builder.build().loadModel();
             if (criteria != null) {
-                // TODO: use has to manually configure batchifer if using dynamic batch
+                // TODO: user has to manually configure birdwatcher if using dynamic batch
```
siddvenk, line 220: what is birdwatcher?

frankfliu: fixed

siddvenk (approved):

PR #498, opened 2023-02-21
[Bugfix]Add permissions to opt/ml/model

sindhuvahinis:
### This PR fixes 3 bugs of the folllowing for AOT partition

- Adding permission to /opt/ml/model to save partitioned model in this location

### DIFF
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -56,6 +56,8 @@ RUN apt-get update && \
     scripts/security_patch.sh deepspeed && \
     useradd -m -d /home/djl djl && \
     chown -R djl:djl /opt/djl && \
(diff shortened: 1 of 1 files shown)


serving/docker/deepspeed.Dockerfile
```diff
@@ -56,6 +56,8 @@ RUN apt-get update && \
     scripts/security_patch.sh deepspeed && \
     useradd -m -d /home/djl djl && \
     chown -R djl:djl /opt/djl && \
+    mkdir -p /opt/ml/model && \
+    chown -R djl:djl /opt/ml/model && \
```
lanking520, line 60: `/opt/ml` folder is read-only on SageMaker inference endpoint

frankfliu (changes requested):
We should not change owner to `djl`
in real world, user run container as root

PR #505, opened 2023-02-23
Reformat python source code

frankfliu:
## Description ##

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,3 +123,10 @@ configure(javaProjects()) {
 }
 
 apply from: file("${rootProject.projectDir}/tools/gradle/publish.gradle")
+tasks.register('formatPython') {
+    doFirst {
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -123,3 +123,10 @@ configure(javaProjects()) {
 }
 
 apply from: file("${rootProject.projectDir}/tools/gradle/publish.gradle")
+tasks.register('formatPython') {
+    doFirst {
+        exec {
+            commandLine "bash", "-c", "find . -name '*.py' -print0 | xargs -0 yapf --in-place"
```
lanking520, line 129: yapf is not installed by default

frankfliu: That's why this is not enforced, developer has to install it manually

lanking520 (approved):

PR #510, opened 2023-03-01
Add security patching for fastertransformers image

siddvenk:
## Description ##

Adds security patching for fastertransformer image on DLC branch

### DIFF
--- a/serving/docker/fastertransformer.Dockerfile
+++ b/serving/docker/fastertransformer.Dockerfile
@@ -73,13 +73,13 @@ RUN git clone https://github.com/NVIDIA/FasterTransformer.git -b ${ft_version} \
     && cp -r bin/*_gemm /usr/local/backends/fastertransformer/bin/ \
     && cd ../../ && rm -rf FasterTransformer
 
-# TODO: Add DLC patching
 RUN apt-get update && \
     scripts/install_djl_serving.sh $djl_version && \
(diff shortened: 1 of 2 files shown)


lanking520 (approved):

serving/docker/fastertransformer.Dockerfile
```diff
@@ -73,13 +73,13 @@ RUN git clone https://github.com/NVIDIA/FasterTransformer.git -b ${ft_version} \
...
     && cd ../../ && rm -rf FasterTransformer
 
-# TODO: Add DLC patching
 RUN apt-get update && \
     scripts/install_djl_serving.sh $djl_version && \
     mkdir -p /opt/djl/bin && cp scripts/telemetry.sh /opt/djl/bin && \
     echo "${djl_version} fastertransformer" > /opt/djl/bin/telemetry && \
     scripts/install_s5cmd.sh x64 && \
     scripts/patch_oss_dlc.sh python && \
+    scripts/security_patch.sh fastertransformer && \
```
rohithkrn, line 82: curious about the function of this? I see it's just installing bunch debian packages.

siddvenk: This script updates some of the OS packages that come from the base image (specifically from Ubuntu20.04 in this case). The base container does not get updated too frequently (if at all, the base image we're using here hasn't been updated in 2 months).

If a vulnerability is identified in one of these packages, it's not always updated in the base image. So we use this script to update those packages

PR #527, opened 2023-03-07
auto-scale by neuron cores

lanking520:
## Description ##

Add neuron cores auto scale

### DIFF
--- a/engines/python/setup/djl_python/transformer-neuronx.py
+++ b/engines/python/setup/djl_python/transformer-neuronx.py
@@ -47,7 +47,11 @@ class TransformerNeuronXService(object):
         self.model = AutoModelForCausalLM.from_pretrained(
             self.model_id_or_path, low_cpu_mem_usage=True)
         logging.info(f"Start model conversion to INF2 format...")
-        load_path = os.path.join(tempfile.gettempdir(), self.model_id_or_path)
+        path = os.environ.get("SERVING_DOWNLOAD_DIR")
+        if path:
+            load_path = tempfile.mkdtemp(dir=path, prefix="converted_model")
+        else:
+            load_path = tempfile.mkdtemp()
         dtype = dtypes.to_torch_dtype(amp)
         for block in self.model.model.decoder.layers:
             block.self_attn.to(dtype)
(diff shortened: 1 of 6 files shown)


engines/python/setup/djl_python/transformer-neuronx.py
```diff
@@ -47,7 +46,7 @@ def convert_opt(self, amp):
         self.model = AutoModelForCausalLM.from_pretrained(
             self.model_id_or_path, low_cpu_mem_usage=True)
         logging.info(f"Start model conversion to INF2 format...")
-        load_path = os.path.join(tempfile.gettempdir(), self.model_id_or_path)
+        load_path = os.path.join("/tmp/tn_model", self.model_id_or_path)
```
frankfliu, line 49: Why hard code /tmp?

lanking520: For saving to a static place in temp directory.

frankfliu: I think tempfile.gettempdir() is better than hard code /tmp. if you want use a subfolder, you can use:

```
load_path = os.path.join(tempfile.gettempdir(), "tn_model",  self.model_id_or_path)
```

lanking520: In S3 downloading, the model id or path is already like /tmp/download4577754, this concat failed

frankfliu: Actually, we should respect do the same as how we download the model:
1. try env var: SERVING_DOWNLOAD_DIR first
2. fallback to tempfile.gettempdir()

frankfliu: > In S3 downloading, the model id or path is already like /tmp/download4577754, this concat failed

then we should get the file name from model_id_or_path instead of full path

lanking520: I found a better way

engines/python/setup/djl_python/transformer-neuronx.py
```diff
@@ -47,7 +47,10 @@ def convert_opt(self, amp):
         self.model = AutoModelForCausalLM.from_pretrained(
             self.model_id_or_path, low_cpu_mem_usage=True)
         logging.info(f"Start model conversion to INF2 format...")
-        load_path = os.path.join(tempfile.gettempdir(), self.model_id_or_path)
+        prefix = self.model_id_or_path
+        if os.path.exists(prefix):
+            prefix = "converted_model"
+        load_path = os.path.join(tempfile.gettempdir(), prefix)
```
frankfliu, line 53:
```suggestion
        load_path = os.environ.get("SERVING_DOWNLOAD_DIR")
        if not load_path:
            load_path = tempfile.gettempdir()

        load_path = os.path.join(load_path, prefix)
```

lanking520: Neuron API will try to make dir with given path. If folder exist, it will fail

lanking520: so we have to make the directory unique

frankfliu (approved):

PR #534, opened 2023-03-08
fix encoding for inf2

lanking520:
## Description ##

Use batch encoding with padding for this case.

Tested on inf2.48xlarge


engines/python/setup/djl_python/transformer-neuronx.py
```diff
@@ -151,8 +151,8 @@ def infer(self, inputs):
                     f"{self.batch_size} batch size not equal to {len(input_text)} prompt size"
                 )
             with torch.inference_mode():
-                input_ids = torch.as_tensor(
-                    [self.tokenizer.encode(text) for text in input_text])
+                input_ids = self.tokenizer.batch_encode_plus(
```
frankfliu, line 154: maybe add a comment here that inf2 still requires padding

frankfliu (approved):

PR #543, opened 2023-03-09
Create CacheWorkflowFunction

zachgk:
This creates a simple CacheWorkflowFunction that will cache the results of models (or other WorkflowFunctions).

It will be used as a precursor to a streaming equivalent.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/workflow/function/cache/CacheEngine.java
+++ b/serving/src/main/java/ai/djl/serving/workflow/function/cache/CacheEngine.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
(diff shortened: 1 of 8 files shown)


serving/src/main/java/ai/djl/serving/workflow/function/cache/CacheEngine.java
```diff
@@ -0,0 +1,45 @@
...
+/** A cache that can be used as part of the {@link CacheWorkflowFunction}. */
+public interface CacheEngine {
+
+    /**
+     * Puts an item in the cache.
+     *
+     * @param hash the item hash
+     * @param data the data to store
+     */
+    void put(int hash, Item data);
```
lanking520, line 29: Item is a very limited data format to only support String format. Could we use Serving's `Output` object instead? It build the seamless experience on the data type.

zachgk: Item can be either a String, List of Items, or an Output. So it's a superset of Output's capabilities

PR #551, opened 2023-03-12
[serving] Adds chunked encoding support

frankfliu:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/outputs.py
+++ b/engines/python/setup/djl_python/outputs.py
@@ -51,6 +51,7 @@ class Output(object):
         self.message = message
         self.properties = dict()
         self.content = PairList()
+        self.stream_content = None
 
     def __str__(self):
         d = dict()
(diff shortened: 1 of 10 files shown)


engines/python/setup/djl_python/outputs.py
```diff
@@ -150,13 +154,43 @@ def encode(self) -> bytearray:
...
-        msg += struct.pack('>h', size)
-        for i in range(size):
-            k = self.content.key_at(i)
-            v = self.content.value_at(i)
-            self.write_utf8(msg, k)
-            msg += struct.pack('>i', len(v))
-            msg += v
-
-        return msg
+        if self.stream_content is None:
```
lanking520, line 157: could we isolate this encoding component in to a single class? Maybe in the future it can be extended to support different encoding types

zachgk (approved):
LGTM

PR #560, opened 2023-03-21
[python] Passing arguments to model.py

frankfliu:
## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -201,6 +201,9 @@ public class PyModel extends BaseModel {
                                 + getProperty("gpu.maxWorkers"));
             }
             mpiWorkers = Integer.parseInt(getProperty("gpu.maxWorkers"));
+
+            properties.forEach((k, v) -> pyEnv.addParameter(k, v));
+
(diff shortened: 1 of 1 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -201,13 +201,21 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
                                 + getProperty("gpu.maxWorkers"));
             }
             mpiWorkers = Integer.parseInt(getProperty("gpu.maxWorkers"));
+
+            for (Map.Entry<String, String> entry : properties.entrySet()) {
```
rohithkrn, line 205: nit: we can probably add a private function and re-use at both places.

Also curious if engine name will be part of properties or will it only include params starting with `options.`

rohithkrn (approved):

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
frankfliu: made it a single line statement.

`option.XXX` will be passed to `loadModel(Path, String, Map options)`, some of them will be added to pyEnv.
anything else will be set as `model.properties`, now we add them to pyEnv

PR #568, opened 2023-03-23
[serving] deprecate s3Url and replace it with model_id

frankfliu:
- [serving] support load entryPoint with url
- [serving] deprecate s3Url and replace it with model_id

## Description ##

### DIFF
--- a/serving/docker/partition/partition.py
+++ b/serving/docker/partition/partition.py
@@ -40,13 +40,14 @@ class PartitionService(object):
         self.download_model_from_s3()
 
     def download_model_from_s3(self):
-        if "s3url" not in self.properties:
+        model_id = self.properties.get("model_id")
+        if not model_id or not model_id.startswith("s3://"):
             return
 
         download_dir = os.environ.get("SERVING_DOWNLOAD_DIR",
                                       '/tmp/download/model/')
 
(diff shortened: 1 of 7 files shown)


tests/integration/llm/prepare.py
```diff
@@ -8,13 +8,13 @@
 
 ds_aot_list = {
     "opt-6.7b": {
-        "option.s3url": "s3://djl-llm/opt-6b7/",
+        "option.model_id": "s3://djl-llm/opt-6b7/",
```
lanking520, line 11: let's keep an s3 url option here. To keep this test in the loop

serving/docker/partition/partition.py
```diff
@@ -40,13 +40,14 @@ def __init__(self, props_manager):
         self.download_model_from_s3()
 
     def download_model_from_s3(self):
-        if "s3url" not in self.properties:
+        model_id = self.properties.get("model_id")
+        if not model_id or model_id.startswith("s3://"):
```
sindhuvahinis, line 44: you mean `not model_id.startswith`?

frankfliu: fixed

sindhuvahinis (approved):

PR #573, opened 2023-03-27
Python engine streaming initial support

rohithkrn:
## Description ##

This PR adds initial streaming support for Accelerate, DeepSpeed and transformers-neuronx engines. This implementation supports only batch size = 1. Will add support for batch size > 1 in subsequent PR as that needs some special handling. Tested locally with bloom, gpt and opt models.

Also added a default output formatter to format output which encodes output from the generator as `{"outputs": <output_from_generator>}\n` which will be used unless specified otherwise by the user. The formatter can be configured by the user by providing it as function input to `add_stream_content()` function of `Output()` class.

Next steps:
* Add integration tests
* Add support for client side batching


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -122,6 +126,7 @@ def _parse_properties(self, properties):
             properties.get("tensor_parallel_degree", 1))
         self.low_cpu_mem_usage = properties.get("low_cpu_mem_usage",
                                                 "true").lower() == "true"
+        self.enable_streaming = bool(properties.get("enable_streaming", False))
```
frankfliu, line 129: better call it streaming_token, this not not a real stream case

rohithkrn: `streaming_token` may not be suitable for cases when we support image models.

engines/python/setup/djl_python/huggingface.py
```diff
@@ -125,10 +135,16 @@ def inference(self, inputs):
             input_map = decode(inputs, content_type)
             data = input_map.pop("inputs", input_map)
             parameters = input_map.pop("parameters", {})
+            outputs = Output()
 
+            if self.enable_streaming:
+                stream_generator = StreamingUtils.get_stream_generator("DeepSpeed")
```
lanking520, line 141: strange to use DeepSpeed generator in pure huggingface

rohithkrn: oops, sorry that's a typo. Will update. This worked because both DeepSpeed and HF share the same generator.

engines/python/setup/djl_python/streaming_utils.py
```diff
@@ -0,0 +1,179 @@
...
+            curr_length += 1
+            new_tokens_count += 1
+            stop_generation = StreamingUtils._has_met_stopping_criteria(token_ids, new_tokens_count,
+                                        max_new_tokens, model.config.eos_token_id)
+            yield token_text
+
+
+    @staticmethod
+    @torch.inference_mode()
+    def _transformers_neuronx_stream_generator(model, tokenizer, inputs, **kwargs):
```
lanking520, line 84: add a todo to migrate this to TNX stream solution

rohithkrn: is TNX implementing streaming natively in their repo and that's what we want to migrate to?

lanking520:
Still a few tests need to fix

lanking520 (approved):

PR #588, opened 2023-03-30
Add default handler for AOT

sindhuvahinis:
## Description ##

### DIFF
--- a/serving/docker/partition/properties_manager.py
+++ b/serving/docker/partition/properties_manager.py
@@ -134,6 +134,13 @@ class PropertiesManager(object):
                 if entry_point_file:
                     self.properties['entryPoint'] = 'model.py'
                 else:
-                    raise FileNotFoundError(
-                        f"model.py not found in model path {self.properties_dir}"
-                    )
+                    engine = self.properties['engine']
+                    if engine == "Deepspeed":
+                        entry_point = "djl_python.deepspeed"
+                    elif engine == "FasterTransformer":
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

serving/docker/partition/properties_manager.py
```diff
@@ -134,6 +134,13 @@ def set_and_validate_entry_point(self):
...
                 else:
-                    raise FileNotFoundError(
-                        f"model.py not found in model path {self.properties_dir}"
-                    )
+                    engine = self.properties['engine']
+                    if engine == "Deepspeed":
+                        entry_point = "djl_python.deepspeed"
+                    elif engine == "FasterTransformer":
+                        entry_point = "djl_python.fastertransformer"
+                    else:
```
frankfliu, line 142:
```
else if self.properties.contains("model_id"):
    entry_point = "djl_python.huggingface"
```

frankfliu: How can we handle transformers-neuronx case?

sindhuvahinis: Can we do AOT in transformers-neuronx case?

frankfliu: Not today, but sometime next week.
We need have a way to detect the script is running on inf2 docker image and on inf2 machine.

sindhuvahinis: Ah I see, Let me do that in the next PR. Also, no AOT for HF as of now.

PR #596, opened 2023-04-01
[serving] Avoid download from s3 multiple time

frankfliu:
- [serving] Refactor per model configuration initialization
- [serving] Avoid download from s3 multiple time

## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -181,24 +181,6 @@ public class PyModel extends BaseModel {
         }
         pyEnv.setEntryPoint(entryPoint);
 
-        Map<String, String> initParameters = pyEnv.getInitParameters();
-        String s3Url = initParameters.get("s3url");
-        String modelId = pyEnv.getInitParameters().get("model_id");
-        if (s3Url != null) {
-            // s3url is deprecated, use model_id instead
-            if (modelId != null) {
-                throw new IllegalArgumentException("model_id and s3url could not both set!");
-            }
-            modelId = s3Url;
-        }
-        if (modelId != null && modelId.startsWith("s3://")) {
-            logger.info("S3 url found, start downloading from {}", modelId);
-            String downloadDir = getDownloadDir().toString();
-            downloadS3(modelId, downloadDir);
-            // point model_id to download directory
-            pyEnv.addParameter("model_id", downloadDir);
-        }
-
         if (pyEnv.isMpiMode()) {
             int partitions = pyEnv.getTensorParallelDegree();
             if (partitions == 0) {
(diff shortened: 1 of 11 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -181,24 +181,6 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
...
-        String modelId = pyEnv.getInitParameters().get("model_id");
-        if (s3Url != null) {
-            // s3url is deprecated, use model_id instead
-            if (modelId != null) {
-                throw new IllegalArgumentException("model_id and s3url could not both set!");
-            }
-            modelId = s3Url;
-        }
-        if (modelId != null && modelId.startsWith("s3://")) {
-            logger.info("S3 url found, start downloading from {}", modelId);
```
lanking520, line 195: Where do you migrate this logic? In the API package?

frankfliu: This PR depends on https://github.com/deepjavalibrary/djl-serving/pull/594

the S3 down is moved into ModelInfo

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -648,11 +658,19 @@ void checkAvailableMemory(Device device) throws IOException {
...
+            if (downloadS3Dir != null) {
+                try (Stream<Path> walk = Files.walk(downloadS3Dir)) {
+                    requiredMemory +=
+                            walk.filter(Files::isRegularFile)
+                                    .mapToLong(ModelInfo::getFileSize)
+                                    .sum();
+                }
+            }
+            // estimate the memory to be 1.2x of file size
+            requiredMemory = requiredMemory * 12 / 10;
```
lanking520, line 673: this memory estimation should only apply on CPU, not GPU. Or there could be some issues. For now, let's not apply a factor of 1.2, just follow it's original size. 

Apart from that, you will never know how user plan to to load these files. For example, if customer do `load_with_low_cpu_memory` option on HuggingFace. The peak model memory will remains at per checkpoint size. So I would suggest to now estimate by this.

lanking520 (changes requested):

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
frankfliu: The current estimation doesn't apply to LMI use case.

For a real production usage, user should provide a more accurate memory value in the model. I think it's better to be more conservative if user didn't provide the memory footprint.

lanking520: This will break many default use cases. All of our notebook has no preset estimated memory. There will be definitely bumpy for many customers to try especially CPU space is not enough.

For example, our AOT will partition model to 46GB 4 shards. If user didn't set memory, it will break here.

lanking520: Although for AWS there is no case CPU memory is less than GPU memory, but many desktop customer may have 8GB CPU memory with 3060 (16GB) GPU card

lanking520: DeepSpeed, HuggingFace, FasterTransformer has mechanism to load checkpoint by checkpoint

frankfliu: the estimated memory will be 0 for LMI case (tp > 1). This won't break our notebook.

lanking520: What if the S3 downloaded data is a lookup table and only used when needed? What if not all data downloaded by S3 is used all in the same time. Like parquet file, why should I worry the total size?

lanking520: It will break many default "work" use cases through forcibly setting a CPU memory to bypass this line

PR #606, opened 2023-04-06
Support text2text-generation task in deepspeed

siddvenk:
## Description ##
This PR covers 2 things:
- Support text2text-generation task in the DeepSpeed default handler
- Remove restriction on model type. With DeepSpeed's new AutoTP functionality, models that do not support kernel injection can be used with the injection policy being determined by DeepSpeed at runtime. What this means for us is we can support many more model types without having to determine and hard code the injection policy for all model types. See https://www.deepspeed.ai/tutorials/automatic-tensor-parallelism/ for more details.

## Testing ##
Tested this on g5.12xl with flant5-xxl model using tp2

Below is a comparison between accelerate and deepspeed for this model. 4 requests with batch sizes (1,2,4,8):
```
DeepSpeed
Time for inference with batch size 1 is 3.2054641246795654s
Time for inference with batch size 2 is 1.2419123649597168s
Time for inference with batch size 4 is 1.0975754261016846s
Time for inference with batch size 8 is 5.2593092918396s

GPU Mem usage in MiB [7431, 7433, 7435, 7435]
```

```
Accelerate
Time for inference with batch size 1 is 4.627055644989014s
Time for inference with batch size 2 is 2.052828550338745s
Time for inference with batch size 4 is 1.58766770362854s
Time for inference with batch size 8 is 7.7868125438690186s

GPU Mem usage in MiB [6597, 6607, 6611, 6615]
```

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -16,6 +16,7 @@ import json
 import torch
 from transformers import (AutoConfig, PretrainedConfig, AutoTokenizer,
                           AutoModelForCausalLM,
+                          AutoModelForSeq2SeqLM,
                           AutoModelForSequenceClassification,
                           AutoModelForQuestionAnswering, AutoModelForMaskedLM,
                           AutoModelForTokenClassification, pipeline,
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

lanking520:
could you also add support to huggingface.py? for T5 support

siddvenk:
> could you also add support to huggingface.py? for T5 support

I think this should work with huggingface already - I just used the default handler to test it in the above comparison.

PR #610, opened 2023-04-08
update fastertransformer to follow huggingface parameters

lanking520:
## Description ##

Unified the the handler experience

### DIFF
--- a/engines/python/setup/djl_python/fastertransformer.py
+++ b/engines/python/setup/djl_python/fastertransformer.py
@@ -12,12 +12,42 @@
 # the specific language governing permissions and limitations under the License.
 
 import fastertransformer as ft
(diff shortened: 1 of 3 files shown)


frankfliu (approved):

engines/python/setup/djl_python/fastertransformer.py
```diff
@@ -12,20 +12,51 @@
...
 import fastertransformer as ft
+from transformers import AutoConfig
 from djl_python import Input, Output
 import logging
 from typing import Optional
 
 
 class FasterTransformerService(object):
 
+    t5_default_args = dict(inputs_embeds=None,
```
rohithkrn, line 23: where are these default args used?

PR #618, opened 2023-04-12
Remove hanging streaming test case

rohithkrn:
## Description ##

Seq length 32 test case is hanging for unknown reason and the test job gets cancelled - https://github.com/deepjavalibrary/djl-serving/actions/runs/4679798912/jobs/8290368439

Same case works locally though.

Removing the test for now to unblock the tests.


lanking520:
Could you reproduce this error?

rohithkrn:
> Could you reproduce this error?

No, it works fine locally.

lanking520 (approved):

PR #623, opened 2023-04-13
Infer lmi engine

siddvenk:
## Description ##

This PR adds support for inferring the specific Python engine to use. This logic should only be called when users don't specify engine in `serving.properties`, and we can reasonably assume the model is intended to be run with a Python backend.

Use Cases covered:
- User only specifies HF_MODEL_ID environment variable (no serving properties, model artifacts, or user code)
- User provides serving properties with model id (either hf hub id or s3 url), but does not specify engine
- model_id is not provided in any form, but model artifacts are present in model_dir

This does not support the use case where users provide their own code that is expected to be invoked via the hugging face inference toolkit. There's no special error handling for that now, it will just fail when the PyProcess tries to load the handler and invoke it.

The logic for inferring the backend is largely copied over from the logic in the PySDK.

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -41,7 +41,6 @@ import ai.djl.util.cuda.CudaUtils;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 
-import jdk.jshell.execution.Util;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -51,7 +50,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.lang.management.MemoryUsage;
 import java.net.URI;
-import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -619,7 +617,11 @@ public final class ModelInfo<I, O> {
         if (huggingFaceHubModelId == null && Files.isRegularFile(modelDir.resolve("config.json"))) {
             modelConfigUri = modelDir.resolve("config.json").toUri();
         } else {
-            modelConfigUri = URI.create("https://huggingface.co/" + huggingFaceHubModelId + "/raw/main/config.json");
+            modelConfigUri =
+                    URI.create(
+                            "https://huggingface.co/"
+                                    + huggingFaceHubModelId
+                                    + "/raw/main/config.json");
         }
 
         JsonObject modelConfig;
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -726,8 +726,11 @@ public final class ModelInfo<I, O> {
                         new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))) {
             return JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
         } catch (IOException e) {
-            logger.warn("Could not read model config from {}", modelConfigUri, e);
-            return null;
+            logger.error(
+                    "The huggingface hub model_id {} is not valid. Please verify this is a valid"
+                            + " model_id",
+                    modelId);
+            throw e;
         }
     }
 
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -54,7 +54,11 @@ import java.net.URL;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.util.*;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Properties;
+import java.util.Scanner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -67,6 +71,18 @@ public final class ModelInfo<I, O> {
     private static final Logger logger = LoggerFactory.getLogger(ModelInfo.class);
 
     private static final Pattern PATTERN = Pattern.compile("MemAvailable:\\s+(\\d+) kB");
+
+    private static final List<String> DEEPSPEED_MODELS =
+            List.of(
+                    "roberta",
+                    "xlm-roberta",
+                    "gpt2",
+                    "bert",
+                    "gpt_neo",
+                    "gptj",
+                    "opt",
+                    "gpt_neox",
+                    "bloom");
 
     private transient String id;
     private String version;
@@ -580,11 +596,6 @@ public final class ModelInfo<I, O> {
             return "XGBoost";
         } else {
             try {
-                return inferLMIEngine();
-            } catch (IOException e) {
-                logger.error("Failed to find model in huggingface hub", e);
-            }
-            try {
                 if (Utils.getCurrentEpoch(modelDir, prefix) >= 0) {
                     // Assume this is DJL model
                     return Engine.getDefaultEngineName();
@@ -592,21 +603,29 @@ public final class ModelInfo<I, O> {
             } catch (IOException e) {
                 logger.warn("Failed search parameter files in folder: " + modelDir, e);
             }
+            engine = inferLMIEngine();
+            if (engine != null) {
+                return engine;
+            }
         }
         throw new ModelNotFoundException("Failed to detect engine of the model: " + modelDir);
     }
 
-    private String inferLMIEngine() throws IOException {
-        String huggingFaceHubModelId = System.getenv("HF_MODEL_ID");
-        String huggingFaceTask = System.getenv("HF_TASK");
+    private String inferLMIEngine(){
+        String huggingFaceHubModelId = Utils.getEnvOrSystemProperty("HF_MODEL_ID");
+        String huggingFaceTask = Utils.getEnvOrSystemProperty("HF_TASK");
         String modelConfigUrl =
                 "https://huggingface.co/" + huggingFaceHubModelId + "/raw/main/config.json";
+
         JsonObject modelConfig;
         try (InputStream is = new URL(modelConfigUrl).openStream();
                 BufferedReader reader = new BufferedReader(new InputStreamReader(is))) {
-           modelConfig =
-                    JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
+            modelConfig = JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
+        } catch (IOException e) {
+            logger.error("Could not read model config for {} from huggingface hub", huggingFaceHubModelId, e);
+            return null;
         }
+
         String modelType = modelConfig.get("model_type").getAsString();
         long numAttentionHeads;
         if (modelConfig.has("num_attention_heads")) {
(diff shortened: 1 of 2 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+            prop.put("option.task", huggingFaceTask);
+        }
+
+        if ("stable-diffusion".equals(modelType)) {
+            // TODO: Move this from hardcoded to deduced in PyModel
+            prop.put("option.entryPoint", "djl_python.stable-diffusion");
+            return "DeepSpeed";
+        }
+
+        if (!isTensorParallelSupported(numAttentionHeads, tensorParallelDegree)) {
```
lanking520, line 670: Maybe DS or FT would have some mechanism from their end to decide how to do model sharding. I would suggest to not check this

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+        }
+        return "Python";
+    }
+
+    private URI generateHuggingFaceConfigUri(String modelId) throws ModelException, IOException {
+        URI configUri = null;
+        if (modelId != null && modelId.startsWith("s3://")) {
+            // This is definitely suboptimal, but for the majority of cases we need to download this
+            // s3 model eventually, so it is not the worst thing to download it now.
+            downloadS3();
```
lanking520, line 687: did you prevent from model redownloading?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+                    modelId);
+            throw e;
+        }
+    }
+
+    private boolean isFasterTransformerRecommended(String modelType) {
+        return "t5".equals(modelType);
+    }
+
+    private boolean isDeepSpeedRecommended(String modelType) {
```
lanking520, line 741: we probably want to build mechanism to check if deepspeed is installed or not.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -65,6 +73,18 @@
 
     private static final Pattern PATTERN = Pattern.compile("MemAvailable:\\s+(\\d+) kB");
 
+    private static final List<String> DEEPSPEED_MODELS =
+            List.of(
```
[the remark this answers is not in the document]
siddvenk: I have t5 as part of the `isFasterTransformerRecommended` method. So far it's the only model I know of we recommend FT over DS for. This list is just for DeepSpeed.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+    }
+
+    private String inferLMIEngine() throws ModelException, IOException {
+        // MMS/Torchserve
+        if (Files.isDirectory(modelDir.resolve("MAR-INF"))) {
+            return "Python";
+        }
+        JsonObject modelConfig = getHuggingFaceModelConfig();
+        if (modelConfig == null) {
+            return "Python";
```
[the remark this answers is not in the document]
siddvenk: This would be the pure Python mode case where user is using DJLServing to run a generic python backend, which is valid. One scenario would be where user just provides a model.py file and maybe a .bin file for the checkpoint. 

In the `getHuggingFaceModelConfig` method, if we have a model_id (either s3url, option.model_id, or HF_MODEL_ID) but we can't load the config for that model_id we throw an exception. For that scenario I think throwing the exception makes sense since if we can't load the config, the python backend will also fail. But we could change that to generic python backend use case.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: the `downloadS3` should prevent that already. We check to see if the tmp/<hash> exists before calling s3

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: We would need something for FT as well right?

I was thinking of this in the context of the HuggingFaceModel from SM PySDK. With separate images for DS and FT, we would be locked into an environment already by the time we're here...

You're right this needs to be reworked...

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: At least for DS, they will throw an exception if this check fails. But really the only practical examples of this we have seen is gpt2-xl.

In the future it's possible that DS and FT change that behavior and can actually accommodate such a model. At that point this method would become incorrect.

I can remove this, since it's going to be validated by the engine anyways. But the benefit of doing it this way is that we don't recommend say gpt2-xl to run with DeepSpeed with TP when we know it won't work.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
         throw new ModelNotFoundException("Failed to detect engine of the model: " + modelDir);
     }
 
+    private boolean isPythonModel(String prefix) {
+        return Files.isDirectory(modelDir.resolve("MAR-INF"))
+                || Files.isRegularFile(modelDir.resolve("model.py"))
+                || Files.isRegularFile(modelDir.resolve(prefix + ".py"))
+                || Utils.getEnvOrSystemProperty("HF_MODEL_ID") != null
+                || Files.isRegularFile(modelDir.resolve("config.json"))
+                || prop.containsKey("option.s3url")
```
frankfliu, line 618: other engine can support model_id and s3url as well.
If user defined a option.model_id, we can assume they can add engine as well.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+        }
+        URI modelConfigUri = generateHuggingFaceConfigUri(modelId);
+        if (modelConfigUri == null) {
+            return null;
+        }
+        try (InputStream is = modelConfigUri.toURL().openStream();
+                BufferedReader reader =
+                        new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))) {
+            return JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
+        } catch (IOException e) {
```
frankfliu, line 728: We should also catch JsonSyntaxException as well

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +609,143 @@ private String inferEngine() throws ModelNotFoundException {
...
+            modelId = prop.getProperty("option.s3url");
+        }
+        URI modelConfigUri = generateHuggingFaceConfigUri(modelId);
+        if (modelConfigUri == null) {
+            return null;
+        }
+        try (InputStream is = modelConfigUri.toURL().openStream();
+                BufferedReader reader =
+                        new BufferedReader(new InputStreamReader(is, StandardCharsets.UTF_8))) {
+            return JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
```
frankfliu, line 727: Better create a class to hold the config (we only need define the field we care about)

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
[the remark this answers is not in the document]
lanking520: You may also need a list for fastertransformer.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
lanking520: This mechanism should execute in a standalone mode, not relate to any server. Which is we should be able to detect if certain package is installed or not

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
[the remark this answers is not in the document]
siddvenk: done

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: done

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: got it - i removed those checks here.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: added.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: done

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -1064,4 +1078,34 @@ public enum Status {
         READY,
         FAILED
     }
+
+    //  This represents  the config of huggingface models NLP models as well
+    // as the config of diffusers models. The config is different for both, but for
+    // now we can leverage a single class since we don't need too much information from the config.
+    static class HuggingFaceModelConfig {
```
frankfliu, line 1085:
```suggestion
    static final class HuggingFaceModelConfig {
```

wlm/src/test/java/ai/djl/serving/wlm/ModelInfoTest.java
```diff
@@ -244,4 +247,75 @@ public void testInitModel() throws IOException, ModelException {
...
+        if (!Files.exists(deepspeedLocation)) {
+            Files.createDirectories(deepspeedLocation);
+            deepspeedExisted = false;
+        }
+        Path fastertransformerLocation = Paths.get("/usr/local/backends/fastertransformer");
+        boolean fastertransformerExisted = true;
+        if (!Files.exists(fastertransformerLocation)) {
+            Files.createDirectories(fastertransformerLocation);
+            fastertransformerExisted = false;
+        }
```
siddvenk, line 267: I added this for the unit tests, but not a big fan of it. We can instead test this with some integration tests?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -724,28 +711,55 @@ private JsonObject getHuggingFaceModelConfig() throws ModelException, IOExceptio
...
-            return JsonUtils.GSON.fromJson(reader, JsonElement.class).getAsJsonObject();
-        } catch (IOException e) {
-            logger.error(
-                    "The huggingface hub model_id {} is not valid. Please verify this is a valid"
-                            + " model_id",
-                    modelId);
-            throw e;
+            Gson gson =
+                    JsonUtils.builder()
+                            .setFieldNamingPolicy(FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES)
```
frankfliu, line 716: Why we need this since we already use @SerializedName?

siddvenk: left over from my testing - this should be removed good catch

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -588,6 +611,149 @@ private String inferEngine() throws ModelNotFoundException {
...
+            downloadS3();
+            if (Files.isRegularFile(downloadS3Dir.resolve("config.json"))) {
+                configUri = downloadS3Dir.resolve("config.json").toUri();
+            } else if (Files.isRegularFile(downloadS3Dir.resolve("model_index.json"))) {
+                configUri = downloadS3Dir.resolve("model_index.json").toUri();
+            }
+        } else if (modelId != null) {
+            prop.put("option.modelId", modelId);
+            configUri = URI.create("https://huggingface.co/" + modelId + "/raw/main/config.json");
+            HttpURLConnection configUrl = (HttpURLConnection) configUri.toURL().openConnection();
```
frankfliu, line 681: We can consider use "OPTION" method instead of GET. I think it should work

frankfliu (approved):

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
siddvenk: I'll explore that and add as follow up if i get it to work.

PR #645, opened 2023-04-19
[docker] Uses tranditional pip command to install specific cuda version of pytorch

frankfliu:
## Description ##


lanking520:
This doesn't make a difference. I found the version regression issue before. It is triggered by open ai. Triton dependency

PR #657, opened 2023-04-24
[serving] Fixes huggingface integration test

frankfliu:
## Description ##


lanking520 (commented):
This is not a right "fix". In CI testing, we are trying our best to mimic the real use cases. Customer will not create this test folder in SageMaker. What we need to do is pointing to the right endpoint name to call

lanking520:
@frankfliu @tosterberg

lanking520:
We can add an endpoint option to override the default test endpoint

lanking520:
Then let's change our test to allow different endpoint names?

PR #658, opened 2023-04-24
Remove streaming from no-code test

siddvenk:
## Description ##

The no-code test client should not expect streaming output. With no code, we don't provide any serving.properties which means by default streaming is disabled. client should not try to read response in streaming fashion.


frankfliu (approved):

sindhuvahinis (approved):

lanking520:
Maybe we need to think about whether we need to make streaming output per request basis. If I am a customer, I would probably use non-streaming in short seq but leave streaming on long seq. I would suggest we leave it as a backlog item.

PR #661, opened 2023-04-25
[serving] Adds workflow model loading for SageMaker

frankfliu:
## Description ##


serving/src/main/java/ai/djl/serving/workflow/WorkflowDefinition.java
```diff
@@ -145,23 +190,9 @@ public static WorkflowDefinition parse(URI uri, Reader input) {
      */
     public Workflow toWorkflow() throws BadWorkflowException {
         if (models != null) {
-            WlmConfigManager wlmc = WlmConfigManager.getInstance();
             for (Entry<String, ModelInfo<Input, Output>> emd : models.entrySet()) {
                 ModelInfo<Input, Output> md = emd.getValue();
                 md.setId(emd.getKey());
-                md.setQueueSize(firstValid(md.getQueueSize(), queueSize, wlmc.getJobQueueSize()));
```
zachgk, line 152: Is there a reason we are removing these parameters from the workflow?

frankfliu: 1. global settings is just the default value, per model setting is override the settings in `serving.properties`. this is a bit confusing
1. In most cases (especially on GPU), each model will have different per model settings
2. We only covered a subset of per model settings.

zachgk: I would still lean towards keeping it.

1. It's a basic system of more specific rules override less specific. `Global < workflow settings < model serving.properties < workflow model settings`
2. The workflow has both settings for all models in the workflow and for particular models
3. Do you mean `loadOnDevices`? It might be worth adding it, then.

I think having this is still useful. When importing models defined externally (like in model hub), this still allows defining some model properties without having to do something difficult to introduce the serving.properties.

But, I can see the argument towards removing it and I don't think it's too important. I'll just approve it and leave it up to you

zachgk (approved):

PR #675, opened 2023-04-28
Create model dir to prevent issues with no code experience in SageMaker

siddvenk:
## Description ##

will push to dlc branch  as well once approved


sindhuvahinis:
I tested no code mode in PySDK. I did not face any issues. When will I able to reproduce this?

lanking520 (approved):

PR #676, opened 2023-04-28
Don't mount model dir for no code tests

siddvenk:
## Description ##

Test run https://github.com/deepjavalibrary/djl-serving/actions/runs/4834079393/jobs/8614905302. 

Although it failed, the issue was with the model name when it was registered (default is model, we expect test). I updated the workflow to match the expected model name on the client side.

Functionally, the test run showed these changes worked.

### DIFF
--- a/.github/workflows/llm_integration.yml
+++ b/.github/workflows/llm_integration.yml
@@ -738,8 +738,8 @@ jobs:
         run: |
           echo -en "HF_MODEL_ID=nomic-ai/gpt4all-j" > docker_env
           mkdir -p models/test
-          ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG $PWD/models deepspeed \
-          serve -m test=file:/opt/ml/model/test/
+          ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG no_code deepspeed \
+          serve -m test=file:/opt/ml/model/
           python3 llm/client.py huggingface no-code/nomic-ai/gpt4all-j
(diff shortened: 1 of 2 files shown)


.github/workflows/llm_integration.yml
```diff
@@ -738,8 +738,8 @@ jobs:
         run: |
           echo -en "HF_MODEL_ID=nomic-ai/gpt4all-j" > docker_env
           mkdir -p models/test
-          ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG $PWD/models deepspeed \
-          serve -m test=file:/opt/ml/model/test/
+          ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG no_code deepspeed \
+          serve -m test=file:/opt/ml/model/
```
lanking520, line 742: remove -m, we also don't need this

siddvenk: if we remove this, is there a way to specify the model name as test? I initially removed this and then ran into issues with client.py expecting the model name to be `test`, but by default the model will be called `model`.

siddvenk: We can definitely update client.py, but I thought this was easier

lanking520: ok, let's leave this now

lanking520 (approved):

.github/workflows/llm_integration.yml
frankfliu: We need specify the model name here. we don't have a way to call management API without model name (this is not supported by SageMaker).

PR #678, opened 2023-04-30
AOT upload checkpoints tests

sindhuvahinis:
Adding tests cases for some for SM Training Job.
- Test AOT with and without passing any model_dir argument 
- Invoke AOT by using the command `train`
- Test cases for upload the partitioned checkpoints to S3
- Adding a few testcases for DeepSpeed with Default handler

### DIFF
--- a/.github/workflows/llm_integration.yml
+++ b/.github/workflows/llm_integration.yml
@@ -428,6 +428,9 @@ jobs:
           python-version: '3.10.x'
       - name: Install pip dependencies
         run: pip3 install requests numpy
+      - name: Install s5cmd
+        working-directory: serving/docker
+        run: sudo scripts/install_s5cmd.sh x64
       - name: Build container name
         run: ./serving/docker/scripts/docker_name_builder.sh fastertransformer ${{ github.event.inputs.djl-version }}
       - name: Download models and dockers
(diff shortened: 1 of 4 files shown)


tests/integration/llm/prepare.py
```diff
@@ -25,13 +25,32 @@
         "option.s3url": "s3://djl-llm/opt-6b7/",
         "option.tensor_parallel_degree": 4,
         "option.task": "text-generation",
-        "option.dtype": "float16"
+        "option.dtype": "float16",
+        "option.save_mp_checkpoint_path": "/opt/ml/input/data/training/partition-test"
     },
-    "gpt-neox-20b": {
```
lanking520, line 30: why swapping the model?

sindhuvahinis: Did not swap it. This was added in prepare.py but never used in llm_intergation.yaml.

sindhuvahinis: Do you want to add gptneox as well in the test cases?

lanking520 (approved):

.github/workflows/llm_integration.yml
```diff
@@ -456,9 +460,10 @@ jobs:
           sudo rm -rf models
           python3 llm/prepare.py fastertransformer_raw_aot t5-small
           ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG $PWD/models deepspeed \
-          partition --model-dir /opt/ml/model/test/
+          train
```
siddvenk, line 463: It looks like we support both `train` and `partition` as commands that invoke the partitioning. 

I'm assuming `train` is needed for SageMaker. Once we add SageMaker tests, can we keep the command consistent here?

siddvenk (approved):

PR #690, opened 2023-05-05
[ci] make performance tests run in parallel

tosterberg:
## Description ##

Update the performance pipeline to run all tests in parallel, uses 4 instances total, 3x g5.12x and 1x g5.48x. Additionally, shortened the llm container setup timeout, since some models are ready earlier than the original 5 minute timer.

### DIFF
--- a/tests/integration/lmic_test_builder.py
+++ b/tests/integration/lmic_test_builder.py
@@ -267,7 +267,7 @@ class LMITestRunner:
             file = open("llm/metrics.log", "r")
             metrics = re.sub("'", r'"', file.readline())
             command = f'aws cloudwatch put-metric-data --namespace "LMIC_performance_{sequence["engine"]}" ' \
-                      f'--region "us-east-1" --metric-data "{metrics}"'
+                      f'--region "us-east-1" --metric-data \'{metrics}\''
         logging.info(command)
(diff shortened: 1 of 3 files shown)


tests/integration/lmic_test_builder.py
```diff
@@ -267,7 +267,7 @@ def log_metrics(self, sequence):
             file = open("llm/metrics.log", "r")
             metrics = re.sub("'", r'"', file.readline())
             command = f'aws cloudwatch put-metric-data --namespace "LMIC_performance_{sequence["engine"]}" ' \
-                      f'--region "us-east-1" --metric-data "{metrics}"'
+                      f'--region "us-east-1" --metric-data \'{metrics}\''
```
siddvenk, line 270: should `\'{metrics}\'` be `\"{metrics}\"` ?

tosterberg: The double quotes weren't parsing properly, so we moved it over to single quotes and confirmed that it was working properly for posting the metrics json in cloudwatch. Just needed the escape character since we were in a f'' string.

siddvenk (approved):

sindhuvahinis (approved):

PR #693, opened 2023-05-08
[console] Fixes log file charset issue

frankfliu:
Fixes: https://github.com/deepjavalibrary/djl/issues/2589

## Description ##

### DIFF
--- a/plugins/management-console/src/main/java/ai/djl/serving/console/ConsoleRequestHandler.java
+++ b/plugins/management-console/src/main/java/ai/djl/serving/console/ConsoleRequestHandler.java
@@ -43,9 +43,11 @@ import io.netty.util.internal.StringUtil;
 
 import org.apache.commons.compress.utils.Charsets;
 
+import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
 import java.io.RandomAccessFile;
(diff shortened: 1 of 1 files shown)


siddvenk (approved):

plugins/management-console/src/main/java/ai/djl/serving/console/ConsoleRequestHandler.java
```diff
@@ -425,19 +427,31 @@ private String getLastLineText(File file, int lines) {
...
-                builder.append(c);
+                bos.write(c);
                 fileLength = fileLength - pointer;
             }
-            builder.reverse();
         } catch (IOException e) {
             throw new InternalServerException("Failed to read log file.", e);
         }
-        return builder.toString();
+        return new String(reverse(bos.toByteArray()), StandardCharsets.UTF_8);
```
siddvenk, line 436: I'm definitely missing something here, but why does`bos.toString(StandardCharsets.UTF_8)` not work for us here? I'm guessing it's the reversal of the array, but i'm also confused why that's needed.

siddvenk: oh i see, i exanded the above code and we're reading in reverse. makes sense.

PR #698, opened 2023-05-10
Add support for streaming Seq2Seq models

rohithkrn:
## Description ##

Add support for streaming Seq2Seq models. Tested with t5 large model.

### DIFF
--- a/engines/python/setup/djl_python/streaming_utils.py
+++ b/engines/python/setup/djl_python/streaming_utils.py
@@ -13,7 +13,9 @@ from transformers import (
 class StreamingUtils:
 
     DEFAULT_MAX_NEW_TOKENS = 50
-    SUPPORTED_MODEL_ARCH_SUFFIXES = ("CausalLM", "GPT2LMHeadModel")
+    SUPPORTED_MODEL_ARCH_SUFFIXES_CAUSAL_LM = ("CausalLM", "GPT2LMHeadModel")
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/streaming_utils.py
```diff
@@ -32,62 +35,105 @@ def get_stream_generator(execution_engine: str):
...
+                decoder_input_ids = decoder_input_ids * unfinished_sequences + tokenizer.pad_token_id * unfinished_sequences.logical_not(
+                )
+                encoder_last_hidden_state = [outputs.encoder_last_hidden_state]
+                token_text = tokenizer.batch_decode(decoder_input_ids)
 
             stop_generation = StreamingUtils._has_met_stopping_criteria(
-                not_eos_token_ids, new_tokens_count, max_new_tokens)
+                unfinished_sequences, new_tokens_count, max_new_tokens)
+
+            if stop_generation:
```
lanking520, line 135: will we miss the last token from the tokenizer?

rohithkrn: good catch, I bought this check to the bottom to prevent sending end of sentence token if that's what is causing the generation to stop but it is incorrect as it will prevent sending last token if max new tokens condition is met. Reverted the change.

lanking520 (approved):

PR #705, opened 2023-05-11
Add SageMaker integration test

siddvenk:
## Description ##

Adding SageMaker Endpoint Testing. This is still a work in progress, but the changes here do at least cover the basic use cases we expose through the Python SDK.
- Single Model Endpoint Tests using all backends + stable diffusion
- No Code experience using HuggingFace Model and model id through env vars

Future work:
- Add MME testing
- Testing with custom code/serving.properties
- Support testing on nightly images (the major testing here should cover our released images, but before new releases we can automate testing with the new containers)

Notes:
- I've split the workflow into two parallel tracks, each having access to a single ml.g5.12xlarge instance. I'm not sure if there is a better way to parallelize, but for now this works.
- Support for testing our own changes to PySDK that live in a different fork/branch.


.github/workflows/sagemaker-integration.yml
```diff
@@ -0,0 +1,148 @@
...
+          echo "sleep 10 seconds to allow endpoint deletion"
+          sleep 10
+      - name: Test opt-1.3b
+        if: success() || failure()
+        working-directory: tests/integration
+        run: |
+          python3 llm/sagemaker-endpoint-tests.py opt-1-3-b djl
+          echo "sleep 10 seconds to allow endpoint deletion"
+          sleep 10
+      - name: Test flan-t5-xxl
```
lanking520, line 87: we need a failsafe at the end of all tests. If in any case the endpoint was not shutting down, we need a thanos script to kill everything. Or let's implement an AWS Cops for SageMaker in our lambda function to kill all active port nightly

tests/integration/llm/sagemaker-endpoint-tests.py
```diff
@@ -0,0 +1,176 @@
...
+}
+
+HUGGING_FACE_NO_CODE_CONFIGS = {
+    "gpt-neo-2-7-b": {
+        "env": {
+            "HF_MODEL_ID": "EleutherAI/gpt-neo-2.7B",
+            "TENSOR_PARALLEL_DEGREE": "1",
+        },
+        "image_uri":
+        sagemaker.image_uris.retrieve(framework="djl-deepspeed",
```
lanking520, line 72: instead of testing public image, maybe we can point to our Alpha container repo to fetch the image there.

lanking520: So the test will cover the case when we did CVEs

.github/workflows/sagemaker-integration.yml
siddvenk: I'll add a script for this, I think we can use a sagemaker.list_endpoints call and delete all the endpoints that are returned.

tests/integration/llm/sagemaker-endpoint-tests.py
siddvenk: We can add a flag here to test on the nightly image, but I'm not sure if that should be the default. 

In my mind, this test suite can serve two purposes:
- Testing new changes in the Python SDK against our released DLCs
- Testing existing Python SDK against our nightly DLCs

I'll modify the workflow to include both options, but let's take that as a follow up. I'm not able to verify the workflow dispatch functionality here until the workflow exists in master branch.

.github/workflows/sagemaker-integration.yml
lanking520: But you don't want to destroy the other endpoint running concurrently right?

sindhuvahinis:
@siddvenk Are you using yapf version 0.32.0 (from brew)? My formatter giving me different results?

siddvenk:
> @siddvenk Are you using yapf version 0.32.0 (from brew)? My formatter giving me different results

I'm just using `./gradlew formatPython`

sindhuvahinis:
> 

Yes Sid, that one seems to use yapf in our machine. 
```
"bash", "-c", "find . -name '*.py' -not -path '*/.gradle/*' -not -path '*/build/*' -print0 | xargs -0 yapf --in-place"
```

.github/workflows/sagemaker-integration.yml
siddvenk: Right, I think this would be a separate workflow or just a lambda that runs everyday in our account. I don't think having a catch all delete endpoints in this workflow will work for all cases. Sometimes the endpoint stays in the creating stage for more than an hour before it fails. I don't think we can delete endpoints that are in creation phase (even if we know that they will eventually fail).

We need this for multiple accounts/regions, so building something that works for both alpha/gamma in all regions is needed.

sindhuvahinis (approved):

PR #712, opened 2023-05-12
Refactor logic for supporting HF_MODEL_ID to support MME use case

siddvenk:
## Description ##

Will move out of draft once tested.

This PR fixes a couple issues:
- Right now, MME is broken through the docker containers due to creation of /opt/ml/model path (model store). The startup of the model server fails when no models are present
- LMI Engine Selection is also incomplete. We should support auto detection of engine when either HF_MODEL_ID env var provided, or option.model_id is present in serving.properties.

To fix these issues, we create model properties when HF_MODEL_ID is present in a tmp model store, and remove the creation of model store from the docker container.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/ModelServer.java
+++ b/serving/src/main/java/ai/djl/serving/ModelServer.java
@@ -50,17 +50,19 @@ import org.apache.commons.cli.ParseException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.FileOutputStream;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
+import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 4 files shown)


serving/docker/Dockerfile
```diff
@@ -13,8 +13,7 @@ FROM ubuntu:20.04 AS base
 ARG djl_version=0.23.0~SNAPSHOT
 
 COPY scripts scripts/
-RUN mkdir -p /opt/djl/conf && \
```
frankfliu, line 16: I think empty model_store directory is a validate use case. We can keep the folder here. I will create a separate PR to fix empty model_store bug.

serving/src/main/java/ai/djl/serving/ModelServer.java
```diff
@@ -449,4 +462,28 @@ private static void printHelp(String msg, Options options) {
...
         formatter.printHelp(msg, options);
     }
+
+    private String createHuggingFaceModel(String modelId) throws IOException {
+        String hash = Utils.hash(modelId);
+        String downloadDir = Utils.getenv("SERVING_DOWNLOAD_DIR", null);
+        Path parent = downloadDir == null ? Utils.getCacheDir() : Paths.get(downloadDir);
+        parent = parent.resolve(modelId);
+        Path huggingFaceModelDir = parent.resolve(hash);
+        if (!Files.exists(huggingFaceModelDir)) {
```
frankfliu, line 472: Should we return null if File already exists?
If the model already registered in config.properties, this may lead to conflict

frankfliu (approved):

frankfliu (approved):

frankfliu (approved):

PR #719, opened 2023-05-16
Creates S3 Cache Engine

zachgk:
This creates the S3 Cache engine. It is put into the same cache plugin by expanding the DDB plugin to handle it as well.

Alongside this, there is some work done to synchronize the efforts on cache engines. A new BaseCacheEngine class is created to contain some common logic between the cache engines. The former tests for the DDB cache engine were generalized a bit and turned into a suite that can be and is run for all of the three supported cache engines. This ensures (and fixes) some inconsistencies in behavior among the cache engines.

Also important is that it adds a new test dependency on localstack (https://localstack.cloud/). This runs a local AWS clone inside a docker container and is used to verify the running of the S3 cache.


plugins/cache/build.gradle
```diff
@@ -5,14 +5,17 @@ configurations {
...
     api platform("software.amazon.awssdk:bom:${awssdk_version}")
     api "software.amazon.awssdk:dynamodb"
-    api "ai.djl.aws:aws-ai"
+    api "software.amazon.awssdk:s3"
 
     testImplementation("org.testng:testng:${testng_version}") {
         exclude group: "junit", module: "junit"
     }
     testImplementation "com.amazonaws:DynamoDBLocal:1.21.1"
+    testImplementation "cloud.localstack:localstack-utils:0.2.15"
```
frankfliu, line 18: We should consider add S3 unit-test for djl-serving and djl.
Currently we missed some of test cases related to S3

frankfliu (approved):

plugins/cache/src/main/java/ai/djl/serving/cache/S3CacheEngine.java
```diff
@@ -0,0 +1,297 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+
+/** A {@link CacheEngine} that stores elements in DynamoDB. */
```
siddvenk, line 43: Should this be S3 instead of DynamoDB?

sindhuvahinis (approved):

PR #743, opened 2023-05-22
Iteration-level scheduler with geedy search and contrastive

KexinFeng:
This is outdated. See #803 

---
This PR is an iteration-level scheduler system, which is mainly implemented as a `class GreedySeqBatchScheduler()` and its parent class `SeqBatchScheduler`.  Another two important classes are: `SeqBatcher` and `Batch` (i.e. the `BatchList` in the java implementation)

See https://github.com/deepjavalibrary/djl/pull/2572 for detailed description of them.

To run this feature, see `test_scheduler.py`, which covers basically all the code. (Ignore the beam search and constrastive search part for now. They will be moved to a different PR)

Here are the main features:
1. Iteration-level adding request and request existing, with padding and pad trimming
2. Enable resusable prefix kv_cache input for init_forward. The kv_cache is pre-calculated.
3. Greedy search

Implementation directory
--- 
There are three major classes: `SeqBatchScheduler`, `SeqBatcher` and `Batch` and one major unit test: `test_greedy_scheduler` in this PR. See also https://github.com/deepjavalibrary/djl/pull/2572.

```python
"""
This is a scheduler, serving as an API to the consumer of the systme, allowing for three major actions: initForward, addBatch, fastForward, collectResults. 
"""
class SeqBatchScheduler(ABC):

    def __init__(self, lm_block: LMBlock, config: SearchConfig):
        self.lm_block = lm_block
        self.results = {}
        self.seq_batcher = None
        self.config = config
        self.search_configs = None

    @abstractmethod
    def init_forward(self, input_ids: torch.Tensor, batch_uids: torch.Tensor,
                     kv_cache: Union[Tuple, None]) -> SeqBatcher:

    def increment_forward(self, count: int) -> torch.Tensor:

    @abstractmethod
    def inference_call(self) -> torch.Tensor:
        pass

    def add_request(self, batch_uids: torch.Tensor,
                    input_ids: torch.Tensor,
                    search_configs: Dict[int, SearchConfig] = None,
                    kv_cache: Union[Tuple, None] = None):

    def collect_results(self):

```

```python
"""
This is a data class, which stores the search state (Batch), the control variables (eg seqLength, offSets, etc), and batch operations like merge_batch and trim()
"""
class SeqBatcher(object):

    def __init__(self, batch: Batch, request_uids: torch.Tensor,
                 offsets: torch.Tensor):
        self.batch = batch
        self.request_uids = request_uids
        self.offsets = offsets
        self.exit_index_end_position = {}

[...]

PR #746, opened 2023-05-22
LMBlock

KexinFeng:
The local testing model file can be found in https://github.com/deepjavalibrary/djl/pull/2547. But it is runnable with huggingface model too.

This is a series of PRs. The updates will be done on the latest one: https://github.com/deepjavalibrary/djl-serving/pull/743

### DIFF
--- a/engines/python/setup/djl_python/scheduler/lm_block.py
+++ b/engines/python/setup/djl_python/scheduler/lm_block.py
@@ -1,0 +1,75 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+from abc import ABC, abstractmethod
+from transformers import GPT2LMHeadModel
+from typing import List, Dict
+
+import torch
+
+class GPT_config:
+    def __init__(self):
+        self.numAttentionHeads = 12
+        self.numLayers = 12
+        self.hiddenStateDim = 768
+        self.logitsDim = 50257
+        self.kvDim = 64
+
+
+class LMBlock(ABC):
+    @abstractmethod
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def forward(self, input, past_key_values):
+        pass
+
+
+class PtLMBlock(LMBlock):
+    def __init__(self, model_urls: List[str], gpt_config: GPT_config):
+        super(PtLMBlock, self).__init__()
+        self.blocks = [torch.jit.load(url) for url in model_urls]
+        self.gpt_config = gpt_config
+
+    def forward(self, input: List[torch.tensor], past_key_values):
+        return self.blocks[0](*input) if past_key_values is None else self.blocks[1](*input, past_key_values)
+
+
+class OrtLMBlock(LMBlock):
+    def __init__(self, model_urls: List[str], gpt_config: GPT_config):
+        super(OrtLMBlock, self).__init__()
+        self.blocks = [torch.jit.load(url) for url in model_urls]
+        self.gpt_config = gpt_config
+
+    def forward(self, input: List[torch.tensor], past_key_values):
+        raise ("Not implemented yet")
+
+
+class HuggingfaceGTP2Block(LMBlock):
+    def __init__(self, model_urls: List[str], config: Dict):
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/scheduler/lm_block.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+class HuggingfaceGTP2Block(LMBlock):
+
+    def __init__(self, model_urls: List[str], config: Dict):
+        super(HuggingfaceGTP2Block, self).__init__()
+        self.config = {
+            'use_cache': config.get('use_cache', True),
+            'token_type_ids': config.get('token_type_ids', None),
+            'return_dict': config.get('return_dict', False),
```
sindhuvahinis, line 73: why return_dict is set to False by default? With return_dict we can easily look up logits, past_key_values etc right?

engines/python/setup/djl_python/scheduler/lm_block.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, model_urls: List[str], gpt_config: GPTConfig):
+        super(OrtLMBlock, self).__init__()
+        self.blocks = [torch.jit.load(url) for url in model_urls]
+        self.gpt_config = gpt_config
+
+    def forward(self, input: List[torch.tensor], past_key_values):
+        raise ("Not implemented yet")
+
+
+class HuggingfaceGTP2Block(LMBlock):
```
lanking520, line 66: why we want to rewrite the GPT2 Block?

sindhuvahinis: I think Qing has a point. 
Other backengines like DS, model.forward is called this way
```
model = deepspeed.init_inference(GPT2LMHeadModel.pretrained("gpt2"), config) # returns the DS engine wrapped model
model.forward(input_ids, attention_mask, ...) # this one internally calls the HF model.forward anyway
```
https://github.com/microsoft/DeepSpeed/blob/d1c3c0df5318cbcb2b221d38d4127dd566f4984d/deepspeed/inference/engine.py#L611
Not sure about FT, not sure when will their model.forward will be exposed. cc @rohithkrn  

In these cases, we might not need LMBlock itself. We could get the model and kwargs directly in SeqBatchScheduler constructor and call the model.forward directly for now.  (just thinking aloud)

And also do we need PtBlock and OrtBlock for Python engine?

sindhuvahinis: I had a discussion with @zachgk. One more thing to consider is that, this LMBlock tries to standardize the forward function but it also loads the model. In future if we introduce more models, we could use AutoModel to load, but for Peft models, the model loading by loading the base model first and then their lora updated parameters like below. 
```
peft_model_id = "mrm8488/Alpacoom"
config = PeftConfig.from_pretrained(peft_model_id)
model = AutoModelForCausalLM.from_pretrained(config.base_model_name_or_path, return_dict=True, load_in_8bit=True, device_map={"":0})
model = PeftModel.from_pretrained(model, peft_model_id)
```
Since model loading can be arbitrary as well, I am not sure whether we should have the model loading tied to the LMBlock. I don't have any hard decision comment on this, just laying out the different sides of facts.

engines/python/setup/setup.py
```diff
@@ -44,7 +44,7 @@ def run(self):
 
     requirements = ['psutil', 'packaging', 'wheel']
 
-    test_requirements = ['numpy', 'requests', 'Pillow']
+    test_requirements = ['numpy', 'requests', 'Pillow', 'transformers', 'torch']
```
frankfliu, line 47: I don't think Python engine should depends on torch or transformers. DJL Python engine should be generic.

One option could be create another pip package and install only in DLC.

engines/python/setup/djl_python/scheduler/lm_block.py
```diff
@@ -0,0 +1,86 @@
...
```
frankfliu, line 66: Typo: HuggingfaceGPT2Block

engines/python/setup/djl_python/scheduler/lm_block.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def __init__(self):
+        self.num_attention_heads = 12
+        self.num_layers = 12
+        self.hidden_state_dim = 768
+        self.logits_dim = 50257
+        self.kv_dim = 64
+
+
+class LMBlock(ABC):
```
frankfliu, line 31: Are we going to have different wrapper for each model architecture?

engines/python/setup/djl_python/scheduler/lm_block.py
```diff
@@ -0,0 +1,86 @@
...
+    @abstractmethod
+    def __init__(self):
+        pass
+
+    @abstractmethod
+    def forward(self, input, past_key_values):
+        pass
+
+
+class PtLMBlock(LMBlock):
```
frankfliu, line 42: I assume we only support Huggingface now. Let's remove Pt and Ort.

engines/python/setup/setup.py
sindhuvahinis: But these are for test requirements, for running unit test cases.

PR #751, opened 2023-05-23
fix permissions for sm pysdk install script

siddvenk:
## Description ##

fixes permissions on `install_sagemaker_pysdk.sh` so actions can run the script.


frankfliu (commented):
Why add an empty sh script file?

Should we at least add header in it?
Should we add a comment in the file?

siddvenk:
> Why add an empty sh script file?
> 
> Should we at least add header in it? Should we add a comment in the file?

I think this is just the git diff render issue. This file does have content https://github.com/deepjavalibrary/djl-serving/blob/master/tests/integration/install_sagemaker_pysdk.sh. 

The change i made was ` git update-index --chmod=+x ./tests/integration/install_sagemaker_pysdk.sh`

frankfliu (approved):

PR #763, opened 2023-05-25
Add mme tests to sagemaker tests

siddvenk:
## Description ##

The main change is adding single gpu MME test case with DeepSpeed image.

I also refactored the script slightly to do the following:
- cleanup any dangling resources from the tests at the end of the workflow
- cleanup s3 model artifacts during the tests. Deleting a sagemaker model still leaves the artifacts in S3
- standardize naming of resources so that the cleanup is simple

Testing:
- MME tests are successful in this run, but I want to do one more full run with all the changes to make sure i didn't break anything else. https://github.com/deepjavalibrary/djl-serving/actions/runs/5083982768/jobs/9135696386

Follow Up:
- I'll raise another PR that better supports running the nightly images from alpha ecr

### DIFF
--- a/tests/integration/llm/sagemaker-endpoint-tests.py
+++ b/tests/integration/llm/sagemaker-endpoint-tests.py
@@ -2,6 +2,7 @@ import sagemaker
 import boto3
 from sagemaker.djl_inference import DJLModel, HuggingFaceAccelerateModel, DeepSpeedModel, FasterTransformerModel
 from sagemaker.huggingface import HuggingFaceModel
+from sagemaker.multidatamodel import MultiDataModel
 from sagemaker.utils import unique_name_from_base
 from argparse import ArgumentParser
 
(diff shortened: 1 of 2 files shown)


tests/integration/llm/sagemaker-endpoint-tests.py
```diff
@@ -98,75 +104,166 @@
...
+                             ROLE,
+                             image_uri=config.get("image_uri"),
+                             name=get_name_for_resource(
+                                 model_config.get("model_id") + '-mme'),
+                             sagemaker_session=session,
+                             **model_config.get("model_kwargs"))
+            model.create()
+            created_models.append(model)
+
+        mme = MultiDataModel(get_name_for_resource(name),
```
sindhuvahinis, line 170: How does this map to the created models?

siddvenk: MME works by having models registered under a specific s3 bucket and prefix. All tar.gz files under there are models that can be loaded by the endpoint. 

Above I create a few models in the same s3 path that the MME is created. Then invoking the mme you pass in the artifact name relative to the s3 bucket/prefix. 

It's a bit strange, but for now this is how it works with the PySDK.

lanking520 (approved):

tests/integration/llm/sagemaker-endpoint-tests.py
sindhuvahinis: Oh interesting. Good to know

sindhuvahinis (approved):

PR #764, opened 2023-05-26
Upload SM Benchmark metrics to cloud watch

sindhuvahinis:
## Description ##

### DIFF
--- a/tests/integration/llm/sagemaker-endpoint-tests.py
+++ b/tests/integration/llm/sagemaker-endpoint-tests.py
@@ -129,6 +129,10 @@ MME_CONFIGS = {
     }
 }
 
+ENGINE_TO_METRIC_CONFIG_ENGINE = {
+    "Python" : "Accelerate"
+}
+
 
 def get_sagemaker_session(default_bucket=DEFAULT_BUCKET,
                           default_bucket_prefix=None):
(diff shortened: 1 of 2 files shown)


lanking520 (approved):

tests/integration/llm/sagemaker-endpoint-tests.py
```diff
@@ -26,13 +29,19 @@
         "payload": {
             "prompt": "A picture of a cowboy in space"
         },
-        "deserializer": sagemaker.deserializers.BytesDeserializer()
+        "deserializer": sagemaker.deserializers.BytesDeserializer(),
+        "metrics_config": {
```
siddvenk, line 33: can we determine this from the class? after we create a model we can get the engine name like

`model.engine[0]`

then we can use a map between the engine name and the desired metric config engine (seems like we only really need  a mapping from python to accelerate)

tests/integration/llm/sagemaker-endpoint-tests.py
```diff
@@ -103,6 +130,67 @@
...
+    for _ in range(iterations):
+        start = time.time()
+        predictor.predict(config.get("payload", DEFAULT_PAYLOAD))
+        latencies.append((time.time() - start) * 1000)
+
+    elapsed = (time.time() - begin) * 1000
+
+    benchmark_data = {}
+    benchmark_data['metric_name'] = _get_metric_name(model_name, config)
+    benchmark_data['throughput'] = (iterations - 10) / elapsed * 1000
```
siddvenk, line 185: elapsed here is the total time for 100 requests, but we're reducing iterations by 10 which seems weird.

Should we have a warmup loop of 10 requests where we don't capture any data, and then an actual benchmark loop of 100 iterations?

siddvenk: Also, in the future it would be great if we can parallelize this a bit to exercise the multi worker setup. Doesn't need to be this PR.

PR #772, opened 2023-05-29
[python] Support non-gpu models for huggingface

frankfliu:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -11,9 +11,7 @@
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
 
-import json
 import logging
-import os
 
 import torch
 from transformers import pipeline, Conversation, AutoModelForCausalLM, AutoModelForSeq2SeqLM, AutoTokenizer, AutoConfig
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -83,7 +82,7 @@ def initialize(self, properties: dict):
         if "device_map" in properties:
             kwargs["device_map"] = properties.get("device_map")
             logging.info(f"Using device map {kwargs['device_map']}")
-        elif tp_degree > 0:
+        elif tp_degree > 1:
```
lanking520, line 85: tp==1 sometimes also need auto to utilize the HF Accelerate module

engines/python/setup/djl_python/huggingface.py
```diff
@@ -229,15 +228,14 @@ def wrapped_pipeline(inputs, *args, **kwargs):
...
 
         def wrapped_pipeline(inputs, *args, **kwargs):
             model = hf_pipeline.model
             tokenizer = hf_pipeline.tokenizer
-            input_tokens = tokenizer(inputs, padding=True,
-                                     return_tensors="pt").to(
-                                         torch.cuda.current_device())
+            input_tokens = tokenizer(inputs, padding=True, return_tensors="pt")
+            if self.device_id >= 0:
+                input_tokens.to(torch.cuda.current_device())
```
lanking520, line 238: you can check if cuda current device existense for using GPU or not. So you don't need to rely on device_id in here

Issue #780, opened 2023-05-31
MosaicML/MPT7b model not working on DJLServing handler


lanking520:
## Description

```
import transformers
model = transformers.AutoModelForCausalLM.from_pretrained(
  'mosaicml/mpt-7b',
  trust_remote_code=True
)
```

The model is adding a new option called

```
trust_remote_code=True
```

This needs to enable to run inference

PR #781, opened 2023-05-31
add trust remote code option

lanking520:
## Description ##

This should fixes: #780

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -108,6 +108,7 @@ class DeepSpeedService(object):
         self.model_config = None
         self.low_cpu_mem_usage = False
         self.enable_streaming = False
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -108,6 +108,7 @@ def __init__(self):
         self.model_config = None
         self.low_cpu_mem_usage = False
         self.enable_streaming = False
+        self.trust_remote_code = os.environ.get("TRUST_REMOTE_CODE", "TRUE").lower() == 'true'
```
siddvenk, line 111: I think the default should be false, we don't have a way to validate that the code is safe. Users should have to explicitly opt in to downloading additional code

frankfliu (approved):

siddvenk (approved):

PR #787, opened 2023-05-31
Improvements in AOT UX

sindhuvahinis:
UX improvement 

- [x] Making  serving.properties optional from UX (giving serving.properties should also work), Add —engine (optional), —tensor-parallel-degree, —save-mp-checkpoint-path as command line arguments 
- [x] Add option.parallel_loading=true in generated serving.properties
- [x] Remove option.model_dir  in generated serving.properties in DeepSpeed 
- [ ] Auto detect the engine like LMI Utils
- [ ] Dont replace config.json if it is already present


.github/workflows/llm_integration.yml
```diff
@@ -461,21 +461,21 @@ jobs:
       - name: Download models and dockers
         working-directory: tests/integration
         run: |
-          docker pull deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG
+          docker pull sindhuvahinis/djlserving:ft-test
```
lanking520, line 464: need to change repo?

sindhuvahinis: Added for testing, Removed it now.

frankfliu (approved):

serving/docker/partition/partition.py
```diff
@@ -191,12 +219,31 @@ def run_partition(self):
...
+                        type=str,
+                        required=False,
+                        help='tensor parallel degree')
 
     args = parser.parse_args()
 
     extract_python_jar(PYTHON_CACHE_DIR)
 
-    properties_manager = PropertiesManager(args.model_dir)
+    properties_manager = PropertiesManager(vars(args))
```
frankfliu, line 247: `vars()` put None value in the map

serving/docker/partition/properties_manager.py
```diff
@@ -31,14 +31,13 @@
...
-    def __init__(self, properties_dir):
-        self.properties = {}
-        self.properties_dir = properties_dir
+    def __init__(self, properties):
+        self.properties = properties
+        self.properties_dir = properties.get('properties_dir')
         self.entry_point_url = None
 
-        self.load_properties()
+        load_properties(self.properties_dir, self.properties)
```
frankfliu, line 39: the command line option should override `serving.properties`, this implementation will override command line

PR #798, opened 2023-06-02
[Fix] [CI/CD] Check if input is empty

sindhuvahinis:
## Description ##

Scheduler does not have the options to get inputs. Inputs are only seems to be available in worflow_dispatch. 

So, made the default value of run_test to be empty. Even if input is not even provided, github seems to set the input param name empty by default.  

Testing:
I removed the whole run_test input for yaml and ran the workflow,  github.event.inputs.run_test env variable is validated to be empty. So the tests were not skipped.

https://github.com/deepjavalibrary/djl-serving/actions/runs/5158317906

### DIFF
--- a/.github/workflows/llm_integration.yml
+++ b/.github/workflows/llm_integration.yml
@@ -8,7 +8,7 @@ on:
         required: false
         default: ''
       run_test:
-        description: 'Run only the tests you need [ft, ds, hf, aot, all]'
+        description: 'Run only the tests you need [ft, ds, hf, aot]'
         required: false
         default: ''
   schedule:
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

.github/workflows/llm_integration.yml
```diff
@@ -10,7 +10,7 @@ on:
       run_test:
         description: 'Run only the tests you need [ft, ds, hf, aot, all]'
         required: false
-        default: 'all'
```
frankfliu, line 13: remove all from description

sindhuvahinis: fixed

frankfliu (approved):

PR #801, opened 2023-06-03
add stream generation for huggingface streamer

lanking520:
## Description ##

Use huggingface streamer for inputs

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -62,7 +62,7 @@ class HuggingFaceService(object):
     def __init__(self):
         self.hf_pipeline = None
         self.initialized = False
-        self.enable_streaming = False
+        self.enable_streaming = None
         self.model = None
         self.device_id = -1
         self.tokenizer = None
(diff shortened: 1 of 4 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -78,8 +78,7 @@ def initialize(self, properties: dict):
         self.device_id = int(properties.get("device_id", "-1"))
         task = properties.get("task")
         tp_degree = int(properties.get("tensor_parallel_degree", "-1"))
-        self.enable_streaming = properties.get("enable_streaming",
-                                               "false").lower() == "true"
+        self.enable_streaming = properties.get("enable_streaming", None)
```
frankfliu, line 81:
```suggestion
        self.enable_streaming = properties.get("enable_streaming", None)
        if self.enable_streaming and self.enable_streaming.lower() == "false":
            self.enable_streaming = None
```

engines/python/setup/djl_python/huggingface.py
```diff
@@ -139,12 +138,18 @@ def inference(self, inputs):
             outputs = Output()
 
             if self.enable_streaming:
-                stream_generator, ctype = StreamingUtils.get_stream_generator(
```
frankfliu, line 142: You need update deepspeed and neuronx as well

frankfliu (approved):

PR #802, opened 2023-06-05
[docker] bump versions and new deepspeed wheel

tosterberg:
## Description ##

PR to update the deepspeed wheel once it is published in Prod, as well as update torch and triton for the container.


lanking520 (approved):

lanking520:
close this since https://github.com/deepjavalibrary/djl-serving/pull/804 monitor the process

PR #807, opened 2023-06-06
fix typo in sm workflow inputs

siddvenk:
## Description ##

Typo in default value of mode was causing issues starting the python script. This should fix that, I kicked off a run here with the changes and it seems to be working so far https://github.com/deepjavalibrary/djl-serving/actions/runs/5191566105

### DIFF
--- a/.github/workflows/sagemaker-integration.yml
+++ b/.github/workflows/sagemaker-integration.yml
@@ -6,7 +6,7 @@ on:
       mode:
         description: "release/nightly containers to test. Default is nightly"
         required: false
-        default: nightly
+        default: 'nightly'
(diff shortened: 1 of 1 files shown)


sindhuvahinis (approved):

.github/workflows/sagemaker-integration.yml
```diff
@@ -6,7 +6,7 @@ on:
       mode:
         description: "release/nightly containers to test. Default is nightly"
         required: false
-        default: nightly
+        default: 'nightly'
```
lanking520, line 9: the true default will be empty (if triggered by scheduler) please make sure you have this checked in the script.

PR #819, opened 2023-06-09
[partition] keep 'option' in properties

sindhuvahinis:
## Description ##

In the last PR, 'option' keyword was not removed until the control is passed to model.py. Looking up every keyword with option keyword. 
This should also fix the actions failure. Tested it manually.

### DIFF
--- a/serving/docker/partition/partition.py
+++ b/serving/docker/partition/partition.py
@@ -42,7 +42,7 @@ class PartitionService(object):
         self.download_model_from_s3()
 
     def download_model_from_s3(self):
-        model_id = self.properties.get("model_id")
+        model_id = self.properties.get("option.model_id")
         if not model_id or not model_id.startswith("s3://"):
             return
 
@@ -71,7 +71,7 @@ class PartitionService(object):
         if not glob.glob(os.path.join(download_dir, '*')):
             raise Exception('Model download from s3url failed')
 
-        self.properties['model_id'] = download_dir
+        self.properties['option.model_id'] = download_dir
 
     def install_requirements_file(self):
         req_file_dir = self.properties_manager.properties_dir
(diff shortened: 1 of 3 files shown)


lanking520 (approved):

serving/docker/partition/partition.py
```diff
@@ -99,20 +99,20 @@ def set_environmental_vars(self):
         if os.environ.get("PYTHONPATH"):
             python_path.append(os.environ.get("PYTHONPATH"))
         python_path.append(PYTHON_CACHE_DIR)
-        if 'model_dir' in self.properties:
-            python_path.append(self.properties['model_dir'])
+        if 'option.model_dir' in self.properties:
```
frankfliu, line 102: 1. Not all perperties has `option.` prefix. 
2. We don't think we are using `model_dir` in the `serving.perperties`, model_dir is passed to `model.py` ad current directory

serving/docker/partition/partition.py
```diff
@@ -121,23 +121,25 @@ def download_config_from_hf(self):
...
             config_files += glob.glob(os.path.join(model_dir, pattern))
 
         for file in config_files:
-            shutil.copy(file, dst=self.properties['save_mp_checkpoint_path'])
+            shutil.copy(file,
+                        dst=self.properties['option.save_mp_checkpoint_path'])
 
     def upload_checkpoints_to_s3(self):
-        if 'upload_checkpoints_s3url' not in self.properties:
+        if 'option.upload_checkpoints_s3url' not in self.properties:
```
frankfliu, line 137: No need to add 'option.` to intermediate key using only for partition

sindhuvahinis: Okay, fixed

serving/docker/partition/partition.py
sindhuvahinis: Alright, removed option for model_dir

frankfliu (approved):

PR #828, opened 2023-06-12
[python] Adds rolling batch support

frankfliu:
## Description ##

### DIFF
--- a/engines/python/src/test/resources/rolling_batch/model.py
+++ b/engines/python/src/test/resources/rolling_batch/model.py
@@ -1,0 +1,127 @@
+#!/usr/bin/env python
+#
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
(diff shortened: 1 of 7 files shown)


lanking520:
Rolling batch supporting

lanking520 (approved):

engines/python/src/test/resources/rolling_batch/model.py
```diff
@@ -0,0 +1,127 @@
...
+            self._merge_request(batch)
+            time.sleep(0.1)
+
+            for i in range(batch_size):
+                req = self.pending_requests[i]
+                res = {
+                    "data": req.get_next_token(),
+                    "last": req.is_last_token(),
+                }
+                outputs.add_as_json(res, batch_index=i)
```
lanking520, line 86: batch index is not safe here. Probably need some id to ensure the order.

frankfliu: Use request_id won't provide extra safety.
But I agree adding request_id can make debug/troubleshooting much easier

PR #843, opened 2023-06-17
Max_seqbatcher_number_threshold_api

KexinFeng:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/scheduler/seq_batch_scheduler.py
+++ b/engines/python/setup/djl_python/scheduler/seq_batch_scheduler.py
@@ -103,7 +103,6 @@ class SeqBatchScheduler:
                                   for seq_batcher in seq_batcher_list)
         return batch_size
 
-    def inference_call(self) -> Tuple[List[List[int]], List[int], List[int]]:
     def optimal_action(self,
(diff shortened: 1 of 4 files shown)


engines/python/setup/djl_python/scheduler/seq_batch_scheduler.py
```diff
@@ -87,8 +89,36 @@ def is_empty(self):
                    for seq_batcher_list in self.seq_batchers.values()
                    for seq_batcher in seq_batcher_list)
 
-    def inference_call(
-            self) -> Tuple[List[List[int]], List[int], List[List[int]]]:
+    def optimal_action(self,
+                       input_ids: torch.Tensor,
+                       request_uids: torch.Tensor,
```
lanking520, line 94: why request id in tensor format?

KexinFeng: The tensor format of request_id is used inside the scheduler, mainly to use the array indexing feature of the tensor. Here the request_id can be converted from List[int] to tensor outside of the add_request or optimal_action functions. But this conversion can be done inside these two functions as well if needed.

lanking520 (approved):

PR #844, opened 2023-06-19
[docker] add llama support and integration tests

tosterberg:
Update to deepspeed container, adds and pins protobuf 3.20.3 as a dependency for conversion of the fast llama tokenizer. Adds integration tests for llama using HF accelerate and DS.

--Edit: Now uses fused kernels


lanking520 (changes requested):
why not fused kernel for DeepSpeed?

PR #856, opened 2023-06-23
fix T5 model not support INT8 issue on handler

### LINKED ISSUE #855: Flan-T5-xxl not supported directly by handler
Failing setup:

```
engine=Python
option.entryPoint=djl_python.huggingface
option.task=text2text-generation
option.tensor_parallel_degree=1
option.model_id=google/flan-t5-xxl
option.dtype=fp16
option.device_map=auto
option.load_in_8bit=true
```

lanking520:
## Description ##

Fixes #855

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -214,6 +214,7 @@ class HuggingFaceService(object):
                                    device=self.device,
                                    **kwargs)
         else:
+            kwargs.pop("tokenizer", None)
             self._init_model_and_tokenizer(model_id_or_path, **kwargs)
             hf_pipeline = pipeline(task=task,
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -214,13 +214,10 @@ def get_pipeline(self, task: str, model_id_or_path: str, kwargs):
                                    device=self.device,
                                    **kwargs)
         else:
-            tokenizer = AutoTokenizer.from_pretrained(model_id_or_path)
-            kwargs.pop("tokenizer", None)
```
frankfliu, line 218: do we need pop tokenizer in _init_model_and_tokenizer()?

frankfliu (approved):

PR #858, opened 2023-06-23
Update netty for cve

siddvenk:
## Description ##

### DIFF
--- a/serving/docker/scripts/install_djl_serving.sh
+++ b/serving/docker/scripts/install_djl_serving.sh
@@ -27,3 +27,18 @@ else
     rm -rf /opt/djl/logs
   fi
 fi
+
+cd /usr/local/djl-serving-*/lib
+rm netty*.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport-classes-epoll/4.1.94.Final/netty-transport-classes-epoll-4.1.94.Final.jar -f -o netty-transport-classes-epoll-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport-classes-kqueue/4.1.94.Final/netty-transport-classes-kqueue-4.1.94.Final.jar -f -o netty-transport-classes-kqueue-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport-native-epoll/4.1.94.Final/netty-transport-native-epoll-4.1.94.Final-linux-aarch_64.jar -f -o netty-transport-native-epoll-4.1.94.Final-linux-aarch_64.jar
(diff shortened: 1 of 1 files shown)


serving/docker/scripts/install_djl_serving.sh
```diff
@@ -27,3 +27,20 @@ else
...
+curl https://repo1.maven.org/maven2/io/netty/netty-codec-http/4.1.94.Final/netty-codec-http-4.1.94.Final.jar -f -o netty-codec-http-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-common/4.1.94.Final/netty-common-4.1.94.Final.jar -f -o netty-common-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-handler/4.1.94.Final/netty-handler-4.1.94.Final.jar -f -o netty-handler-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-resolver/4.1.94.Final/netty-resolver-4.1.94.Final.jar -f -o netty-resolver-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport/4.1.94.Final/netty-transport-4.1.94.Final.jar  -f -o netty-transport-4.1.94.Final.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport-native-epoll/4.1.94.Final/netty-transport-native-epoll-4.1.94.Final-linux-x86_64.jar -f -o netty-transport-native-epoll-4.1.94.Final-linux-x86_64.jar
+curl https://repo1.maven.org/maven2/io/netty/netty-transport-native-kqueue/4.1.94.Final/netty-transport-native-kqueue-4.1.94.Final-osx-aarch_64.jar -f -o netty-transport-native-kqueue-4.1.94.Final-osx-aarch_64.jar
```
lanking520, line 44: do we need to ship OSX jars?

siddvenk: we don't need to, I can remove these. I wrote a script to translate the existing netty jars into the updated ones, so they're all here. 

Though, if anyone is still pulling these from dockerhub and trying to use python engine on mac, i guess that would no longer work.

frankfliu: We don't need them, but the package is small

lanking520: they will run in a linux container, OSX jar will not work

siddvenk: I removed the osx jars

lanking520 (approved):

PR #866, opened 2023-06-25
Rolling batch integration tests

sindhuvahinis:
## Description ##

### DIFF
--- a/.github/workflows/rolling_batch_integration.yml
+++ b/.github/workflows/rolling_batch_integration.yml
@@ -8,7 +8,7 @@ on:
         required: false
         default: ''
   schedule:
-    - cron: '0 15 * * *'
+    - cron: '0 6 * * *'
 
 
 jobs:
--- a/.github/workflows/rolling_batch_integration.yml
+++ b/.github/workflows/rolling_batch_integration.yml
@@ -1,0 +1,172 @@
+name: Rolling Batch Integration tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      djl-version:
+        description: 'The released version of DJL'
+        required: false
+        default: ''
+  schedule:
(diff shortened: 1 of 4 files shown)


tests/integration/rb_client.py
```diff
@@ -0,0 +1,151 @@
...
+# sends multiple requests with different data json concurrently
+def test_concurrent_with_mul_reqs(model, test_spec, spec_name):
+    if model not in test_spec:
+        raise ValueError(
+            f"{args.model} is not one of the supporting models {list(test_spec.keys())}"
+        )
+
+    processes = []
+    spec = test_spec[model]
+    for index in range(len(spec.get("input_texts"))):
```
lanking520, line 116: why multiple processes and concurrent clients used at the same time?

sindhuvahinis: Multiple process with one concurrent client. Need to send different requests with different decoding strategy requests. For example, greedy and contrastive can be sent to the SeqBatchScheduler at the same time. But SeqBatchScheduler will call the forward separately for greedy and contrastive.  This is a scheduler only use case.

.github/workflows/rolling_batch_integration.yml
```diff
@@ -0,0 +1,172 @@
...
+
+on:
+  workflow_dispatch:
+    inputs:
+      djl-version:
+        description: 'The released version of DJL'
+        required: false
+        default: ''
+  schedule:
+    - cron: '0 15 * * *'
```
lanking520, line 11: let's schedule a different time, 3PM GMT is too busy.

sindhuvahinis: Done

lanking520 (approved):

PR #877, opened 2023-06-28
[VLLM] add vllm rolling batch and add hazard handling

lanking520:
## Description ##

Add vLLM rolling batch implementation


engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -64,6 +64,19 @@ def is_last_token(self) -> bool:
         return self.last_token
 
 
+def stop_on_any_exception(func):
+
+    def try_catch_handling(self, input_data, parameters):
+        try:
+            return func(self, input_data, parameters)
+        except Exception as e:
```
frankfliu, line 72: logging the exception here

frankfliu (approved):

sindhuvahinis (approved):

PR #878, opened 2023-06-29
Sampling decoding implementation

KexinFeng:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/tests/test_scheduler.py
+++ b/engines/python/setup/djl_python/tests/test_scheduler.py
@@ -117,6 +117,55 @@ class TestScheduler(unittest.TestCase):
             results[4][:30]
         ) == '!!!!!!!!!!DeepMind Company is a company that is dedicated to the advancement of artificial intelligence. We are a company'
 
+    def test_sampling_scheduler(self):
+        model_id = "gpt2"
+        model = GPT2LMHeadModel.from_pretrained(model_id)
+        tokenizer = GPT2Tokenizer.from_pretrained(model_id)
+        lm_block = HuggingfaceBlock(model)
+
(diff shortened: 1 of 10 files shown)


engines/python/setup/djl_python/tests/test_scheduler.py
```diff
@@ -117,6 +119,55 @@ def test_greedy_scheduler(self):
...
 
+    def test_sampling_scheduler(self):
+        numpy.random.seed(20220611)
+
+        model_id = "gpt2"
+        model = GPT2LMHeadModel.from_pretrained(model_id)
+        tokenizer = GPT2Tokenizer.from_pretrained(model_id)
+        lm_block = HuggingfaceBlock(model)
+
+        scheduler = SeqBatchScheduler(lm_block, "greedy", SearchConfig())
```
lanking520, line 130: why here is greedy?

KexinFeng: Now the GreedySeqBatcher contains both greedy and sampling. The name hasn't changed yet since the front end will need to change accordingly. I'm thinking of changing the name GreedySeqBatcher -> SimpleSeqBatcher.

KexinFeng: I don't know how much change it introduces on the front end. So I kept the old name, so the sake of not delaying the delivery. Maybe we can change the name in a different PR, which contains both this change and front end change? The user still can now use the sampling by setting use_sampling=True in the search_config, though.

KexinFeng: It is now temporarily solved: from the front end this naming of greedySeqBatcher won't cause confusion now.

lanking520 (approved):

PR #879, opened 2023-06-29
Runtime kv cache

KexinFeng:
## Description ##


engines/python/setup/djl_python/scheduler/seq_batch_scheduler.py
```diff
@@ -41,15 +42,82 @@ def __init__(self, lm_block: LMBlock, default_search_algorithm: str,
...
+        self.lru_max_size = 10
 
     def add_request(self,
                     input_ids: torch.Tensor,
                     request_uids: torch.Tensor,
                     search_algorithm: str = None,
                     search_configs: List[SearchConfig] = None,
-                    kv_cache: Union[Tuple, None] = None):
+                    kv_cache: Union[Tuple, None] = None,
+                    kv_cache_prompt_ids: Union[Dict[int, torch.tensor], None] = None):
```
lanking520, line 57: why do you need index for the prompt id? Python support Dict with no value/no key

KexinFeng: The index is needed here since some users may send prompt but some users may not send prompt. So the storage here needs to be tagged by request id

lanking520: Can we use prompt ids as the key? Or Just the tensor of the input prompts. Look up should be on the input prompt other than the request id

lanking520: But your key is a single digit integer

```
Dict[int, torch.tensor]
```

Should this be List[Int] ? Or Numpy array

lanking520 (approved):

PR #881, opened 2023-06-29
Non streaming for rolling batch

sindhuvahinis:
## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
@@ -124,7 +124,7 @@ class RollingBatch implements Runnable {
                     throw new TranslateException("Time out in: " + timeout);
                 }
             }
-            Request req = new Request(input);
+            Request req = new Request(input, enableStreaming);
             list.add(req);
             canRead.signal();
             return req.output;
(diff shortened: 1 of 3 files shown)


lanking520 (commented):
please also change the python handler code. Enabling streaming will make the handler fall streaming path by default

lanking520 (approved):

lanking520 (changes requested):
please fix the PMD issue

engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
```diff
@@ -48,11 +48,13 @@ class RollingBatch implements Runnable {
     private ReentrantLock lock;
     private Condition canAdd;
     private Condition canRead;
+    private boolean enableStreaming;
 
-    RollingBatch(PyProcess process, int maxRollingBatchSize, int timeout) {
+    RollingBatch(PyProcess process, int maxRollingBatchSize, int timeout, boolean enableStreaming) {
```
frankfliu, line 53: In which case we want to disable streaming if rolling batch is enabled?

lanking520: Some cases customer just want to get output in 1 go. Also for short sequence generation, not necessary for rolling batch

lanking520 (approved):

PR #883, opened 2023-06-29
Raise inference failure exceptions in default handlers

rohithkrn:
## Description ##

Currently inference failure exceptions are suppressed in default handlers preventing OOM exceptions to be handled correctly. `Output` is constructed [here](https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python_engine.py#L124-L128) anyway.

This is needed for MME on SM


siddvenk (approved):

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -368,7 +368,7 @@ def inference(self, inputs: Input):
             outputs.add_property("content-type", "application/json")
         except Exception as e:
             logging.exception("DeepSpeed inference failed")
-            outputs = Output().error((str(e)))
+            raise e
```
frankfliu, line 371: This will kill the python process and cause SM restart the host, we don't really know if this error is recoverable or not

frankfliu, line 371: If we rethrow the exception, we don't need to catch it at the first place

rohithkrn: yeah, I thought about it too. I did not remove it as I was not sure if the log statement right above provided any value.

frankfliu: the exception will be caught at higher level and will be logged there

rohithkrn: done, updated

frankfliu (approved):

PR #898, opened 2023-07-03
Add oom unit tests for load and invoke

rohithkrn:
## Description ##

Add oom unit tests for load and invoke

### DIFF
--- a/engines/python/src/test/resources/echo/model.py
+++ b/engines/python/src/test/resources/echo/model.py
@@ -21,6 +21,7 @@ import time
 from djl_python import Input
 from djl_python import Output
 
+
 def stream_token():
     for i in range(5):
         time.sleep(1)
--- a/engines/python/src/test/resources/echo/model.py
+++ b/engines/python/src/test/resources/echo/model.py
@@ -21,10 +21,6 @@ import time
 from djl_python import Input
 from djl_python import Output
 
-
-class OutOfMemoryError(Exception):
-    pass
-
 def stream_token():
     for i in range(5):
         time.sleep(1)
@@ -43,7 +39,7 @@ def handle(inputs: Input):
     elif inputs.contains_key("exit"):
         sys.exit()
     elif inputs.contains_key("OOM"):
-        raise OutOfMemoryError
+        raise MemoryError
 
     data = inputs.get_as_bytes()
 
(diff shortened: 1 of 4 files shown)


engines/python/src/test/resources/echo/model.py
```diff
@@ -39,6 +42,8 @@ def handle(inputs: Input):
         return "invalid_type"
     elif inputs.contains_key("exit"):
         sys.exit()
+    elif inputs.contains_key("OOM"):
+        raise OutOfMemoryError
```
frankfliu, line 46: You can just use built-in exception MemoryError

engines/python/src/test/resources/oom/serving.properties
```diff
@@ -0,0 +1,4 @@
+option.pythonExecutable=python3
```
frankfliu, line 1: this file is unnecessary, we can remove it

engines/python/src/test/resources/oom/model.py
```diff
@@ -0,0 +1,38 @@
...
+    pass
+
+
+def handle(inputs: Input):
+    """
+    Default handler function
+    """
+
+    if inputs.is_empty():
+        raise OutOfMemoryError()
```
frankfliu, line 35:
```suggestion
        raise MemoryError()
```

engines/python/src/test/resources/echo/model.py
rohithkrn: pytorch throws `OutOfMemoryError`, so thought it makes more sense to test that particular case.

engines/python/src/test/resources/oom/serving.properties
rohithkrn: updated.

engines/python/src/test/resources/echo/model.py
frankfliu: This is mock test anyway. No difference between built-in exception vs customized class

rohithkrn: Updated to keep it clean. Although there's a minor difference, when code change happens for some reason where we check for `OutOfMemoryError`, this test catches incorrect changes.

frankfliu (approved):

PR #906, opened 2023-07-05
Fix Jacoco report on some python code and artifacts

lanking520:
## Description ##

Skip python for invalid content validation


jacoco/build.gradle
```diff
@@ -30,3 +30,11 @@ reporting {
 tasks.named("check") {
     dependsOn tasks.named("testCodeCoverageReport", JacocoReport)
 }
+
+testCodeCoverageReport {
```
frankfliu, line 34: I have fixed this issue. We should not checkin .pt file in repo

PR #927, opened 2023-07-10
[fix] add current device for tp > 1 scenario on huggingface handler

tosterberg:
Current HF handler sets self.device to None when running on multiple GPU's, but our wrapped pipeline uses self.device for moving the input tokens to the device. Since this is None we needed another control path for getting the input tokens onto the GPU. The current implementation leaves the input tokens on CPU, degrading the performance.

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -384,6 +384,8 @@ class HuggingFaceService(object):
             input_tokens = tokenizer(inputs, padding=True, return_tensors="pt")
             if self.device:
                 input_tokens = input_tokens.to(self.device)
+            else:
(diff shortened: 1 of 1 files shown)


siddvenk (approved):

engines/python/setup/djl_python/huggingface.py
```diff
@@ -384,6 +384,8 @@ def wrapped_pipeline(inputs, *args, **kwargs):
             input_tokens = tokenizer(inputs, padding=True, return_tensors="pt")
             if self.device:
                 input_tokens = input_tokens.to(self.device)
+            else:
+                input_tokens = input_tokens.to(torch.cuda.current_device())
```
frankfliu, line 388: Do we really need this, I thought pipeline will convert the device.

PR #934, opened 2023-07-11
[python] set default maxWorkers to 1 if not configured for TP

frankfliu:
## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -230,7 +230,7 @@ public class PyModel extends BaseModel {
         } else {
             int tensorParallelDegree = pyEnv.getTensorParallelDegree();
             if (tensorParallelDegree > 0) {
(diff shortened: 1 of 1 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -230,7 +230,7 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
         } else {
             int tensorParallelDegree = pyEnv.getTensorParallelDegree();
             if (tensorParallelDegree > 0) {
-                if (tensorParallelDegree > 1) {
+                if (getProperty("maxWorkers") == null && getProperty("gpu.maxWorkers") == null) {
```
lanking520, line 233: In MPI mode, I set maxWorker to 16

frankfliu: in MPI mode, you can only restrict maxWorkers, you cannot set to 16

lanking520 (approved):

PR #952, opened 2023-07-13
Fix some issues with remote code for lora

siddvenk:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -147,14 +147,17 @@ class HuggingFaceService(object):
             if "device_map" not in kwargs:
                 raise ValueError(
                     "device_map should set when load_in_8bit is set")
-            kwargs["load_in_8bit"] = properties.get("load_in_8bit").lower() == 'true'
+            kwargs["load_in_8bit"] = properties.get(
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -300,9 +300,11 @@ def get_pipeline(self, task: str, model_id_or_path: str, kwargs):
                 base_model = PEFT_MODEL_TASK_TO_CLS[
                     self.peft_config.task_type].from_pretrained(
                         self.peft_config.base_model_name_or_path, **kwargs)
-                lora_model = PeftModel.from_pretrained(base_model,
+                self.model = PeftModel.from_pretrained(base_model,
                                                        model_id_or_path)
-                self.model = lora_model.merge_and_unload()
+                # Peft currently does not support merging when target dtype is quantized
+                if "load_in_8bit" not in kwargs and "load_in_4bit" not in kwargs:
```
frankfliu, line 306: Should we print a warning if these flag is set?

siddvenk: yes I think that's a good think to log.

frankfliu (approved):

PR #953, opened 2023-07-14
Allow overriding truncate parameter in request

maaquib:
## Description ##

Allows overriding the default 1000 value for truncation in the request parameters


engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -151,7 +151,8 @@ def preprocess_requests(self, requests, **kwargs):
                     id=r.id,
                     inputs=r.input_text,
                     parameters=parameters,
-                    stopping_parameters=stop_parameters))
+                    stopping_parameters=stop_parameters,
+                    truncate=param.get("truncate", 1000)))
```
lanking520, line 155: Where do we pass this param?

lanking520 (approved):

PR #960, opened 2023-07-19
Enable multi-gpu inference (device_map='auto') on seq_batch_scheduler

KexinFeng:
kv cache on cuda may have different indices, which may cause device mismatch error: `RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cuda:1)`

This PR makes sure that when `device_map="auto"` is called, i.e. multi-gpu is allocated, everything still works. 

Before this, `model.to('cuda')` was called, which lockes the model in a single gpu.

### DIFF
--- a/engines/python/setup/djl_python/tests/test_scheduler.py
+++ b/engines/python/setup/djl_python/tests/test_scheduler.py
@@ -1,8 +1,6 @@
 import unittest
 from collections import defaultdict
 
-import numpy
-
 from djl_python.scheduler import HuggingfaceBlock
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/tests/test_scheduler.py
```diff
@@ -20,7 +20,7 @@ class TestScheduler(unittest.TestCase):
 
     def test_lm_block(self):
         model_id = "gpt2"
-        model = GPT2LMHeadModel.from_pretrained(model_id, device_map="auto")
+        model = GPT2LMHeadModel.from_pretrained(model_id).to(device)
```
frankfliu, line 23: How can we load the large model (requires multi-gpu) without device_map?

PR #969, opened 2023-07-24
[fix] Fix llama model support by specifying pad token id = 0(on gpu)

KexinFeng:
## Description ##
Testing model: openlm-research/open_llama_3b

* Specify the padding token id to be 0
Padding value 50256 will trigger error:
```
        RuntimeError: CUDA error: device-side assert triggered
        Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```
* Slightly refactor the unit test

### DIFF
--- a/engines/python/setup/setup.py
+++ b/engines/python/setup/setup.py
@@ -46,7 +46,7 @@ if __name__ == '__main__':
 
     test_requirements = [
         'numpy', 'requests', 'Pillow', 'transformers', 'torch', 'einops',
-        'accelerate', 'sentencepiece'
+        'accelerate', 'sentencepiece', 'protobuf'
     ]
 
     setup(name='djl_python',


engines/python/setup/djl_python/scheduler/search_config.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, **kwargs):
         self.max_new_seqlen = kwargs.get('max_new_tokens', 30)
         self._max_seqlen = 0
         self.eos_token_id = kwargs.get('eos_token_id', 50256)
-        self.pad_token_id = kwargs.get('pad_token_id', 50256)
+        self.pad_token_id = kwargs.get('pad_token_id', 0)
```
frankfliu, line 22: why we hard code pad_token_id?

frankfliu (approved):

PR #975, opened 2023-07-26
[python] Update lmi-dist

xyang16:
## Description ##

### DIFF
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -16,6 +16,7 @@ ARG python_version=3.9
 ARG torch_version=2.0.1
 ARG torch_vision_version=0.15.2
 ARG deepspeed_wheel="https://publish.djl.ai/deepspeed/deepspeed-nightly-py2.py3-none-any.whl"
+ARG vllm_wheel="https://publish.djl.ai/vllm/vllm-0.0.0-cp39-cp39-linux_x86_64.whl"
 ARG lmi_dist_wheel="https://publish.djl.ai/lmi_dist/lmi_dist-nightly-py3-none-any.whl"
 ARG protobuf_version=3.20.3
 ARG transformers_version=4.30.2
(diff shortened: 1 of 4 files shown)


serving/docker/deepspeed.Dockerfile
```diff
@@ -60,10 +61,11 @@ RUN apt-get update && \
     scripts/install_s5cmd.sh x64 && \
     DEBIAN_FRONTEND=noninteractive apt-get install -yq libaio-dev libopenmpi-dev && \
     pip3 install torch==${torch_version} torchvision==${torch_vision_version} --extra-index-url https://download.pytorch.org/whl/cu118 \
-    ${deepspeed_wheel} ${lmi_dist_wheel} protobuf==${protobuf_version} transformers==${transformers_version} \
+    ${deepspeed_wheel} ${vllm_wheel} ${lmi_dist_wheel} protobuf==${protobuf_version} transformers==${transformers_version} \
     mpi4py sentencepiece einops accelerate==${accelerate_version} bitsandbytes==${bitsandbytes_version}\
     diffusers[torch]==${diffusers_version} peft==${peft_version} opencv-contrib-python-headless safetensors scipy && \
     scripts/install_flash_attn.sh && \
+    scripts/install_flash_attn_v2.sh && \
```
frankfliu, line 68: Why we need both?

serving/docker/deepspeed.Dockerfile
```diff
@@ -16,6 +16,7 @@ ARG python_version=3.9
 ARG torch_version=2.0.1
 ARG torch_vision_version=0.15.2
 ARG deepspeed_wheel="https://publish.djl.ai/deepspeed/deepspeed-nightly-py2.py3-none-any.whl"
+ARG vllm_wheel="https://publish.djl.ai/vllm/vllm-0.0.0-cp39-cp39-linux_x86_64.whl"
```
lanking520, line 19: I would check with @frankfliu , vllm introduced a lot of dependencies you don't want, that's why we didn't have it for V5. If you already understand this, please feel free to ignore

xyang16: Yes, I discussed with Frank about this yesterday. So this solution to install wheel will not install the dependencies.

xyang16: Total size increase is 0.5GB.

frankfliu (approved):

PR #983, opened 2023-07-28
[python] Clean up dangling process in java

frankfliu:
This can prevent mpi sub process kill each other

## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/Connection.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/Connection.java
@@ -51,11 +51,15 @@ import java.io.IOException;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

engines/python/src/main/java/ai/djl/python/engine/Connection.java
```diff
@@ -75,6 +79,26 @@ class Connection {
...
             throws IOException {
+        Path tmp = Paths.get(System.getProperty("java.io.tmpdir"));
+        try (Stream<Path> stream = Files.list(tmp)) {
+            stream.forEach(
+                    p -> {
+                        try {
+                            String name = p.toFile().getName();
+                            if (name.startsWith("djl_sock." + port) && name.endsWith(".pid")) {
+                                long pid = Long.parseLong(Files.readString(p));
+                                Optional<ProcessHandle> handle = ProcessHandle.of(pid);
```
lanking520, line 90: This will prevent 2 Model server starting from the same environment right?

PR #999, opened 2023-08-08
Fix batch offset computation in FT handler

rohithkrn:
## Description ##

Was looking at the code and this seemed like a bug as were are not updating offset for each item in the batch.


rohithkrn:
@xyang16 I see you originally authored the batching code. Could you please take a look?

engines/python/setup/djl_python/fastertransformer.py
```diff
@@ -189,6 +189,7 @@ def inference(self, inputs: Input):
                 "generated_text": s
             } for s in result[offset:offset + input_size[i]]]
             outputs.add(generated_text, key=inputs.get_content().key_at(i))
+            offset = offset + input_size[i]
```
xyang16, line 192:
```suggestion
            offset += input_size[i]
```

xyang16 (approved):

PR #1008, opened 2023-08-09
[python] validate each request in the batch

frankfliu:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -10,7 +10,7 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-
+import json
 import logging
 import os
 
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -199,54 +199,73 @@ def initialize(self, properties: dict):
...
         outputs = Output()
 
+        input_data, input_size, parameters, errors, batch = self.parse_inputs(inputs)
         if self.rolling_batch_type:
             if inputs.get_property("reset_rollingbatch"):
                 self.rolling_batch.reset()
             result = self.rolling_batch.inference(input_data, parameters)
-            for i in range(inputs.get_batch_size()):
-                outputs.add(result[i], key="data", batch_index=i)
+            for i in range(len(input_data)):
```
lanking520, line 256: use input size

engines/python/setup/djl_python/huggingface.py
```diff
@@ -199,54 +199,73 @@ def initialize(self, properties: dict):
...
+                err = errors.get(i)
+                if err:
+                    outputs.add(err, key="data", batch_index=i)
+                else:
+                    outputs.add(result[i], key="data", batch_index=i)
```
lanking520, line 261: i is not the right index, probably we need to +1 everytime we insert a new value. Or use iterator to iter a value.

lanking520 (approved):

engines/python/setup/djl_python/huggingface.py
```diff
@@ -199,84 +199,117 @@ def initialize(self, properties: dict):
...
+                        "Accelerate")
+                    outputs.add_stream_content(
+                        stream_generator(self.model, self.tokenizer,
+                                         input_data, self.device,
+                                         **parameters[0]))
+                return outputs
+
+            prediction = self.hf_pipeline(input_data, **parameters[0])
+        else:
+            prediction = []
```
lanking520, line 288: need to check rolling batch here. If rolling batch, need to return all errors

engines/python/setup/djl_python/huggingface.py
```diff
@@ -199,54 +199,79 @@ def initialize(self, properties: dict):
...
+        return input_data, input_size, parameters, errors, batch
 
+    def inference(self, inputs):
         outputs = Output()
 
+        input_data, input_size, parameters, errors, batch = self.parse_inputs(
+            inputs)
+        if len(input_data) == 0:
+            for i in range(len(batch)):
+                outputs.add(errors.get(i), key="data", batch_index=i)
```
lanking520, line 253: Will this work for dynamic batch?

lanking520: this place may also need the json double dumps

lanking520 (approved):

lanking520 (approved):

PR #1009, opened 2023-08-09
[serving] Allows print access log to console

frankfliu:
## Description ##


zachgk (approved):
Also, we should probably add a doc somewhere explaining the log4j configurations we have and how to use them

lanking520 (approved):

PR #1021, opened 2023-08-16
Simplify handling of min/max workers

zachgk:
This unifies the handling of min/max workers into the two pieces of WorkerGroup and ModelInfo. The source becomes clearly sourced from modelInfo unless a default is passed during a WorkerGroup creation through a scaling call. Before, there were more layers of passed in data and defaults in ways that were difficult to track.

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -20,6 +20,7 @@ import ai.djl.engine.Engine;
 import ai.djl.engine.EngineException;
 import ai.djl.modality.Input;
 import ai.djl.modality.Output;
+import ai.djl.ndarray.NDManager;
 import ai.djl.repository.Artifact;
 import ai.djl.repository.FilenameUtils;
 import ai.djl.repository.MRL;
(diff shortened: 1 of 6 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -457,22 +458,124 @@ public int getQueueSize() {
...
     /**
      * Returns the minimum number of workers.
      *
+     * @param device the device to get the min workers for
      * @return the minimum number of workers
      */
-    public int getMinWorkers() {
-        return minWorkers;
+    public int getMinWorkers(Device device) {
+        if (minWorkers != null && minWorkers >= 0) {
```
frankfliu, line 482: should this be minWorkers > 0?

I think we need at least one worker

zachgk: This is really a check for null and `-1` which are the "use default" values. I can add that validation, although I would add it on the side of setting the minWorkers and maxWorkers. I agree that `maxWorkers > 0`, but there is a case for `minWorkers = 0`. For the use case where you have many CPU models with low usage of each, there are two approaches:

1. Allow `minWorkers == 0` and spin up the worker when the model is first called. 
2. Require the `minWorkers > 0` which would result in large numbers of workers (`workers >> numCpus`) and most of those workers should just remain idle.

Have we compared the two approaches?

frankfliu: Ideally, we should allow minWorkers == 0, but it requires we support scale to 0, which we don't currently.

frankfliu (approved):

PR #1022, opened 2023-08-18
Worker type

zachgk:
This is a decent change I wanted to get some early feedback on. It still needs a
bit of clean up, and I am not quite happy with the name of "workerType" (let me
know if anyone has any better ideas).

The basic idea is that I replaced the `ModelInfo` inside of the WLM with an
abstraction called `WorkerType`. The WorkerType corresponds to a WorkerPool and
determines what data is saved and work is done within a `WorkerThread`. The main
WorkerType is still the ModelInfo which extends it. I also added a
LambdaWorkerType that can be used for other processing. Unlike a
WorkflowFunction, it runs in a dedicated worker thread rather than on the common
CompleteableFuture pool.

I think this also helps clean up the abstractions a bit. Within the ModelInfo,
it essentially removed all of the pieces which are more about interacting with
DJL Serving and the WLM more than deep learning. It should also help define
clearer boundaries for when we add expansions and other pieces to the model such
as adapters. I'll also update the architecture section after this (which
probably needed an update regardless).

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/Job.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/Job.java
@@ -15,30 +15,30 @@ package ai.djl.serving.wlm;
 /** A class represents an inference job. */
 public class Job<I, O> {
 
-    private ModelInfo<I, O> modelInfo;
+    private WorkerPoolConfig<I, O> workerPoolConfig;
     private I input;
     private long begin;
 
     /**
      * Constructs a new {@code Job} instance.
      *
-     * @param modelInfo the model to run the job
+     * @param wpc the model to run the job
      * @param input the input data
      */
-    public Job(ModelInfo<I, O> modelInfo, I input) {
-        this.modelInfo = modelInfo;
+    public Job(WorkerPoolConfig<I, O> wpc, I input) {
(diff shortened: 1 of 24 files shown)


wlm/src/main/java/ai/djl/serving/wlm/Job.java
```diff
@@ -15,30 +15,30 @@
...
     /**
-     * Returns the model that associated with this job.
+     * Returns the worker pool config that is associated with this job.
      *
-     * @return the model that associated with this job
+     * @return the worker pool config that is associated with this job
      */
-    public ModelInfo<I, O> getModel() {
-        return modelInfo;
+    public WorkerPoolConfig<I, O> getWpc() {
```
frankfliu, line 40: the same for this function

zachgk: This one makes more sense as getWpc. If we were to later change it to have the model as a property on the wpc, then this would still be correct

frankfliu: Agree, maybe use `getWorkerPoolConfig()`

frankfliu (approved):

PR #1040, opened 2023-08-28
[docker] add version labels for sagemaker

tosterberg:
## Description ##

Adding labels with djl version in them for sagemaker telemetry.


rohithkrn:
Could you elaborate more why `djl_version` is not used as a value with label name being constant across versions? This change means we will have to update the label for each version?

PR #1043, opened 2023-08-30
[fix] Two major bugs in rolling_batch_scheduler

KexinFeng:
Two major bugs:
1. get_new_requests. the loop control bug; the input_data is not correctly indexed, causing missing input.
2. postprocess_results. bug: only output results when there is new requests. When there is no new_requests the output are skipped.

Both are effective in lmi_dist_scheduler. Might produce the buggy results shown in the ticket.

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -178,6 +178,7 @@ class RollingBatch(ABC):
                               params if params else {})
             self.pending_requests.append(request)
             self.req_id_counter += 1
+
             if self.max_batch_size is not None and len(
                     self.pending_requests) == self.max_batch_size:
                 warnings.warn(


engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -168,34 +171,36 @@ def inference(self, input_data, parameters):
...
+        init_batch_size = len(self.pending_requests)
+        for data, params in zip(input_data, parameters):
+            request = Request(self.req_id_counter, data,
+                              params if params else {})
+            self.pending_requests.append(request)
+            self.req_id_counter += 1
+
+            if self.max_batch_size is not None and len(
+                    self.pending_requests) == self.max_batch_size:
+                warnings.warn(
```
frankfliu, line 184: Avoid using `warnings`, user can change logging configuration.
do you really want to just logging or you want to throw an exception?

frankfliu: frontend should ensure the batch size is never exceed that max, did you ever seen this happen?

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -168,34 +171,36 @@ def inference(self, input_data, parameters):
...
-                request = Request(self.req_id_counter, data, params)
-                self.pending_requests.append(request)
-                new_requests.append(request)
-                self.req_id_counter += 1
+    def get_new_requests(self, input_data, parameters):
+        init_batch_size = len(self.pending_requests)
+        for data, params in zip(input_data, parameters):
+            request = Request(self.req_id_counter, data,
+                              params if params else {})
+            self.pending_requests.append(request)
```
sindhuvahinis, line 179: QQ: Here all the incoming input_data is new requests here? I thought in the front end we have a list of all the requests that are not yet completed.

sindhuvahinis: Got it, Yeah so front end has a list of all unfinished requests and append new requests to the list. It would send the whole input str for the new requests, but empty strings for already pending requests. instead of None, like this 
`"inputs": [""]`

Ref: https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java#L200

Try with this and check whether it works. 
```
for _ in range(50):
    result = rolling_batch.inference(["", ""], [{}, {}])
    for i, res in enumerate(result):
        output_all[i].append(res['data']) 
    print(' '.join(out) for out in output_all)
```

PR #1064, opened 2023-09-08
[Handler] make huggingface streamer default

lanking520:
## Description ##

Make huggingface streamer used by default

### DIFF
--- a/engines/python/setup/djl_python/transformers-neuronx.py
+++ b/engines/python/setup/djl_python/transformers-neuronx.py
@@ -229,18 +229,18 @@ class TransformersNeuronXService(object):
 
             if self.enable_streaming:
                 outputs.add_property("content-type", "application/jsonlines")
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/transformers-neuronx.py
```diff
@@ -229,18 +229,18 @@ def infer(self, inputs):
 
             if self.enable_streaming:
                 outputs.add_property("content-type", "application/jsonlines")
-                if self.enable_streaming == "huggingface":
-                    outputs.add_stream_content(
-                        StreamingUtils.use_hf_default_streamer(
-                            self.model, self.tokenizer, input_text, None,
-                            **model_kwargs))
-                else:
+                if self.enable_streaming == "local":
```
frankfliu, line 232: 1. We should find a better name.
2. Why huggingface is better? In which case user should use "local", should we drop "local"?

PR #1073, opened 2023-09-09
[docker] Change default OMP_NUM_THREADS back to 1 for GPU

frankfliu:
## Description ##

### DIFF
--- a/serving/docker/pytorch-inf2.Dockerfile
+++ b/serving/docker/pytorch-inf2.Dockerfile
@@ -30,7 +30,7 @@ COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
 RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
 WORKDIR /opt/djl
 ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
-ENV NO_OMP_NUM_THREADS=true
+# ENV NO_OMP_NUM_THREADS=true
 ENV MODEL_SERVER_HOME=/opt/djl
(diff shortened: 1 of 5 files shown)


lanking520 (commented):
Where do you set to 1?

frankfliu:
It's set in the startup shell script: https://github.com/deepjavalibrary/djl-serving/blob/master/serving/build.gradle#L109-L110

lanking520:
OK, then you cannot enable OMP_TREAD=1 for INF2, it will slow down SD inference speed.

frankfliu:
The problem is what should be the default value.

without OMP_NUM_THREADS=1:
1. most of CV models will be suffered
2. LLM is not impacted by OMP_NUM_THREADS
3. SD model has sighttly better performance

I think we should document this for SD model and use OMP_NUM_THREADS=1 by default

lanking520:
I would suggest you to set all other containers to OMP_THREAD=1 besides Neuron container.

serving/docker/pytorch-inf2.Dockerfile
```diff
@@ -30,7 +30,7 @@ COPY dockerd-entrypoint.sh /usr/local/bin/dockerd-entrypoint.sh
 RUN chmod +x /usr/local/bin/dockerd-entrypoint.sh
 WORKDIR /opt/djl
 ENV JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64
-ENV NO_OMP_NUM_THREADS=true
+# ENV NO_OMP_NUM_THREADS=true
```
lanking520, line 33: let's keep INF2 change as is. And rest looks all good

lanking520 (approved):

PR #1074, opened 2023-09-11
[fix] Fix device map

KexinFeng:
1. AttributeError: 'str' object has no attribute 'type'
2. TypeError: transformers.models.auto.auto_factory._BaseAutoModelClass.from_pretrained() got multiple values for keyword argument 'device_map'

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
@@ -104,7 +104,10 @@ class SchedulerRollingBatch(RollingBatch):
                     device_map = device
                 elif isinstance(device,
                                 torch.device) and device.type == "cuda":
+                    # TODO: enable specifying the cuda device
                     device_map = 'auto'
+                else:
+                    raise Exception("Wrong input type of device")
             if 'device_map' in kwargs:
                 device_map = kwargs.pop('device_map')
 


engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
```diff
@@ -98,11 +98,18 @@ def _init_model_and_tokenizer(self,
                 "ForConditionalGeneration"):
             raise ValueError('Seq2Seq model is not supported by scheduler')
         else:
+            device_map = "cpu"
+            if device:
+                if isinstance(device, str):
+                    device_map = device
```
lanking520, line 104: can you pass device_map as 0 or 1 when we consider different GPU?

lanking520: no, there could be a scenario that the input device is the specific GPU ID, did we offer anything could work on this value?

lanking520: device=0

lanking520: anyway, we can leave this later since DJLServing serving layer could cover it. Could you add a TODO in here to support specific GPU IDs?

engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
```diff
@@ -98,11 +98,18 @@ def _init_model_and_tokenizer(self,
...
+                elif isinstance(device,
+                                torch.device) and device.type == "cuda":
+                    device_map = 'auto'
```
lanking520, line 107: add an else below to throw an Error

lanking520 (approved):

PR #1080, opened 2023-09-14
Fixes API responses

zachgk:
This is an update to fix the changes to the workflow response from https://github.com/deepjavalibrary/djl-serving/pull/1022. Before, we had no test that checks the response components decided by the names of Java response object properties (workflows, models, etc.). This adds some new JSON testing to help ensure that these are not accidentally changed, fixing this issue and hopefully avoiding future breakages.

### DIFF
--- a/serving/src/test/java/ai/djl/serving/ModelServerTest.java
+++ b/serving/src/test/java/ai/djl/serving/ModelServerTest.java
@@ -38,6 +38,7 @@ import ai.djl.util.Utils;
 import ai.djl.util.ZipUtils;
 import ai.djl.util.cuda.CudaUtils;
 
+import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
 import com.google.gson.reflect.TypeToken;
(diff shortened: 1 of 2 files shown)


serving/src/test/java/ai/djl/serving/ModelServerTest.java
```diff
@@ -596,6 +600,16 @@ private void testListModels(Channel channel) throws InterruptedException {
                             .anyMatch(w -> expectedModel.equals(w.getModelName())));
         }
 
+        // Test pure JSON works
+        Map<String, Object> rawResult = JsonUtils.GSON.fromJson(result, Map.class);
```
frankfliu, line 604: the result has been deserialized into `resp` object, can we just assert each fields are not null?

zachgk: I'm not sure that helps. The issue was that I changed the property name of the list from `models` to `wpcs`. This changes both the serialization in the code and deserialization in the test, so no explicit changes were necessary in the test. I wanted to use something more raw so that the names of the fields explicitly show up in our testing

frankfliu (approved):

PR #1082, opened 2023-09-15
Adds adapter support

zachgk:
This PR serves to add adapter support. It is supported with two different customer experiences which can be seen in the ModelServerTest file. The first is through the new adapter management API and adding an adapter query parameter to prediction calls. The second is by creating a workflow with the AdapterWorkflowFunction.

Both are testing using the adaptecho model. It is a simple python echo model that prepends the adapter contents. It can be used as a simple example of how the adapter should function for a model handler. Note the usage of additional functions register_adapter and unregister_adapter alongside the main handler.

Then, there is a new Adapter abstract class. It only has a PyAdapter implementation for now. In the future, this should depend on the Patch from DJL and will require an addition Engine.loadAdapter(name, url) function to create the engine specific patches. This is left as a future improvement. For now, it has functionality to both register and unregister the adapter through the predictor. It also has helpers to integrate the registration and unregistration flows alongside both the API and workflow approach.

After that, those Adapters are stored in the ModelInfo. It adds a new concept of configJobs that are run on worker threads. The ModelInfo will create config jobs to register all adapters on thread creation and to register new adapters on existing workers.

To support the API approach, the main changes are through the new AdapterManagementRequestHandler. They are all testing in the ModelServerTest.

To support the workflow approach, it begins by making both workflows and workflow functions contain both a prepare and close function. Preparing the AdapterWorkflowFunction will register it and closing it will unregister it. The test features two workflows with a shared model (adaptecho) that are loaded together.

Another miscellaneous improvement to mention is the change to Job. Job has two new arguments. The first is a Job runner that determines what will run. It can either take the default behavior of executing the predictor through the ThreadConfig or contain custom behavior (register and unregister). The second change is that Output is stored in the job rather than returned directly. This helps deal with some job runners that may be successful for part of the batch but fail with other parts. It also cleans up the logic with respect to grouping bu job runner. The JobFunction is also a new functional interface to handle exceptions.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
+++ b/serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.serving.http;
+
+import ai.djl.ModelException;
+import ai.djl.modality.Input;
+import ai.djl.modality.Output;
+import ai.djl.serving.http.list.ListAdaptersResponse;
+import ai.djl.serving.http.list.ListPagination;
+import ai.djl.serving.models.ModelManager;
+import ai.djl.serving.util.NettyUtils;
+import ai.djl.serving.wlm.Adapter;
+import ai.djl.serving.wlm.ModelInfo;
+import ai.djl.serving.wlm.WorkerPool;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.QueryStringDecoder;
+
+import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/** A class handling inbound HTTP requests to the management API for adapters. */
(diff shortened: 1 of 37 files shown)


wlm/src/main/java/ai/djl/serving/wlm/Adapter.java
```diff
@@ -0,0 +1,172 @@
...
+     *
+     * @param wp the worker pool to remove the adapter from
+     * @param adapterName the adapter name
+     * @param <I> the input type
+     * @param <O> the output type
+     */
+    public static <I, O> void unregister(WorkerPool<I, O> wp, String adapterName) {
+        ModelInfo<I, O> wpc = (ModelInfo<I, O>) wp.getWpc();
+        Adapter adapter = wpc.unregisterAdapter(adapterName);
+        for (WorkerGroup<I, O> wg : wp.getWorkerGroups().values()) {
```
lanking520, line 79: maybe add a TODO to allow model-instance level adapter scheduling instead of register/unregister all.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -234,6 +244,9 @@ public void load(Device device) throws ModelException, IOException {
             }
 
             ZooModel<I, O> m = builder.build().loadModel();
+            if (engineName == null) {
+                engineName = m.getNDManager().getEngine().getEngineName();
```
lanking520, line 248: check if this collide with the `HF_MODEL_ID` design in LLM inference for infer engine

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
```diff
@@ -0,0 +1,166 @@
...
+import io.netty.handler.codec.http.QueryStringDecoder;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.regex.Pattern;
+
+/** A class handling inbound HTTP requests to the management API for adapters. */
+public class AdapterManagementRequestHandler extends HttpRequestHandler {
+    private static final Pattern ADAPTERS_PATTERN =
+            Pattern.compile("^/models/.+/adapters([/?][^/]*)?");
```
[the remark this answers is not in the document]
zachgk: Yeah, I will put documents in the next PR because there is a good amount required:

- Document HTTP adapters management API
- Document adapter usage
- Document adapter creation
- Document adapter options and configurations

lanking520 (approved):

PR #1086, opened 2023-09-15
[feature] Enable model sharding on seq_scheduler tested on gpt_neox_20B

KexinFeng:
## Description ##

1. This depends on the package `lmi_dist`.

`test_rolling_batch_scheduler2.py` and `test_scheduler_sharded.py` are demos

2. Fix the support for Falcon type of model

3. Fix the bug that the pad_token_id is not consistent with the tokenizer in seq_scheduler_rolling_batch

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
@@ -116,7 +116,7 @@ class SchedulerRollingBatch(RollingBatch):
             if 'device_map' in kwargs:
                 device_map = kwargs.pop('device_map')
 
-            if "tgi_sharding" == multi_gpu:
+            if "lmi_dist_sharding" == multi_gpu:
                 if 'neox' in model_id_or_path:
                     try:
                         from lmi_dist.models.gpt_neox import GPTNeoxSharded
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
```diff
@@ -111,17 +113,30 @@ def _init_model_and_tokenizer(self,
             if 'device_map' in kwargs:
                 device_map = kwargs.pop('device_map')
 
-            self.model = AutoModelForCausalLM.from_pretrained(
-                model_id_or_path, device_map=device_map, **kwargs)
+            if 'neox' in model_id_or_path and int(
```
lanking520, line 116: you can just grab the model from lmi_dist init function right?

lanking520: could we add some special flags to use LMI dist library? Because sometimes we need fallback option of using HF accelerate

engines/python/setup/djl_python/rolling_batch/scheduler_rolling_batch.py
```diff
@@ -111,17 +116,35 @@ def _init_model_and_tokenizer(self,
             if 'device_map' in kwargs:
                 device_map = kwargs.pop('device_map')
 
-            self.model = AutoModelForCausalLM.from_pretrained(
-                model_id_or_path, device_map=device_map, **kwargs)
+            if "tgi_sharding" == multi_gpu:
```
lanking520, line 119: let's call this lmi_dist_sharding

lanking520 (approved):

PR #1088, opened 2023-09-17
Add unmerged lora support in HF handler

rohithkrn:
## Description ##

Add unmerged lora support in HF handler. Handler expects extra `adapters` parameter to be passed to it.

E.g request: 
```
curl -X POST http://127.0.0.1:8080/invocations \
    -H "Content-Type: application/json" \
    -H "X-Amzn-SageMaker-Target-Model: base-1.tar.gz" \
    -d '{"inputs": ["How is the weather"], "adapters": ["adapter_1"], "parameters": {"max_new_tokens": 25}}'
```

Added two new functions which can be called to load and unload adapters
1. register_adapter
2. unregister_adapter

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -21,7 +21,7 @@ from transformers import (pipeline, Pipeline, Conversation, AutoModelForCausalLM
                           AutoModelForTokenClassification,
                           AutoModelForQuestionAnswering)
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
-from peft import PeftConfig, PeftModel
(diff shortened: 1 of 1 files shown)


siddvenk:
Are we able to support different generation parameters per adapter/request in a batch?

rohithkrn:
> Are we able to support different generation parameters per adapter/request in a batch?

No, all adapters will need to have same generation config. For the dynamic batching case that is already enforced in the current handler.

zachgk:
Can we add testing with this?

rohithkrn:
> Can we add testing with this?

we don't have unit tests for handlers at this point. I wanted to add integration tests which I can do after the frontend changes are merged so that I can tests loading and unloading adapters.

zachgk (approved):

PR #1094, opened 2023-09-19
Creates adapters by directory

zachgk:
This enables support for automatically adding a directory of adapters. It will
be read on the first load and utilize the existing registration flow.

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -248,6 +248,24 @@ public final class ModelInfo<I, O> extends WorkerPoolConfig<I, O> {
             if (engine == null) {
                 engine = m.getNDManager().getEngine();
             }
+
+            if (models.isEmpty()) {
+                // Check for adapters on first load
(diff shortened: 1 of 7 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -248,6 +248,24 @@ public void load(Device device) throws ModelException, IOException {
             if (engine == null) {
                 engine = m.getNDManager().getEngine();
             }
+
+            if (models.isEmpty()) {
```
rohithkrn, line 252: why are we checking for this condition?

rohithkrn:
also, we would not be unregistering adapters with adapter dir approach right?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
zachgk: This models is specifically a map from device to model. So, it checks that it is empty to ensure that this loading only happens on the first device load instead of every new device you add the model to

zachgk:
You can unregister these adapters using the management API

rohithkrn (approved):

PR #1101, opened 2023-09-21
[feature] Test llama-7b-gptq on scheduler_rolling_batch

KexinFeng:
## Description ##

### DIFF
--- a/.github/workflows/rolling_batch_integration.yml
+++ b/.github/workflows/rolling_batch_integration.yml
@@ -85,6 +85,15 @@ jobs:
           serve -m test=file:/opt/ml/model/test/
           python3 rb_client.py scheduler_single_gpu bloom-560m
           docker rm -f $(docker ps -aq)
+      - name: Test llama2-7b-chat-gptq
+        working-directory: tests/integration
+        run: |
+          rm -rf models
+          python3 llm/prepare.py rolling_batch_scheduler llama2-7b-chat-gptq
+          ./launch_container.sh deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG $PWD/models deepspeed \
+          serve -m test=file:/opt/ml/model/test/
+          python3 rb_client.py scheduler_single_gpu llama2-7b-chat-gptq
+          docker rm -f $(docker ps -aq)
       - name: Print outputs
         working-directory: tests/integration
         run: for file in outputs/*; do if [ -f "$file" ]; then echo "Contents of $file:"; cat "$file"; echo; fi; done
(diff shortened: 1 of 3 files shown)


lanking520 (commented):
could you add this test to our CI testing pipelines?

tests/integration/llm/prepare.py
```diff
@@ -380,6 +380,12 @@
         "option.max_rolling_batch_size": 4,
         "option.tensor_parallel_degree": 4,
         "load_on_devices": 0
+    },
+    "llama2-7b-chat-gptq": {
+        "option.model_id": "TheBloke/Llama-2-7b-Chat-GPTQ",
+        "engine": "Python",
+        "option.max_rolling_batch_size": 4,
```
lanking520, line 387: do we need this settings for scheduler?

.github/workflows/rolling_batch_integration.yml
```diff
@@ -85,6 +85,15 @@ jobs:
           serve -m test=file:/opt/ml/model/test/
           python3 rb_client.py scheduler_single_gpu bloom-560m
           docker rm -f $(docker ps -aq)
+      - name: Test llama2-7b-chat-gptq
+        working-directory: tests/integration
+        run: |
+          rm -rf models
+          python3 llm/prepare.py rolling_batch_scheduler llama2-7b-chat-gptq
```
lanking520, line 92: will this test work if we didn't install `autogptq` and `optimum` ?

tests/integration/llm/prepare.py
lanking520: I don't think this is necesary. If you specify rolling batch

.github/workflows/rolling_batch_integration.yml
lanking520: You can refer to this: https://github.com/deepjavalibrary/djl-serving/blob/master/tests/integration/llm/prepare.py#L649

lanking520: you can pass with requirements.txt to install those dependencies on the fly

tests/integration/llm/prepare.py
lanking520: This is a different error, the name you give to the model is wrong

tests/integration/llm/prepare.py
```diff
@@ -380,6 +380,12 @@
         "option.max_rolling_batch_size": 4,
         "option.tensor_parallel_degree": 4,
         "load_on_devices": 0
+    },
+    "llama2-7b-chat-gptq": {
```
lanking520, line 384: you are now putting them in a CI, but not installing the necessary wheel for the test. How this will work?

lanking520 (changes requested):
Please make sure the CI test are passable

lanking520 (approved):

PR #1109, opened 2023-09-25
Creates auto increment ID for models

zachgk:
This adds an auto increment ID to models, their toString, and logging. It allows them to be differentiated if two models share the same name. It is unified with the existing auto increment system in worker threads (which also had it's logging improved).

The IDs have both a prefix and leading zeros for consistent search:
WT-0001 for worker threads
M-0001 for models


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -97,13 +97,18 @@ public final class ModelInfo<I, O> extends WorkerPoolConfig<I, O> {
     private transient Map<String, Adapter> adapters;
     private transient Engine engine;
 
+    private ModelInfo() {
+        super();
```
frankfliu, line 101: `super()` is not necessary, default constructor will be invoked automatically.

siddvenk (approved):

PR #1112, opened 2023-09-25
[vLLM] capture max_rolling_batch settting issues

lanking520:
## Description ##

Bug fix on the integer parsing

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
@@ -27,20 +27,18 @@ class VLLMRollingBatch(RollingBatch):
         :param properties: other properties of the model, such as decoder strategy
         """
         super().__init__(-1, **kwargs)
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
```diff
@@ -27,21 +27,19 @@ def __init__(self, model_id_or_path, device, properties, **kwargs):
         :param properties: other properties of the model, such as decoder strategy
         """
         super().__init__(-1, **kwargs)
-        self.dtype = kwargs.pop("dtype", 'auto')
+        self.dtype = properties.get("dtype", 'auto')
```
rohithkrn, line 30: should we also change handlers to pass these parameters in `properties`?

lanking520: they already in properties

rohithkrn (approved):

PR #1126, opened 2023-09-29
[INF2] remove neuron settings on cache hit for the folder

lanking520:
## Description ##

Based on the current setup through model type and tensor parallel degree. It is not possible to find out the diff between LLAMA-7B TP2 FP16 vs LLAMA 13B TP2 FP16. Fallback to the inefficient way of reloading.


lanking520:
@tosterberg

tosterberg (approved):

PR #1134, opened 2023-10-02
[INF2][Handler] remove type conversion in Neuron

lanking520:
## Description ##

Remove all type conversion since it is handled by neuron model loading script.

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx.py
+++ b/engines/python/setup/djl_python/transformers_neuronx.py
@@ -15,9 +15,9 @@ import shutil
 import tempfile
 import os
 import logging
+import time
 
 from transformers import AutoModelForCausalLM, AutoConfig, AutoTokenizer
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -176,6 +139,8 @@ def load_model(self, model_type):
         else:
             self.model.to_neuron()
         os.chdir(path)
+        elapsed = time.time() - start
+        logging.info(f"Compilation completed with {elapsed}s")
```
aws-bowencc, line 143: The info message is misleading here, I would simply replace it with `to_neuron completed with...`. `to_neuron` contains more than just compilation, but also weight loading and sharding.

lanking520: Thank you, addressed.

zachgk (approved):

PR #1137, opened 2023-10-03
[fix] Gptq dependency

KexinFeng:
It is confirmed that the previous test on llama2-gptq passed in an old version of deepspeed-nightly docker image. After updating the deepspeed-nightly image, the dependency error is reproduced. 

The fix is to install the two required packages: auto-gptq and optimum. The unittests all pass now in the latest deepspeed-nightly docker image.


davidthomas426 (approved):

lanking520 (approved):

lanking520:
Do a git pull --rebase upstream master

PR #1138, opened 2023-10-03
[feat] Modify deepspeed handler to support smoothQuant.

chen3933:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -10,6 +10,15 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
+
+# smoothquant reference -
+# @InProceedings{xiao2023smoothquant,
+#     title = {{S}mooth{Q}uant: Accurate and Efficient Post-Training Quantization for Large Language Models},
+#     author = {Xiao, Guangxuan and Lin, Ji and Seznec, Mickael and Wu, Hao and Demouth, Julien and Han, Song},
+#     booktitle = {Proceedings of the 40th International Conference on Machine Learning},
+#     year = {2023}
+# }
+
 import logging
 import os
 import json
@@ -28,6 +37,20 @@ from djl_python.outputs import Output
 from djl_python.streaming_utils import StreamingUtils
 from typing import Optional
 from peft import PeftConfig, PeftModel
+
+SUPPORTED_QUANTIZATION_MODE = [
+    "smoothquant",
+    "dynamic_int8"
+]
+
+SMOOTHQUANT_SUPPORTED_MODEL_TYPES = {
+    "gpt2",
+    "gpt_neo",
+    "gptj",
+    "gpt_neox",
+    "bloom",
+    "llama",
+}
 
 OPTIMIZED_MODEL_TYPES = {
     "roberta",
@@ -119,6 +142,9 @@ class DeepSpeedService(object):
         self.tokenizer = None
         self.revision = None
 
+        self.quantize_mode = None
+        self.smoothquant_alpha = None
+
     def initialize(self, properties: dict):
         self._parse_properties(properties)
         self._read_model_config()
@@ -155,11 +181,32 @@ class DeepSpeedService(object):
                 "trust_remote_code").lower() == "true"
         if "revision" in properties:
             self.revision = properties['revision']
+
+        # SmoothQuant properties
+        self._parse_smoothquant_properties(properties)
+
         if properties.get("deepspeed_config_path"):
             with open(properties.get("deepspeed_config_path"), "r") as f:
                 self.ds_config = json.load(f)
         else:
             self.ds_config = self._get_ds_config(properties)
+
+    def _parse_smoothquant_properties(self, properties):
+        if 'quantize' in properties:
+            if properties['quantize'] not in SUPPORTED_QUANTIZATION_MODE:
+                self.logger.warning(
+                    f"DeepSpeed does not currently support quantization mode: ${properties['quantize']}, "
+                    f"this setting will be ignored.")
+                return
+            self.quantize_mode = properties['quantize']
+        if 'smoothquant_alpha' in properties:
+            try:
+                if 0 <= float(properties['smoothquant_alpha']) <= 1:
+                    self.smoothquant_alpha = properties['smoothquant_alpha']
+                else:
+                    raise ValueError(f"${properties['smoothquant_alpha']} is not between 0 and 1.")
+            except ValueError:
+                raise ValueError(f"${properties['smoothquant_alpha']} cannot convert to float number.")
 
     def _get_ds_config(self, properties: dict):
         ds_config = {
@@ -186,6 +233,17 @@ class DeepSpeedService(object):
             if self.data_type is None:
                 raise ValueError(
                     "dtype should also be provided for checkpoint loading")
+
+        if self.quantize_mode:
+            ds_config['dynamic_quant'] = {'enabled': True, 'use_cutlass': False}
+            if self.quantize_mode == 'smoothquant':
+                smoothing_value = {
+                    'smooth': True,
+                    'calibrate': True
+                }
+                if self.smoothquant_alpha:
+                    smoothing_value['alpha'] = self.smoothquant_alpha
+                ds_config['smoothing'] = smoothing_value
         return ds_config
 
     def _validate_model_type_and_task(self):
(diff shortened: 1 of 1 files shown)


maaquib (approved):

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -155,12 +182,35 @@ def _parse_properties(self, properties):
...
+                    f"DeepSpeed does not currently support quantization mode: ${properties['quantize']}, "
+                    f"this setting will be ignored.")
+                return
+            self.quantize_mode = properties['quantize']
+        if 'smoothquant_alpha' in properties:
+            try:
+                if 0 <= float(properties['smoothquant_alpha']) <= 1:
+                    self.smoothquant_alpha = properties['smoothquant_alpha']
+                else:
+                    self.logger.warning(f"${properties['smoothquant_alpha']} is not between 0 and 1, "
```
lanking520, line 208: We should throw an exception here?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -155,12 +182,35 @@ def _parse_properties(self, properties):
...
+                                        "DeepSpeed will use default alpha setting.")
+            except ValueError:
+                self.logger.warning(f"${properties['smoothquant_alpha']} is not a float number, "
```
lanking520, line 211: Throw exception instead

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -236,17 +300,71 @@ def infer_task_from_model_architecture(self, config: PretrainedConfig):
...
+        model = None
+        tokenizer = None
+        done = False
+
+        if use_mmap_loader:
+            try:
+                model, tokenizer, state_dict_mmap = load_model_with_mmap(model_id_or_path, loading_method)
+                done = True
+            except:
+                self.logger.warning(f'failed to load model with mmap loader, will load model normally')
```
lanking520, line 341: When do we enter here? What if the loading get into unrecoverable states?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -259,28 +377,25 @@ def create_model_pipeline(self):
...
-        if self.model_config.model_type in OPTIMIZED_MODEL_TYPES:
-            self.ds_config["replace_with_kernel_inject"] = True
-        self.model = deepspeed.init_inference(model, config=self.ds_config)
-        if self.peft_config:
-            self.tokenizer = AutoTokenizer.from_pretrained(
-                self.peft_config.base_model_name_or_path)
-        else:
-            self.tokenizer = AutoTokenizer.from_pretrained(
-                self.model_id_or_path)
+        self.model = deepspeed.init_inference(model, state_dict_mmap, config=self.ds_config)
```
lanking520, line 397: Totally getting rid of peft Lora?

lanking520 (commented):
Please address these comments

engines/python/setup/djl_python/deepspeed.py
chen3933: `elif self.peft_config is not None:`(ln 368 - ln 384) block should work as before.

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -236,17 +301,71 @@ def infer_task_from_model_architecture(self, config: PretrainedConfig):
                 f"Task could not be inferred from model config. "
                 f"Please manually set `task` in serving.properties.")
 
+    def get_model_pretrained(self, model_id_or_path, torch_dtype='auto', **kwargs):
+        tokenizer = AutoTokenizer.from_pretrained(model_id_or_path)
```
lanking520, line 305: please add some logic to allow `trust_remote_code` option

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -236,17 +301,71 @@ def infer_task_from_model_architecture(self, config: PretrainedConfig):
...
+        model = TASK_TO_MODEL[self.task].from_pretrained(model_id_or_path, torch_dtype=torch_dtype, **kwargs)
+        return model, tokenizer
+
+    def get_model_from_config(self, model_id_or_path, **kwargs):
+        tokenizer = AutoTokenizer.from_pretrained(model_id_or_path)
```
lanking520, line 310: please add some logic to allow `trust_remote_code` option

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -236,17 +301,71 @@ def infer_task_from_model_architecture(self, config: PretrainedConfig):
...
+        model = TASK_TO_MODEL[self.task].from_config(self.model_config, **kwargs)
+        return model, tokenizer
+
+    def get_model(self, model_id_or_path, loading_method, **kwargs):
+        if loading_method == 'from_config':
+            return self.get_model_from_config(model_id_or_path, **kwargs)
+        elif loading_method == 'pretrained':
```
lanking520, line 317: this logic seemed to be the same? just 

```
if loading_method in ['from_config', 'pretrained']:
```

lanking520 (changes requested):

engines/python/setup/djl_python/deepspeed.py
lanking520: `revision`

engines/python/setup/djl_python/deepspeed.py
lanking520: also need to take care `revision`

chen3933: trust_remote_code and other parameter will get merged into **kwargs.
For example:
on line 384 and 391, 
```
            model, self.tokenizer, state_dict_mmap = self.load_model(
                self.model_id_or_path,
                'pretrained',
                self.ds_config['replace_with_kernel_inject'],
                low_cpu_mem_usage=self.low_cpu_mem_usage,
                trust_remote_code=self.trust_remote_code,
                **kwargs
            )
```
both 
```
low_cpu_mem_usage=self.low_cpu_mem_usage,
trust_remote_code=self.trust_remote_code,
```
will merge into **kwargs in self.load_model function
```
 def load_model(self, model_id_or_path, loading_method, use_mmap_loader, **kwargs):
```

engines/python/setup/djl_python/deepspeed.py
chen3933: Same as above

engines/python/setup/djl_python/deepspeed.py
chen3933: For AutoTokenizer, I didn't see **kwargs get passed in, following are the original code.
```
        if self.peft_config:
            self.tokenizer = AutoTokenizer.from_pretrained(
                self.peft_config.base_model_name_or_path)
        else:
            self.tokenizer = AutoTokenizer.from_pretrained(
                self.model_id_or_path)
```

engines/python/setup/djl_python/deepspeed.py
chen3933: Oh man, good catch. line 316 should call `self.get_model_pretrained`
Will fix this typo in next iteration.

engines/python/setup/djl_python/deepspeed.py
davidthomas426: The loading fails because:
* Not all model formats are supported. We could limit it to ValueError or something if you'd prefer to cover this case at least.
* mmaploader is not available or fails to import for some reason.

davidthomas426: If the state is "unrecoverable", how do you think the behavior would be better than what will happen if we just try falling back to normal model loading?

engines/python/setup/djl_python/deepspeed.py
davidthomas426: Did that not work before? What changed?

davidthomas426: Is that covered by `kwargs`?

lanking520: For example, model like https://huggingface.co/Salesforce/codegen25-7b-multi used a custom tokenizer rather than the built in one. So `revision` and `trust_remote_code` is necessary for the inputs

lanking520 (approved):
Let's capture those tokenizer gaps in 1 go in the next PR. LGTM

lanking520:
@chen3933 try to do a squash on your commits

engines/python/setup/djl_python/deepspeed.py
chen3933: Will address this use case in the followup PR. Probably need some refactor of the code

chen3933:
squashed commits. please don't push yet. Code still under testing.

lanking520:
> squashed commits. please don't push yet. Code still under testing.

added an WIP in title, feel free to remove when it is ready

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -129,6 +155,7 @@ def initialize(self, properties: dict):
             f"model: {self.model_id_or_path}\n"
             f"task: {self.task}\n"
             f"data_type: {self.ds_config['dtype']}\n"
+            f"deepspeed config: {self.ds_config}\n"
```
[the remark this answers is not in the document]
chen3933: Removed

PR #1139, opened 2023-10-03
Adding smoothquant integ tests

maaquib:
## Description ##

### DIFF
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -73,7 +73,7 @@ RUN apt-get update && \
     ${deepspeed_wheel} ${flash_attn_wheel} ${dropout_layer_norm_wheel} ${rotary_emb_wheel} ${flash_attn_2_wheel} \
     ${vllm_wheel} ${lmi_dist_wheel} ${seq_scheduler_wheel} ${peft_wheel} ${mmaploader_wheel} protobuf==${protobuf_version} \
     transformers==${transformers_version} zstandard datasets==${datasets_version} \
-    mpi4py sentencepiece einops accelerate==${accelerate_version} bitsandbytes==${bitsandbytes_version \
+    mpi4py sentencepiece einops accelerate==${accelerate_version} bitsandbytes==${bitsandbytes_version} \
     optimum==${optimum_version} auto-gptq==${auto_gptq_version} \
     diffusers[torch]==${diffusers_version} opencv-contrib-python-headless safetensors scipy && \
     scripts/install_aitemplate.sh && \
--- a/serving/docker/deepspeed.Dockerfile
+++ b/serving/docker/deepspeed.Dockerfile
@@ -32,7 +32,6 @@ ARG diffusers_version=0.16.0
 ARG bitsandbytes_version=0.41.1
 ARG optimum_version=1.13.2
 ARG auto_gptq_version=0.4.2
-ARG cutlass_version=3.1.0
 ARG datasets_version=2.14.4
 
 EXPOSE 8080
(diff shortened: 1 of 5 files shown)


serving/docker/deepspeed.Dockerfile
```diff
@@ -69,8 +71,8 @@ RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -yq libaio-dev libopenmpi-dev && \
     pip3 install torch==${torch_version} torchvision==${torch_vision_version} --extra-index-url https://download.pytorch.org/whl/cu118 \
     ${deepspeed_wheel} ${flash_attn_wheel} ${dropout_layer_norm_wheel} ${rotary_emb_wheel} ${flash_attn_2_wheel} \
-    ${vllm_wheel} ${lmi_dist_wheel} ${seq_scheduler_wheel} ${peft_wheel} protobuf==${protobuf_version} \
-    transformers==${transformers_version} \
+    ${vllm_wheel} ${lmi_dist_wheel} ${seq_scheduler_wheel} ${peft_wheel} ${mmaploader_wheel} protobuf==${protobuf_version} \
+    transformers==${transformers_version} zstandard datasets==${datasets_version} \
     mpi4py sentencepiece einops accelerate==${accelerate_version} bitsandbytes==${bitsandbytes_version} \
     optimum=${optimum_version} auto-gptq=${auto_gptq_version} \
```
tosterberg, line 77: Not ours but the optimum=${optimum_version} will fail, need a "==" for the version

tests/integration/llm/prepare.py
```diff
@@ -718,6 +737,16 @@ def build_unmerged_lora_correctness_model(model):
     shutil.copyfile("llm/unmerged_lora.py", "models/test/model.py")
 
 
+def build_ds_smoothquant_model(model):
+    if model not in ds_smoothquant_model_list.keys():
+        raise ValueError(
+            f"{model} is not one of the supporting handler {list(ds_smoothquant_model_list.keys())}"
+        )
+    options = ds_smoothquant_model_list[model]
+    options["engine"] = "DeepSpeed"
```
tosterberg, line 746: I believe you will also want to specify your handler here options["entryPoint"] = "djl_python.deepspeed" since we aren't using a specific model.py

tests/integration/llm/client.py
```diff
@@ -397,6 +397,24 @@ def get_model_name():
...
 
+ds_smoothquant_model_spec = {
+    "gpt-j-6b": {
+        "max_memory_per_gpu": [9.0, 10.0, 11.0, 12.0],
+        "batch_size": [1, 2, 4, 8],
+        "seq_length": [64, 128, 256],
+    },
+    "llama2-13b": {
+        "max_memory_per_gpu": [22.0],
+        "batch_size": [1],
```
tosterberg, line 408: Lets do a test with batch size 2+ as well, we don't need the whole series but we will be able to observe output errors if batch is not handled correctly.

tests/integration/llm/client.py
```diff
@@ -397,6 +397,24 @@ def get_model_name():
...
+        "seq_length": [64, 128, 256],
+    },
+    "gpt-neox-20b": {
+        "max_memory_per_gpu": [22.0],
+        "batch_size": [1],
```
tosterberg, line 413: Same as above, add a batch size 2 or 4 test case to this

tosterberg (changes requested):

tests/integration/llm/client.py
maaquib: Will update. Need to update the `max_memory_per_gpu` with some more accurate numbers

serving/docker/deepspeed.Dockerfile
maaquib: Yes. I already reached out to @KexinFeng for that. He has a PR for the fix

tosterberg (approved):

PR #1141, opened 2023-10-03
[WIP][FasterTransformer] use python 3.10.0 and upgrade pytorch

lanking520:
## Description ##

Upgrade PyTorch and bump up python to 3.10


zachgk:
Doesn't this only change Python for FT but leave it the same (3.9) in the other containers?

lanking520:
defered

PR #1147, opened 2023-10-03
Block remote adapter url and handler override

zachgk:


serving/src/main/java/ai/djl/serving/http/RequestParser.java
```diff
@@ -92,6 +92,11 @@ public Input parseRequest(FullHttpRequest req, QueryStringDecoder decoder) {
             byte[] content = NettyUtils.getBytes(req.content());
             input.add("data", content);
         }
+
+        if (input.getProperties().containsKey("handler")) {
+            throw new BadRequestException("The handler can't be overridden in a request");
```
frankfliu, line 97: Why we want to block handler override?

This at least should be an option (a global conf.properties) to disable handler override if we want disable it.

wlm/src/main/java/ai/djl/serving/wlm/Adapter.java
```diff
@@ -53,6 +55,14 @@ public static Adapter newInstance(WorkerPoolConfig<?, ?> wpc, String name, Strin
             String modelName = wpc.getId();
             throw new IllegalArgumentException("The worker " + modelName + " is not a model");
         }
+
+        // TODO Allow URL support
+        try {
+            new URL(src);
```
rohithkrn, line 61: if we only want to support local path, we could just check if `src` is a local path?

serving/src/main/java/ai/djl/serving/http/RequestParser.java
zachgk: I wrote a comment about this in slack

wlm/src/main/java/ai/djl/serving/wlm/Adapter.java
frankfliu: Better use URI, and check scheme is file or null, it more readable

serving/src/main/java/ai/djl/serving/http/RequestParser.java
zachgk: @frankfliu If we want to support this, I think it should be a property to enable rather than disable. Do you know if we have users using custom handlers now?

frankfliu: This was designed to allow user to call pre/postprocessing:

https://github.com/deepjavalibrary/djl-demo/tree/master/development/python

frankfliu: We can default to disable handler override

zachgk: I think we can leave this discussion for later. For now, I added an env/system prop `ALLOW_REQUEST_HANDLER_OVERRIDE` to re-enable handler overrides

lanking520 (approved):

PR #1156, opened 2023-10-05
Add error message for quantization when using checkpoint loading.

chen3933:
## Description ##

Add error message for quantization when using checkpoint loading.

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -357,6 +357,11 @@ class DeepSpeedService(object):
             self.ds_config["replace_with_kernel_inject"] = False
         state_dict_mmap = {}
         if "checkpoint" in self.ds_config:
+            if self.quantize_mode:
+                raise ValueError(
+                    f"quantize option does NOT currently work WITH DeepSpeed checkpoints using checkpoint option. "
+                    f"Please using quantization with a standard HuggingFace checkpoint or "
+                    f"turn off quantization and try again.")
             model, self.tokenizer, state_dict_mmap = self.load_model(
                 self.model_id_or_path,
                 'from_config',


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -357,6 +357,10 @@ def create_model_pipeline(self):
             self.ds_config["replace_with_kernel_inject"] = False
         state_dict_mmap = {}
         if "checkpoint" in self.ds_config:
+            if self.quantize_mode:
+                raise ValueError(
+                    f"Quantization does NOT currently work WITH checkpoint loading. "
+                    f"Please turnoff quantization and try again.")
```
davidthomas426, line 363: 1. typo: "turnoff" -> "turn off"
2. Let's say "`quantize_mode` option does NOT currently work WITH DeepSpeed checkpoints using `checkpoint` option." to be painfully clear
3. Could first suggest using quantization with a standard HuggingFace checkpoint, before adding "or turn off quantization". Our new model loading with a standard checkpoint on SageMaker is even more memory efficient and almost as fast to load as a presharded DeepSpeed checkpoint.

chen3933: Done

davidthomas426 (approved):
LGTM

tosterberg (approved):

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -357,6 +357,11 @@ def create_model_pipeline(self):
             self.ds_config["replace_with_kernel_inject"] = False
         state_dict_mmap = {}
         if "checkpoint" in self.ds_config:
```
lanking520, line 359: should we just disable checkpoint option entirely, given we already have cpu loading optimization?

davidthomas426: It would be a feature regression. Is anyone using it? We can unify things pretty easily in future version so that checkpoint loading works in unison with all other features, we just didn't get to that yet.

PR #1160, opened 2023-10-06
Adds versions as labels in dockerfiles

zachgk:
This begins adding the versions to dockerfiles so they can later be picked up by the DLC pipeline.

### DIFF
--- a/serving/docker/Dockerfile
+++ b/serving/docker/Dockerfile
@@ -55,6 +55,8 @@ LABEL dlc_major_version="1"
 LABEL com.amazonaws.ml.engines.sagemaker.dlc.framework.djl.cpu="true"
 LABEL com.amazonaws.sagemaker.capabilities.multi-models="true"
 LABEL com.amazonaws.sagemaker.capabilities.accept-bind-to-port="true"
+LABEL djl-version=$djl_version
+
 
 FROM base AS cpu-full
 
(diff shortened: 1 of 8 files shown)


serving/docker/Dockerfile
```diff
@@ -51,6 +51,8 @@ LABEL dlc_major_version="1"
 LABEL com.amazonaws.ml.engines.sagemaker.dlc.framework.djl.cpu="true"
 LABEL com.amazonaws.sagemaker.capabilities.multi-models="true"
 LABEL com.amazonaws.sagemaker.capabilities.accept-bind-to-port="true"
+LABEL djl-version=$djl_version
```
frankfliu, line 54: 1. Our label already contains version info: 

```
com.amazonaws.ml.engines.sagemaker.dlc.framework.djl.v0-24-0.cpu-full="true"
```
2. the version should be part of the label name, the value can be override if user extends from base image
3. the framework version should have 1:1 mapping to djl version.

zachgk: This is mostly for the DLC release. Right now, we pass this information through environment variables manually and this should add a way for it to be passed automatically

zachgk: Having some way of passing this data automatically will be good for our DLC release pipeline. It will avoid another point of failure (especially for patching old releases) and could be used to make the step fully automatic, especially if we also trigger it automatically from our actions

frankfliu (approved):

PR #1172, opened 2023-10-10
Faster in-memory weight transfer for transformers-neuronx

Lokiiiiii:
## Overview ##

This PR improves the latency of transferring model weights from the HF representation to the INF2 representation by doing it in-memory as opposed to using the disk as a buffer.  For OPT-66B on inf2.48xl, this improves the latency of weight transfer from 840s to 5s. 

## Description ## 

Splitting the model into small pieces on the disk and then loading it, reduces CPU memory requirements. However, this operation demands high IOPS from the disk. Typical EBS volumes used by EC2 (gp3) and SageMaker (gp2) do not provide high IOPS by default. This leads to EBS throttling, which manifests as a massive latency to transfer model weights (upto 1400 secs).

However, Large Language Models are typically used with large instances with large CPU memory. As a result, for instances like inf2.24xl, inf2.48xl, trn1.32xl, loading from disk is wasteful in terms of instance hours. Direct in-memory transfer is much faster (<5s) in these cases.

This PR modifies the load_model code path to first try in-memory loading and then fall back to disk loading in the event of an overflow.

## Testing ##
Tested in EC2 on inf2.48xl and inf2.24xl with OPT-66B and OPT-13B

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx.py
+++ b/engines/python/setup/djl_python/transformers_neuronx.py
@@ -69,6 +69,7 @@ class TransformersNeuronXService(object):
         self.revision = None
         self.rolling_batch = None
         self.load_in_8bit = False
+        self.low_cpu_mem_usage = False
 
     def init_load_path(self, model_type):
         path = os.environ.get("SERVING_DOWNLOAD_DIR")
@@ -146,11 +147,11 @@ class TransformersNeuronXService(object):
     def load_model(self, model_type):
         self.model = self.load_hf_model()
         load_path = self.get_load_path(model_type)
-        try:
-            logging.info("Trying to transfer weights in-memory")
+        if not self.low_cpu_mem_usage:
+            logging.info("Transferring weights from HF to INF2 in-memory")
             self.model = self.load_inf2_model_from_memory(model_type)
-        except Exception as exc:
-            logging.info(f"Falling back to disk transfer due to {str(exc)}")
+        else:
+            logging.info("Transferring weights from HF to INF2 through disk")
             self.model = self.load_inf2_model_from_disk(model_type, load_path)
         logging.info(f"LLM sharding and compiling Started ...")
         start = time.time()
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -122,10 +119,39 @@ def load_inf2_model(self, model_type, load_path):
             n_positions=self.n_positions,
             unroll=self.unroll)
 
+    def load_inf2_model_from_memory(self, model_type):
```
lanking520, line 122: what is making a difference with above one? Just an extra step to load with low memory?

lanking520: does this work as the same as we are passing in `low_cpu_mem_usage=True`

Lokiiiiii: It provides an alternative to the model split to disk and load from disk workflow.
It directly transfers the model weights in-memory.

Lokiiiiii: More details provided in the CR description

engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -122,10 +119,39 @@ def load_inf2_model(self, model_type, load_path):
...
+        else:
+            model = MODEL_TYPE_TO_MODEL[model_type](
+                self.model.config,
+                batch_size=self.batch_size,
+                amp=self.amp,
+                tp_degree=self.tensor_parallel_degree,
+                n_positions=self.n_positions,
+                unroll=self.unroll)
+        model.load_state_dict_low_memory(self.model.state_dict())
+        return model.load_state_dict_low_memory(self.model.state_dict())
```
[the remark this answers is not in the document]
Lokiiiiii: Typo ! Fixed in new revision.

Lokiiiiii:
Additionally tested `facebook/opt-1.3b` and `OpenAssistant/llama2-13b-orca-8k-3319` on inf2.24xl (EC2)

lanking520 (approved):

PR #1181, opened 2023-10-12
[bug fix] add entrypoint camel case recovery

lanking520:
## Description ##

Bug fix on entrypoint parsing


zachgk (approved):

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -702,6 +702,9 @@ private void configPerModelSettings() throws ModelException {
             String key = entry.getKey();
             if (key.startsWith("OPTION_")) {
                 key = key.substring(7).toLowerCase(Locale.ROOT);
+                if ("entrypoint".equals(key)) {
```
frankfliu, line 705: We have DJL_ENTRY_POINT env var already

PR #1183, opened 2023-10-12
Add CI performance test for deepspeed smoothquant.

chen3933:
## Description ##

Added benchmark for smoothquant

Tested on ec2 instance.

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -280,22 +280,105 @@ performance_test_list = {
         "option.task": "text-generation",
         "option.model_id": "s3://djl-llm/opt-30b/"
     },
-    "open-llama-13b": {
+    "open-llama-13b-fp16-deepspeed": {
         "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "deepspeed",
         "option.model_id": "s3://djl-llm/open-llama-13b/"
     },
-    "gpt-j-6b": {
+    "open-llama-13b-fp16-huggingface": {
         "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "huggingface",
+        "option.model_id": "s3://djl-llm/open-llama-13b/"
+    },
+    "open-llama-13b-bf16-deepspeed": {
+        "option.task": "text-generation",
+        "option.dtype": "bf16",
+        "option.engine": "deepspeed",
+        "option.model_id": "s3://djl-llm/open-llama-13b/"
+    },
+    "open-llama-13b-bf16-huggingface": {
+        "option.task": "text-generation",
+        "option.dtype": "bf16",
+        "option.engine": "huggingface",
+        "option.model_id": "s3://djl-llm/open-llama-13b/"
+    },
+    "open-llama-13b-smoothquant": {
+        "option.task": "text-generation",
+        "option.model_id": "s3://djl-llm/open-llama-13b/",
+        "option.dtype": "fp16",
+        "option.engine": "deepspeed",
+        "option.quantize": "smoothquant"
+    },
+    "gpt-j-6b-fp16-deepspeed": {
+        "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "deepspeed",
         "option.model_id": "s3://djl-llm/gpt-j-6b/"
+    },
+    "gpt-j-6b-fp16-huggingface": {
+        "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "huggingface",
+        "option.model_id": "s3://djl-llm/gpt-j-6b/"
+    },
+    "gpt-j-6b-bf16-deepspeed": {
+        "option.task": "text-generation",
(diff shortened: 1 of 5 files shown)


tosterberg (approved):

tests/integration/llm/prepare.py
```diff
@@ -598,9 +681,17 @@ def build_performance_model(model):
...
+    if options.get('option.dtype') is None:
+        raise ValueError("Need to provide dtype for performance benchmark")
     options["option.tensor_parallel_degree"] = args.tensor_parallel
-    for k, v in default_accel_configs[args.engine].items():
+    engine = options.get('option.engine')
+    if args.engine:
+        engine = args.engine
+    if engine is None:
+        raise ValueError("Need to provide engine for performance benchmark")
+    for k, v in default_accel_configs[engine].items():
```
maaquib, line 694: I think this won't be required with the above change

tests/integration/llm/prepare.py
```diff
@@ -280,22 +280,105 @@
         "option.task": "text-generation",
         "option.model_id": "s3://djl-llm/opt-30b/"
     },
-    "open-llama-13b": {
+    "open-llama-13b-fp16-deepspeed": {
         "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "deepspeed",
```
maaquib, line 286:
```suggestion
        "engine": "DeepSpeed",
```

tests/integration/llm/prepare.py
```diff
@@ -280,22 +280,105 @@
...
         "option.model_id": "s3://djl-llm/open-llama-13b/"
     },
-    "gpt-j-6b": {
+    "open-llama-13b-fp16-huggingface": {
+        "option.task": "text-generation",
+        "option.dtype": "fp16",
+        "option.engine": "huggingface",
```
maaquib, line 292:
```suggestion
        "engine": "Python",
        "option.entryPoint": "djl_python.huggingface"
```

maaquib (changes requested):

maaquib (approved):

tests/integration/llm/prepare.py
```diff
@@ -280,22 +280,105 @@
...
+    },
+    "open-llama-13b-bf16-huggingface": {
+        "option.task": "text-generation",
+        "option.dtype": "bf16",
+        "option.engine": "huggingface",
+        "option.model_id": "s3://djl-llm/open-llama-13b/"
+    },
+    "open-llama-13b-smoothquant": {
+        "option.task": "text-generation",
+        "option.model_id": "TheBloke/Llama-2-13B-fp16",
```
chen3933, line 309: Is this the correct model I should use for smoothquant benchmark?

tests/integration/profiles/gpt_neox_20b.json
```diff
@@ -1,8 +1,6 @@
 {
 	"test_series":"performance",
-	"engine":["deepspeed","huggingface"],
-	"model": "gpt-neox-20b",
-	"dtype": ["fp16","bf16"],
+	"model": ["gpt-neox-20b-fp16-deepspeed", "gpt-neox-20b-fp16-huggingface", "gpt-neox-20b-bf16-deepspeed", "gpt-neox-20b-bf16-huggingface", "gpt-neox-20b-smoothquant"],
```
chen3933, line 3: One callout : This will change metric_name. Is this going to affect downstream service that processes this metrics?

tosterberg: It shouldn't be an issue, but is there a way that we can do this without affecting the old metrics? IE. make a new entry with engine:"deepspeed", model:gpt-neox-20b-smoothquant-int8 with the associated changes to the serving.properties in the prepare.py, and dtype: fp16 can stay. It will generate a new metric just for the SQ version of this model.

tests/integration/llm/prepare.py
tosterberg: We will want to move it over to an s3 bucket to minimize container startup, but I think we want apples to apples comparisons, so this should be the s3 llama model used by the normal deepspeed performance test.

tests/integration/profiles/gpt_neox_20b.json
chen3933: The metric_name is associate with `model` (https://github.com/deepjavalibrary/djl-serving/blob/master/tests/integration/profiles/gpt_neox_20b.json#L4). Also dtype in json will get expended and applied to all models. But we don't want to run smoothquant only on bf16. I cannot find a easy change to apply to our use case which preserve old metrics_name.

tests/integration/llm/prepare.py
chen3933: Got it. Let me update it.

tests/integration/profiles/gpt_neox_20b.json
tosterberg: The above would have the downside of needing another profile, and stage in the yaml, so it is a tradeoff that we would have to consider.

PR #1197, opened 2023-10-18
Adds documentation for adapters

zachgk:
## Description ##

### DIFF
--- a/serving/docs/adapters.md
+++ b/serving/docs/adapters.md
@@ -98,6 +98,8 @@ There are a few things to keep in mind which choosing a calling technique.
 
 This passes the adapter as part of the requeset body.
 It will also work with client-side batching and will allow multiple adapters to be passed with one for each input.
+If the adapters is not passed, the base model will be used for inference.
+You can also specify the base model for an element in the batch by using the empty string `""`.
 
 ```
 
@@ -150,9 +152,9 @@ There are two kinds of LoRA that can be put onto various engines.
 With our default handlers, we currently support unmerged LoRA for CausalLM through the huggingface handler.
 Support for other model types and handlers is coming soon.
 
-### Writing Adapter Models
+### Writing Custom Adapter Models
 
-Right now, adapters are only supported through our Python engine.
+Right now, adapters are only supported through our Python engine and not any of the other DJL engines.
 See instructions to get started with writing a [python handler](modes.md#python-mode).
 
 To add support for adapters, you must first add the register and unregister like below.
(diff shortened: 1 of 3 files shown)


serving/docs/adapters.md
```diff
@@ -0,0 +1,180 @@
...
+
+DJL Serving has first class support for adapters.
+Adapters are patches or changes that can be made to a model to fine tune it for a particular usage.
+The benefit of adapters rather than whole model fine-tuning is that they are often smaller and easier to distribute alongside a base model.
+This can allow for multiple adapters used at the same system, and sometimes even in the same batch.
+
+With DJL Serving, it is possible to easily work with adapters.
+You can create models that accept adapters, use DJL Serving to manage your available adapters, and call models with adapters.
+
+For a concrete usage, see the [large model inference adapters example notebook](http://docs.djl.ai/docs/demos/aws/sagemaker/large-model-inference/sample-llm/multi_lora_adapter_inference.html).
```
rohithkrn, line 13: this link does not seem to work

zachgk: Yeah. It goes to the notebook from https://github.com/deepjavalibrary/djl-demo/pull/372. But, that was just merged and not built into our docs yet. When it is built in, it should work

serving/docs/adapters.md
```diff
@@ -0,0 +1,180 @@
...
+  It has a higher inference latency for the additional adapter operations.
+  However, it does support multi-adapter batches.
+  It works best for use cases with large numbers of adapters.
+
+With our default handlers, we currently support unmerged LoRA for CausalLM through the huggingface handler.
+Support for other model types and handlers is coming soon.
+
+### Writing Adapter Models
+
+Right now, adapters are only supported through our Python engine.
```
rohithkrn, line 155: we can narrow it down to HF Accelerate. Python engine also includes other backends like Triton core.

serving/docs/adapters.md
```diff
@@ -0,0 +1,180 @@
...
+  name = inputs.get_properties()["name"]
+  # Do adapter unregistration tasks
+  return Output().add("Successfully unregistered adapter")
+
+
+def handle(inputs: Input):
+  ...
+```
+
+Within the handler, you must parse the adapter properties from the inputs.
```
rohithkrn, line 178: we should also add that `adapters` arg should be passed to `generate` function as a `kwarg`. For e.g
```
generate(input_ids=input_tokens.input_ids,
                attention_mask=input_tokens.attention_mask,
                adapters=[adapter_1, adapter_2],
               **kwargs
)

```

serving/docs/adapters.md
zachgk: This is referring to the non-Python engine pieces such as the DJL OnnxRuntime engine or DJL MXNet engine. With the python engine, they could support it in other engines if they were willing to write their own handler instead of using the default ones

serving/docs/adapters.md
zachgk: Can you give more context about the generate function? Is this PySDK or something?

rohithkrn (approved):

PR #1198, opened 2023-10-18
[feat] fail continuous test if python is not formatted in CI

tosterberg:
## Description ##

PR looks to add a test in the continuous workflow, so that when the build test is done on the ubuntu image we will also test to see if running ./gradlew formatPython changes any files. If it does then we fail with an error informing the PR author to run the command and recommit to pass the pipeline.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,12 +121,3 @@ configure(javaProjects()) {
         }
     }
 }
-
-apply from: file("${rootProject.projectDir}/tools/gradle/publish.gradle")
-tasks.register('formatPython') {
-    doFirst {
-        exec {
-            commandLine "bash", "-c", "find . -name '*.py' -not -path '*/.gradle/*' -not -path '*/build/*' -print0 | xargs -0 yapf --in-place"
-        }
-    }
-}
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -122,11 +122,32 @@ configure(javaProjects()) {
...
 tasks.register('formatPython') {
     doFirst {
-        exec {
-            commandLine "bash", "-c", "find . -name '*.py' -not -path '*/.gradle/*' -not -path '*/build/*' -print0 | xargs -0 yapf --in-place"
-        }
+        runPythonFormatter()
     }
 }
+
+tasks.register('testFormat') {
```
frankfliu, line 138: To be consistent with `verifyJava`, let rename this to `verifyPython`

build.gradle
```diff
@@ -122,11 +122,32 @@ configure(javaProjects()) {
     }
 }
 
+def runPythonFormatter = { ->
```
frankfliu, line 125: can we move them into tools/gradle/python-formatter.gradle

frankfliu (approved):

PR #1199, opened 2023-10-19
Add TRTLLM TRT build from our managed source

lanking520:
## Description ##

resolving nccl version mismatch problems

### DIFF
--- a/serving/docker/scripts/install_tensorrt.sh
+++ b/serving/docker/scripts/install_tensorrt.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+set -ex
+
+install_ubuntu_requirements() {
+    CUDNN_VERSION="8"
+    apt-get update
+    apt-get install -y --no-install-recommends libcudnn${CUDNN_VERSION} libcudnn${CUDNN_VERSION}-dev libnccl-dev=2.18.5-1+cuda12.2
+    apt-get clean
+    rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


serving/docker/scripts/install_trt_llm.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
rohithkrn, line 1: is the plan to use this script in the dockerfile? 
I would suggest we do similar to FT container, where we pull the binaries directly in the dockerfile instead of building each time since it only changes per release and we don't need nightly builds.

lanking520: good points, let's move this to a different place

serving/docker/scripts/install_tensorrt.sh
```diff
@@ -0,0 +1,47 @@
...
+install_ubuntu_requirements() {
+    CUDNN_VERSION="8"
+    apt-get update
+    apt-get install -y --no-install-recommends libcudnn${CUDNN_VERSION} libcudnn${CUDNN_VERSION}-dev libnccl-dev=2.18.5-1+cuda12.2
+    apt-get clean
+    rm -rf /var/lib/apt/lists/*
+}
+
+install_tensorrt() {
+    TENSOR_RT_VERSION="9.1.0.4"
```
rohithkrn, line 14: we can make these as args

PR #1218, opened 2023-10-25
[TRTLLM][SAMPLE] add trtllm rough rolling batcher

lanking520:
## Description ##

Add basic rollingbatcher for trtLLM

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -114,6 +114,9 @@ def get_rolling_batch_class_from_str(rolling_batch_type: str, is_mpi: bool,
         logging.warning(
             "vLLM rolling batcher is experimental, use with caution")
         return VLLMRollingBatch
+    elif rolling_batch_type == "trtllm":
+        from djl_python.rolling_batch.trtllm_rolling_batch import TRTLLMRollingBatch
+        return TRTLLMRollingBatch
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -114,6 +114,9 @@ def get_rolling_batch_class_from_str(rolling_batch_type: str, is_mpi: bool,
         logging.warning(
             "vLLM rolling batcher is experimental, use with caution")
         return VLLMRollingBatch
+    elif rolling_batch_type == "trtllm":
```
rohithkrn, line 117: is this temporary? I don't think we want to use huggingface handler for trt-llm

engines/python/setup/djl_python/trtllm.py
```diff
@@ -0,0 +1,90 @@
...
+    input_map = decode(inputs, content_type)
+    _inputs = input_map.pop("inputs", input_map)
+    parameters = input_map.pop("parameters", {})
+    if not isinstance(_inputs, list):
+        _inputs = [_inputs]
+    batch_size = len(_inputs)
+    reqs = {}
+    for i in range(batch_size):
+        payload = get_sample_payload(_inputs[i], parameters, tokenizer)
+        reqs[i] = model.inference_async(payload)
```
rohithkrn, line 61: ideally for dynamic batching, we don't have to use the inflight batching backend. We should batch the inputs here and run `inference` method instead.

PR #1222, opened 2023-10-26
Cleans tensorParallelDegree with MultiDevice

zachgk:
This is a refactor to simplify the handling of tensor parallel degree. Before, it is read independently in 3+ locations in code and the behavior determining the tpDegree is hard to follow. This moves the reading to a single place and then represents it using a MultiDevice. This also means that the behavior can be entirely visible - a worker group will show up with the tp devices that it's worker will be using.

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/Connection.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/Connection.java
@@ -13,6 +13,7 @@
 package ai.djl.python.engine;
 
 import ai.djl.Device;
+import ai.djl.Device.MultiDevice;
 import ai.djl.Model;
 import ai.djl.engine.EngineException;
 import ai.djl.inference.streaming.ChunkedBytesSupplier;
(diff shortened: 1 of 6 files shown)


sindhuvahinis (approved):

sindhuvahinis:
Now that we have MultiDevice, Here getVisibleDevices could make of use of this as well right? https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/src/main/java/ai/djl/python/engine/Connection.java#L207

zachgk:
> Now that we have MultiDevice, Here getVisibleDevices could make of use of this as well right? https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/src/main/java/ai/djl/python/engine/Connection.java#L207

Yeah. The engine would already have the list of devices that should be part of the connection already. It doesn't use the `CUDA_VISIBLE_DEVICES` though.

Also, I talked with @frankfliu and we may save this PR for post-reinvent as it is hard to get the behavior right and we don't test every case even in integration. Although, I did add a bunch of tests which may be sufficient

PR #1226, opened 2023-10-27
[fix] use fast sample as default sample method for inf2

tosterberg:
## Description ##

Updating the default sample strategy to use the neuron models implementation for improved performance. Additionally, adding the alias "fast_sample" to the "use_sample" parameter as use_sample is unclear as it really means "use neuronx model's sample implementation".

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx.py
+++ b/engines/python/setup/djl_python/transformers_neuronx.py
@@ -49,6 +49,29 @@ MODEL_TYPE_TO_MODEL = {
 }
 
 
+class NeuronXSampleAdapter(HuggingFaceGenerationModelAdapter):
+
+    def __init__(self, _config, _model):
+        super().__init__(_config, _model)
+        self.model_type = _config.model_type
+        self.sample_options = ["start_ids", "top_k"]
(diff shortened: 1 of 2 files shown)


sindhuvahinis (approved):

engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -334,15 +356,13 @@ def inference(self, inputs):
...
-        if use_sample:
-            # TODO: Watch transformer-neuronx release for fix on gpt-neox generate functionality
-            output_tokens = self.model.sample(
-                encoded_inputs.input_ids,
-                sequence_length=self.n_positions,
-                pad_token_id=self.tokenizer.pad_token_id,
-                eos_token_id=self.tokenizer.eos_token_id,
-                **parameters)
+        fast_sample = (parameters.pop("use_sample", True)
+                       and parameters.pop("fast_sample", True))
```
lanking520, line 360: why we introduce a new param here?

tosterberg: We don't need to - I am removing it.

PR #1227, opened 2023-10-27
[neuronx] bump to 2.15 for tnx container and scripts

tosterberg:
## Description ##

Update the tnx container to 2.15 sdk


lanking520:
Missing another half of compiler and runtime updates. (In script)

tosterberg:
Added the script - and saw the clarification on torch 2 vs 1.13. I am leaving it as 1.13 per that discussion.

lanking520 (approved):

PR #1235, opened 2023-10-30
Add trt-llm engine build step during model initialization

rohithkrn:
## Description ##

Add TRT-LLM engine build step during model initialization

TODO: Tests

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -163,6 +163,8 @@ public class PyModel extends BaseModel {
                 } else if ("nc".equals(manager.getDevice().getDeviceType())
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
+                } else if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND"))) {
+                    entryPoint = "djl_python.trtllm";
                 } else if (pyEnv.getInitParameters().containsKey("model_id")) {
                     entryPoint = "djl_python.huggingface";
                 } else {
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -164,7 +164,7 @@ public class PyModel extends BaseModel {
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
                 } else if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND"))) {
-                    entryPoint = "djl_python.trtllm";
+                    entryPoint = "djl_python.tensorrt_llm";
                 } else if (pyEnv.getInitParameters().containsKey("model_id")) {
                     entryPoint = "djl_python.huggingface";
                 } else {
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -15,6 +15,7 @@ package ai.djl.python.engine;
 import ai.djl.BaseModel;
 import ai.djl.Device;
 import ai.djl.Model;
+import ai.djl.ModelException;
 import ai.djl.engine.EngineException;
 import ai.djl.inference.Predictor;
 import ai.djl.ndarray.NDManager;
@@ -150,6 +151,21 @@ public class PyModel extends BaseModel {
             pyEnv.setFailOnInitialize(false);
         }
 
+        // Handle TRT-LLM
+        if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND")) || Boolean.parseBoolean(getProperty("option.trt_llm"))) {
+            try {
+                Path trtLlmRepoDir = TrtLLMUtils.initTrtLlmModel(this);
+                if (trtLlmRepoDir != null) {
+                    String modelId = trtLlmRepoDir.toAbsolutePath().toString();
+                    setProperty("model_id", modelId);
+                    pyEnv.addParameter("model_id", modelId);
+                    entryPoint = "djl_python.tensorrt_llm";
+                }
+            } catch (ModelException e) {
+                throw new RuntimeException(e);
+            }
+        }
+
         if (entryPoint == null) {
             entryPoint = Utils.getenv("DJL_ENTRY_POINT");
             if (entryPoint == null) {
@@ -163,8 +179,6 @@ public class PyModel extends BaseModel {
                 } else if ("nc".equals(manager.getDevice().getDeviceType())
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
-                } else if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND"))) {
-                    entryPoint = "djl_python.tensorrt_llm";
                 } else if (pyEnv.getInitParameters().containsKey("model_id")) {
                     entryPoint = "djl_python.huggingface";
                 } else {
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -39,6 +39,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -152,14 +153,14 @@ public class PyModel extends BaseModel {
         }
 
         // Handle TRT-LLM
-        if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND")) || Boolean.parseBoolean(getProperty("option.trt_llm"))) {
+        boolean isTrtLlmBackend = "TRT-LLM".equals(Utils.getenv("LMI_BACKEND"));
+        if (isTrtLlmBackend) {
             try {
-                Path trtLlmRepoDir = TrtLLMUtils.initTrtLlmModel(this);
-                if (trtLlmRepoDir != null) {
-                    String modelId = trtLlmRepoDir.toAbsolutePath().toString();
+                Optional<Path> trtLlmRepoDir = TrtLLMUtils.initTrtLlmModel(this);
+                if (trtLlmRepoDir.isPresent()) {
+                    String modelId = trtLlmRepoDir.get().toAbsolutePath().toString();
                     setProperty("model_id", modelId);
                     pyEnv.addParameter("model_id", modelId);
-                    entryPoint = "djl_python.tensorrt_llm";
                 }
             } catch (ModelException e) {
                 throw new RuntimeException(e);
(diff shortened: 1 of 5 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -964,6 +968,51 @@ void downloadS3() throws ModelException, IOException {
         }
     }
 
+    void build_trt_llm_engine() throws ModelException, IOException {
+        logger.info("Start building TensorRT-LLM engine");
```
lanking520, line 972: Start converting model to TensorRT LLM format

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -964,6 +968,51 @@ void downloadS3() throws ModelException, IOException {
...
+        final String trt_llm_model_repo = "/tmp/tensorrtllm";
+        prop.putIfAbsent("option.trt_llm_model_repo", trt_llm_model_repo);
```
lanking520, line 974: If we decided this location as a fixed location, you don't need to overwrite it. Just hardcode to cover in the toolkit code base

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -964,6 +968,51 @@ void downloadS3() throws ModelException, IOException {
...
+        FileOutputStream fileOutputStream = new FileOutputStream(propertiesFile.toFile());
+        prop.store(fileOutputStream, "write serving.properties");
+        fileOutputStream.close();
+        
+        // invoke trt-llm build script
+        List<String> commandList = new ArrayList<>();
+        commandList.add("python");
+        commandList.add("/opt/djl/partition/trt_llm_partition.py");
+        commandList.add("--properties_dir");
+        commandList.add(propertiesDir.toString());
```
lanking520, line 992: Sometimes people pass in all properties through environment variables. So there is no physical  file

serving/docker/partition/trt_llm_partition.py
```diff
@@ -0,0 +1,55 @@
...
+    create_model_repo(model_id_or_path, **kwargs)
+
+def main():
+    logging.basicConfig(stream=sys.stdout,
+                        format="%(message)s",
+                        level=logging.INFO)
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--properties_dir',
```
lanking520, line 38: Be precise, let's define a way to pass the key-value pair instead of a file. The script is aimed to handle:

1. HF repo or ID and do download + conversion
2. A converted Triton Repo and skip conversion

rohithkrn: This script currently reads serving.properties file from the given path, hence named it `properties_dir`. Since we are passing data from one process to other, I felt passing file path is neat. To pass key-value pairs, we have to pass varargs and felt not so clean.
1. Covers this case now in the current state
2. This is not covered. I was thinking that if Triton repo is set, we should not overload it in same `option.model_id` property and use a different property like `option.trt_llm_model_repo` since they represent two different entities. We also need to send `model_id` anyway to define tokenizer.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: The idea here is to not overwrite the property but rather setting this property and guide the toolkit to create model repo in this path. The same property is again passed to the toolkit in `init_inference` call. The same can be achieved by hardcoding the path in the toolkit except that this makes it easier to change in case we want to make it configurable in the future by just making the change here.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: Right, this is not relying on user set `serving.properties`. I am writing a new file in #L984 and by that time `prop` is [updated](https://github.com/deepjavalibrary/djl-serving/blob/master/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java#L703-L710) with environment variables already, so it will env variables information as well.

serving/docker/partition/trt_llm_partition.py
lanking520: Let's don't introduce two different concept to users. User just need to provide a model_id, that can be a Triton repo, can be a HF model. Our script will determine how to deal with it. In the container, we can set an environment variable called TRITON_REPO and point to /tmp/triton_repo. We always pass converted model there.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
lanking520: You can save the new information to PyEnv instead of writing to a file

serving/docker/partition/trt_llm_partition.py
lanking520: In my POC, the way we define a Triton ready model is

1. It has config.json and tokenizer components for HF basic components
2. The path/tensorrt_llm/ has config.pbtxt

Use those two criteria to find out if a repo is Triton compatible.

If not compatible, convert the model and save to default location like /tmp/tritonrepo for model loading. If compatible, pass conversion and allow inference script use it

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: using existing serving.properties file and reading env vars on the python side.

serving/docker/partition/trt_llm_partition.py
rohithkrn: Updated

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -165,6 +165,8 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
                 } else if ("nc".equals(manager.getDevice().getDeviceType())
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
+                } else if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND"))) {
```
lanking520, line 168: @sindhuvahinis

serving/docker/partition/trt_llm_partition.py
```diff
@@ -0,0 +1,76 @@
...
+    parser.add_argument(
+        '--model_path',
+        type=str,
+        required=False,
+        default=None,
+        help='local path to downloaded model'
+    )
+
+    args = parser.parse_args()
+    properties = load_properties(args.properties_dir)
```
lanking520, line 72: Just check here: do we take care condition that there are no serving.properties?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -1000,6 +1012,92 @@ private void downloadS3(String src, String dest) throws ModelException {
...
+        }
+
+        List<Path> configFiles = new ArrayList<>();
+        List<Path> engineFiles = new ArrayList<>();
+        List<Path> tokenizerFiles = new ArrayList<>();
+        try (Stream<Path> walk = Files.walk(dirToCheck)) {
+            walk.filter(Files::isRegularFile).forEach(
+            path -> {
+                if (path.getFileName().toString().endsWith(".engine")) {
+                    engineFiles.add(path);
```
lanking520, line 1079: I think we can just check pbtxt files. No need to walk with engine files in case some customers have special structured folders

lanking520 (commented):
Your logic doesn't cover a condition that the model is SageMaker uncompressed model: model is saved to /opt/ml/models (read only). The model can be a triton repo or HF standard model. This case you should scan the model_dir and find out if there are model files there.

serving/docker/partition/trt_llm_partition.py
rohithkrn: good point, I was handling this when I was creating a new file. Will update this.

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: this doesn't check for folder structures, it checks recursively in sub directories and looks for files ending with `.engine`. I am okay to just check `config.pbtxt` as it's hard to enforce full repo structure and customers should take some of that responsibility.

rohithkrn:
> Your logic doesn't cover a condition that the model is SageMaker uncompressed model: model is saved to /opt/ml/models (read only). The model can be a triton repo or HF standard model. This case you should scan the model_dir and find out if there are model files there.

This change relies on  model_id set in `serving.properties` or env var. For an uncompressed model, what would model_id look like?

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: Updated to check config.pbtxt and it's presence at depth=2

lanking520 (approved):

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -463,6 +471,10 @@ public void initialize() throws IOException, ModelException {
         downloadModel();
         loadServingProperties();
         downloadS3();
+        isTrtLlmBackend = "TRT-LLM".equals(Utils.getenv("LMI_BACKEND"));
+        if (isTrtLlmBackend) {
+            initTrtLlmModel();
```
zachgk, line 476: Please move these changes to the python engine. Probably the `PyModel.load()` function. The ModelInfo is a higher abstraction that shouldn't be modified by these changes. It should also be easier to test because you can just test it by loading and predicting with the Python Engine using the standard DJL predictor API

zachgk (changes requested):

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
rohithkrn: Agree to your point. Will refactor this in a separate PR as I don't have the bandwidth to refactor and test it currently.

rohithkrn:
CI seems to be flaky. 
Tests pass locally on mac with PR branch.

lanking520:
1 PMD rule violations were found. See the report at: file:///D:/a/djl-serving/djl-serving/wlm/build/reports/pmd/main.html

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
rohithkrn: that doesn't cover the case when customers do not want to use rolling batch. Also, I think there's an effort to make rolling_batch as boolean property

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -150,6 +151,21 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
             pyEnv.setFailOnInitialize(false);
         }
 
+        // Handle TRT-LLM
+        if ("TRT-LLM".equals(Utils.getenv("LMI_BACKEND")) || Boolean.parseBoolean(getProperty("option.trt_llm"))) {
```
rohithkrn, line 155: env var is always set in the container and we don't want to introduce other option. I will update this,

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -150,6 +151,21 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
...
+            try {
+                Path trtLlmRepoDir = TrtLLMUtils.initTrtLlmModel(this);
+                if (trtLlmRepoDir != null) {
+                    String modelId = trtLlmRepoDir.toAbsolutePath().toString();
+                    setProperty("model_id", modelId);
+                    pyEnv.addParameter("model_id", modelId);
+                    entryPoint = "djl_python.tensorrt_llm";
```
rohithkrn, line 162: does this block setting custom entrypoint set by the user?

zachgk: Yes. If that isn't desired, you can move it back or set only if entryPoint is null

zachgk (approved):

PR #1243, opened 2023-10-31
remove tensorrt installation

lanking520:
## Description ##

Don't install tensorrt


rohithkrn:
are we making this change due to license concerns?

serving/docker/scripts/install_tensorrt.sh
```diff
@@ -1,47 +0,0 @@
-#!/bin/bash
```
rohithkrn, line 1: i am leveraging this script in the internal build, should we remove that too?

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -9,12 +9,15 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-ARG version=12.2.2-runtime-ubuntu22.04
+ARG version=12.2.2-cudnn8-runtime-ubuntu22.04
```
rohithkrn, line 12: does it have tensorrt pre-installed?

serving/docker/scripts/install_tensorrt.sh
lanking520: no, internal build you still need trt

serving/docker/tensorrt-llm.Dockerfile
lanking520: no

serving/docker/scripts/install_tensorrt.sh
rohithkrn: i meant if we are okay to use it from legal point of view

serving/docker/tensorrt-llm.Dockerfile
rohithkrn: curious why are we switching from runtime image?

serving/docker/scripts/install_tensorrt.sh
lanking520: to build it is ok, but we have some issues with libnvinfer.so

serving/docker/tensorrt-llm.Dockerfile
lanking520: this is a clean install and save 2 GB of memory. Now the entire container takes 11GB uncompressed

rohithkrn: cudnn image takes less space than runtime image?

rohithkrn (approved):

serving/docker/tensorrt-llm.Dockerfile
lanking520: takes more, but we anyway install everything manually, so it is the same

lanking520: during installation of tensorrt, we install cudnn8

PR #1248, opened 2023-11-01
[TRTLLM] some clean up on trtllm handler

lanking520:
## Description ##

Clean up on the handler

### DIFF
--- a/engines/python/setup/djl_python/tensorrt_llm.py
+++ b/engines/python/setup/djl_python/tensorrt_llm.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 #
-# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -58,56 +33,24 @@ def __init__(self):
...
         if "waiting_steps" in properties:
             kwargs["waiting_steps"] = int(properties.get("waiting_steps"))
-        self.rolling_batch_type = self.rolling_batch_type.lower()
-        is_mpi = properties.get("engine") != "Python"
-        if is_mpi:
-            self.device = int(os.getenv("LOCAL_RANK", 0))
-        _rolling_batch_cls = get_rolling_batch_class_from_str(
-            self.rolling_batch_type, is_mpi, self.model_config)
-        self.rolling_batch = _rolling_batch_cls(model_id_or_path, self.device,
+        if "trt_llm_model_name" in properties:
```
rohithkrn, line 48: instead of converting `properties` to `kwargs` in the handler, it's better if we pass the properties to the toolkit and handler everything there. This conversion misses `trt_*` properties that users can set

lanking520: addressed

PR #1252, opened 2023-11-02
Inf2 properties refactoring using pydantic

sindhuvahinis:
## Description ##

Using Pydantic library, to assign the configurations for the handlers. So installed in inf2 containers. And also added unit test cases, so added for python tests requirement.

Todo:
Define a model using pydantic for DJLServing configurations, which makes easy to deprecate configurations. Will do in the next PR.

Test:
Build inf2 container and ran inf2 integration test cases.  https://github.com/deepjavalibrary/djl-serving/actions/runs/6740207548
Added unit test case for pydantic

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/tnx_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/tnx_properties.py
@@ -100,9 +100,3 @@ class TransformerNeuronXProperties(Properties):
         if self.quantize and self.quantize.value == TnXQuantizeMethods.bitsandbytes8.value:
             self.load_in_8bit = True
         return self
-
-    @model_validator(mode='after')
-    def set_load_split_model(self, validation_info):
-        if self.load_split_model:
-            self.low_cpu_mem_usage = True
-        return self
--- a/engines/python/setup/djl_python/properties_manager/tnx_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/tnx_properties.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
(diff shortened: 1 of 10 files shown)


tosterberg (approved):

engines/python/setup/djl_python/properties_manager/tnx_properties.py
```diff
@@ -0,0 +1,108 @@
...
+    @model_validator(mode='after')
+    def set_quantize(self, validation_info):
+        if self.quantize and self.quantize.value == TnXQuantizeMethods.bitsandbytes8.value:
+            self.load_in_8bit = True
+        return self
+
+    @model_validator(mode='after')
+    def set_load_split_model(self, validation_info):
+        if self.load_split_model:
+            self.low_cpu_mem_usage = True
```
sindhuvahinis, line 107: @lanking520 Is this okay?

lanking520: Let's have @Lokiiiiii comments here. I think Split model loading doesn't need low_cpu_mem_usage

lanking520 (approved):

engines/python/setup/djl_python/properties_manager/tnx_properties.py
Lokiiiiii: `low_cpu_mem_usage` and `load_split_model` are not connected.

`low_cpu_mem_usage` is a known HuggingFace flag to save CPU memory by pipelining to Accelerator memory.
where as `load_split_model` was introduced to save CPU memory at the cost of DiskIO. 

I don't believe these orthogonal features should be bundled together.

sindhuvahinis: Thank you for commenting. Will make the changes.

PR #1273, opened 2023-11-06
[feat] Expose max sparse params

KexinFeng:
## Description ##


zachgk (commented):
Can you also add documentation for this? This includes it's own page (if necessary for an explanation with context) and to the all configurations page

zachgk (approved):
Yeah, that's fine

PR #1274, opened 2023-11-06
Add tokenizer check for triton repo

rohithkrn:
## Description ##

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
@@ -267,9 +267,9 @@ public final class LmiUtils {
                     .forEach(
                             p -> {
                                 Path confFile = p.resolve("config.pbtxt");
-                                // Path tokenizer = p.resolve("1/tokenizer_config.json");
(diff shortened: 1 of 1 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -267,9 +267,8 @@ static boolean isValidTrtLlmModelRepo(Path modelPath) throws IOException {
                     .forEach(
                             p -> {
                                 Path confFile = p.resolve("config.pbtxt");
-                                // Path tokenizer = p.resolve("1/tokenizer_config.json");
-                                // && Files.isRegularFile(tokenizer)
-                                if (Files.isRegularFile(confFile)) {
+                                Path tokenizer = p.resolve("tokenizer_config.json");
```
frankfliu, line 270: Where do we put the tokenizer file? in 1 folder or the same location as config.pbtxt?

rohithkrn: same location as config.pbtxt

frankfliu (approved):

PR #1282, opened 2023-11-07
trt-llm integration tests

rohithkrn:
## Description ##

Added three integration tests
1. llama-13b HF model with tp=4
2. Falcon-7b triton repo with tp = 1
3. llama-7b HF model with tp=4 and smoothquant enabled.

Notes:
1. Need to investigate correctness issue for falcon model
2. Smoothquant conversion is taking significant GPU memory. For e.g llama-7b conversion fails on g5 with tp=1

Test log: https://github.com/deepjavalibrary/djl-serving/actions/runs/6780148882/job/18428397430


.github/workflows/llm_integration.yml
```diff
@@ -687,10 +687,117 @@ jobs:
...
+          serve
+          python3 llm/client.py trtllm falcon-7b
+          rm -rf docker_env
+          docker rm -f $(docker ps -aq)
+      - name: On fail step
+        if: ${{ failure() }}
+        working-directory: tests/integration
+        run: |
+          docker rm -f $(docker ps -aq) || true
+          cat logs/serving.log
```
sindhuvahinis, line 742: we could remove docker_env and models on fail step as well.

sindhuvahinis (approved):

tests/integration/launch_container.sh
```diff
@@ -107,6 +107,10 @@ if $is_llm; then
   if [[ "$platform" == *"inf2"* ]]; then
     total=80
   fi
+  if [[ "$platform" == *"trtllm-sq"* ]]; then
+    echo "extra sleep of 15 min for smoothquant calibration"
```
lanking520, line 111: could we save quantized artifact instead of doing this at runtime?

rohithkrn: I set this number conservatively as a threshold. For the test I added, it doesn't take that much time. From test logs attached, entire process - smoothquant calibration, partitioning., loading took less than 15min. 
This is the only test we have covering quantization path.

PR #1294, opened 2023-11-08
[CI] TnX load split model llama2 test and fix handler

sindhuvahinis:
## Description ##


lanking520 (approved):

sindhuvahinis:
While testing manually, load split model throws error with llama2 artifacts. So will not be merging this changes until tnx 2.16.

sindhuvahinis:
Testing without compiled graph passed. It could be because the compiled graph artifacts were faulty.

PR #1295, opened 2023-11-08
[DS] Deepspeed rolling batch support

maaquib:
## Description ##

### DIFF
--- a/.github/workflows/rolling_batch_integration.yml
+++ b/.github/workflows/rolling_batch_integration.yml
@@ -402,10 +402,83 @@ jobs:
           name: vllm-logs
           path: tests/integration/logs/
 
+  deepspeed-test:
+    if: contains(fromJson('["", "deepspeed"]'), github.event.inputs.run_test)
+    runs-on: [ self-hosted, g5 ]
+    timeout-minutes: 60
+    needs: create-runners
+    steps:
+      - uses: actions/checkout@v3
+      - name: Clean env
+        run: |
+          yes | docker system prune -a --volumes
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -203,9 +205,24 @@ def preprocess_requests(self, requests):
 
     def postprocess_results(self):
         results = []
-        for i in range(len(self.active_requests)):
-            req = self.active_requests[i]
-            res = {"data": req.get_next_token(), "last": req.is_last_token()}
+        err_reqs = dict((r.id, err) for r, err in self.error_requests)
```
lanking520, line 208: why we need this error requests?

maaquib: In case there is a issue with prefill for one of the requests for the rolling batch. If one of the requests has input length > max_seq_len we return that as an error request instead of failing the whole prefill batch

lanking520: If you belive we should have a global mechanism to handle this behavior, we can prepare another PR and do with all backends in 1 go

lanking520 (approved):

maaquib:
Nightly build: https://github.com/deepjavalibrary/djl-serving/actions/runs/6882455410
Large model integration tests: https://github.com/deepjavalibrary/djl-serving/actions/runs/6882839289
Rolling Batch Integration tests: https://github.com/deepjavalibrary/djl-serving/actions/runs/6883468941

PR #1297, opened 2023-11-08
[TRTLLM] split further for container layers

lanking520:
## Description ##

Build with more layers to speed up downloading

### DIFF
--- a/serving/docker/tensorrt-llm.Dockerfile
+++ b/serving/docker/tensorrt-llm.Dockerfile
@@ -65,10 +65,13 @@ RUN apt-get update && apt-get install -y wget unzip openmpi-bin libopenmpi-dev l
     pip3 cache purge && \
     apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
-# Install PyTorch and TensorRT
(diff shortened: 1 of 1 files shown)


serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -65,10 +65,13 @@ RUN apt-get update && apt-get install -y wget unzip openmpi-bin libopenmpi-dev l
     pip3 cache purge && \
     apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
-# Install PyTorch and TensorRT
+# Install PyTorch
 RUN pip install torch==${TORCH_VERSION} transformers==${transformers_version} accelerate==${accelerate_version} ${peft_wheel} sentencepiece \
```
frankfliu, line 69: Should torch on a single layer, torch is more than 1 G

lanking520: no, some of those library are trying to install a different version of torch. This is to keep them consistent

frankfliu (approved):

PR #1304, opened 2023-11-09
[CI] Adding one more machine for inf2 integration test

sindhuvahinis:
## Description ##

Adding one more machine for inf2 integration tests. Since we have 13 test cases now. Adding one more machine should reduce the runtime from 1 hour to 30 mins.


lanking520 (approved):
let's merge once migration is completed

sindhuvahinis:
> let's merge once migration is completed

manually pulled the changes cc @tosterberg

PR #1309, opened 2023-11-10
[TRTLLM] format better for logging

lanking520:
## Description ##

Improve logging and reduce crashes

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -112,14 +112,13 @@ def stop_on_any_exception(func):
     def try_catch_handling(self, input_data, parameters):
         try:
             return func(self, input_data, parameters)
-        except Exception as e:
+        except Exception:
             logging.exception("Rolling batch inference error")
-            err = json.dumps({"code": 500, "error": str(e)})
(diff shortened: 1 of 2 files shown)


rohithkrn (approved):

rohithkrn (approved):

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -112,14 +113,18 @@ def stop_on_any_exception(func):
     def try_catch_handling(self, input_data, parameters):
         try:
             return func(self, input_data, parameters)
-        except Exception as e:
+        except Exception:
             logging.exception("Rolling batch inference error")
-            err = json.dumps({"code": 500, "error": str(e)})
+            err = {
```
frankfliu, line 118: I don't think this is necessary. When process crashed, these data is not send back anyway

lanking520: this is generic for all rolling batch, if python exception reached, we send back all

frankfliu: I don't think we should send stacktrace to end user.

frankfliu (approved):

PR #1312, opened 2023-11-10
Change default top_k and temperature parameters in TRTLLM rolling batch

ydm-amazon:
## Description ##

Changes the default settings in TRTLLM rolling batch (top_k from 2 to 5 and the temperature from 0.7 to 0.8) to produce higher quality, less repetitive output. Also added default setting for top_p.

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
@@ -51,6 +51,7 @@ class TRTLLMRollingBatch(RollingBatch):
             parameters["runtime_top_k"] = parameters.get("runtime_top_k", 5)
             parameters["runtime_top_p"] = parameters.get("runtime_top_p", 0.85)
             parameters["temperature"] = parameters.get("temperature", 0.8)
+            parameters["repetition_penalty"] = parameters.get("repetition_penalty", 1.2)
         parameters["streaming"] = parameters.get("streaming", True)
         return parameters
 
--- a/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
@@ -48,10 +48,9 @@ class TRTLLMRollingBatch(RollingBatch):
         if "seed" in parameters.keys():
             parameters["random_seed"] = int(parameters.pop("seed"))
         if parameters.pop("do_sample", False):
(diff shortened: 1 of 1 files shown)


sindhuvahinis (dismissed):

engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
```diff
@@ -48,9 +48,8 @@ def translate_triton_params(self, parameters):
         if "seed" in parameters.keys():
             parameters["random_seed"] = int(parameters.pop("seed"))
         if parameters.pop("do_sample", False):
-            # TODO: Set better default values for do_sample
-            parameters["runtime_top_k"] = parameters.get("runtime_top_k", 2)
-            parameters["temperature"] = parameters.get("temperature", 0.7)
+            parameters["runtime_top_k"] = parameters.get("runtime_top_k", 5)
```
frankfliu, line 51: 1. 5 will have quit big impact on benchmark number.
2. We need add `repetition_penalty=1.2`

ydm-amazon: 1. @rohithkrn can probably give a better answer on this
2. Shared with you a quip documenting runs with different parameters, repetition_penalty seems like it's already set by default, as repetition_penalty=0.1 actually increases repetition. Also, repetition_penalty=1.2 does not help output quality

rohithkrn: @frankfliu it's just sampling the final output probabilities, I don't think changing from 2 to 5 will impact performance in a considerable way.

@ydm-amazon I also feel repetition_penalty should be defaulted to 1.2. That's what the paper recommends - https://arxiv.org/pdf/1909.05858.pdf. Currently there's no default for repetition penalty.

frankfliu: Base on my test, the difference is hug for TRTLLM (llama-2-13b and llama-2-70b)

frankfliu (approved):

engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
ydm-amazon: @rohithkrn I observed repetition penalty of 1.2 caused quality to sometimes fall apart (see row 14 of the table in the quip document)

PR #1315, opened 2023-11-13
[CI] Adds instance benchmark tabular recording

zachgk:
This improves the instant benchmark action with support to record the data in a DynamoDB table. It saves the input, serving.properties, and all accessible outputs as well. The table with some existing data can be seen [here](https://us-east-1.console.aws.amazon.com/dynamodbv2/home?region=us-east-1#item-explorer?operation=SCAN&table=RubikonBenchmarks) from this [input](https://gist.githubusercontent.com/zachgk/c0090ab15c00a3fda06b63f7a61a370a/raw/867f9b4db107b225b4993e8328f6c8bbf608e107/gistfile1.txt).

### DIFF
--- a/.github/workflows/instant_benchmark.yml
+++ b/.github/workflows/instant_benchmark.yml
@@ -26,7 +26,17 @@ on:
           - inf2.24xlarge
           - trn1.2xlarge
           - trn1.32xlarge
+      record:
+        description: 'Whether to record the results'
+        default: 'none'
+        type: choice
+        options:
+          - none
+          - table
 
+permissions:
+  id-token: write
+  contents: read
 
 jobs:
   create-runners:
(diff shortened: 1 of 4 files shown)


.github/workflows/instant_benchmark.yml
```diff
@@ -95,6 +110,19 @@ jobs:
           --job ${{ matrix.job }} --instance ${{ github.event.inputs.instance }} \
           --container deepjavalibrary/djl-serving:${{ github.event.inputs.container }}
           bash instant_benchmark.sh
+      - name: Configure AWS Credentials
```
lanking520, line 113: you don't have to assume role, the default role carried inside of hosted runner should have db write permission

lanking520: those permission only exposed to EC2 launched-runner

zachgk: I would like to leave it like that for now. In the future, we probably need to look at what role these runners are getting to help lock down our security. It should probably be either a new role or the same one I am explicitly assuming here with minimal permissions as used in our actions. Maybe even a blank role that has no permissions at all which would require this explicit role assumption to indicate the escalating permissions

lanking520 (approved):

PR #1329, opened 2023-11-17
[fix] Context length estimate datatype

sindhuvahinis:
## Description ##

After reinvent fix. TransformerNeuronX expects context length estimate as int or List of int. 

Testing:
Modified unit test case. 
Added context length estimate to CI and test it in my ec2 inf2 machine.

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/tnx_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/tnx_properties.py
@@ -14,7 +14,7 @@ import json
 import logging
 import os
 import re
-from typing import Optional
+from typing import Optional, Union, List
 
 from pydantic import validator, root_validator
 from enum import IntEnum, Enum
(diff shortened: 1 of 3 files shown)


lanking520 (approved):

tests/integration/llm/prepare.py
```diff
@@ -389,6 +389,7 @@
         "option.n_positions": 512,
         "option.model_loading_timeout": 2400,
         "option.load_split_model": True,
+        "option.context_length_estimate": [256, 512, 1024]
```
frankfliu, line 392: Why not just use a string: "256,512,1024"?

engines/python/setup/djl_python/properties_manager/tnx_properties.py
```diff
@@ -50,7 +50,7 @@ class TransformerNeuronXProperties(Properties):
     load_in_8bit: Optional[bool] = False
     low_cpu_mem_usage: bool = False
     load_split_model: bool = False
-    context_length_estimate: Optional[dict] = None
+    context_length_estimate: Optional[Union[List[int], int]] = None
```
frankfliu, line 53: In our case, we only need support List[int], user won't directly call this API

PR #1331, opened 2023-11-20
[python] Makes max output message size configurable

frankfliu:
## Description ##


engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
```diff
@@ -58,15 +58,23 @@ class PyProcess {
         this.pyEnv = pyEnv;
         this.workerId = workerId;
         int port = counter.getAndIncrement();
+        String size = model.getProperty("max_output_size");
```
zachgk, line 61: Can you document this on http://docs.djl.ai/docs/serving/serving/docs/configurations_model.html?

PR #1336, opened 2023-11-21
[doc] seq_scheduler_document

KexinFeng:
## Description ##

### DIFF
--- a/serving/docs/lmi/tutorials/seq_scheduler_tutorial.md
+++ b/serving/docs/lmi/tutorials/seq_scheduler_tutorial.md
@@ -1,0 +1,99 @@
+#  Seq-Scheduler Tutorial
+
+Seq-scheduler is a language model serving system built on native Huggingface inference backend, and works with any 
+huggingface model out-of-box. It features dynamic batching, dynamic batch splitting and prompt caching, with the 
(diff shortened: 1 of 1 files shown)


serving/docs/lmi/tutorials/seq_scheduler_tutorial.md
```diff
@@ -0,0 +1,99 @@
...
+# seq-scheduler parameters
+option.max_sparsity=0.33
+option.max_splits=3
+option.decoding_strategy=contrastive  # other options: greedy, sample
+```
+
+Then inside a djl-serving docker image, by running the following command, the server is started.
+```bash
+djl-serving -m <directory-of-serving.properties>
+```
```
sindhuvahinis, line 33: Why should it be inside a djl-serving? Users can use this serving.properties in docker run or in SageMaker right?

sindhuvahinis: No these will be added in DJL docs which will be shared with our customers. bashing into a docker image and running it, is un-neccessary and confusing for users. Here is another tutorial, we just published now https://docs.djl.ai/docs/serving/serving/docs/lmi/tutorials/trtllm_aot_tutorial.html

sindhuvahinis:
We should add some changes like, include a djl-demo on sagemaker with scheduler and replace awscurl command with curl command.  Approving it for now.

sindhuvahinis (approved):

PR #1347, opened 2023-11-28
[0.24.0-dlc] Fix for beam width with dynamic batching

sindhuvahinis:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/fastertransformer.py
+++ b/engines/python/setup/djl_python/fastertransformer.py
@@ -158,6 +158,11 @@ class FasterTransformerService(object):
         output_len = parameters.pop("max_seq_len", max_length)
         if self.use_triton:
             output_length = [output_len] * len(input_data)
+            if "beam_width" in parameters:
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/fastertransformer.py
```diff
@@ -158,6 +158,12 @@ def inference(self, inputs: Input):
         output_len = parameters.pop("max_seq_len", max_length)
         if self.use_triton:
             output_length = [output_len] * len(input_data)
+            if "beam_width" in parameters:
+                beam_width = parameters.pop('beam_width')
+                logging.info(f"beam width {beam_width}")
```
frankfliu, line 163: should this be debug level?

frankfliu (approved):

engines/python/setup/djl_python/fastertransformer.py
sindhuvahinis: removed it.

PR #1355, opened 2023-12-02
[TRTLLM][UX] add trtllm changes to support stop reason and also log prob

lanking520:
## Description ##

Add TRTLLM details function

Upgrade container as well

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
@@ -12,7 +12,7 @@
 # the specific language governing permissions and limitations under the License.
 import logging
 import tensorrt_llm_toolkit
-from djl_python.rolling_batch.rolling_batch import RollingBatch, stop_on_any_exception
+from djl_python.rolling_batch.rolling_batch import RollingBatch, stop_on_any_exception, Token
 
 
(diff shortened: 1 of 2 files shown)


frankfliu (approved):

sindhuvahinis (approved):

engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
```diff
@@ -63,17 +63,33 @@ def inference(self, input_data, parameters):
...
+            generation = trt_resp.fetch()
+            log_prob = generation.cum_logprob - self.request_cache[
+                request.id]["cumulative_logprob"]
+            self.request_cache[
+                request.id]["cumulative_logprob"] = generation.cum_logprob
+            token = Token(generation.token_id, generation.token_text, log_prob,
+                          None)
+            if generation.finished:
+                finish_reason = "eos_token" if generation.seq_length < self.request_cache[
+                    request.id]["out_length"] else "length"
```
rohithkrn, line 86: nit: i think we should call finished_reason as `max_length` instead of `length` if max_length is reached to make it more readable

lanking520: This is the openAI default name, TGI also used a similar name

rohithkrn (approved):

PR #1357, opened 2023-12-03
[UX][CI] fix a few bugs

lanking520:
## Description ##

Fixes the following issues

- P4 test timeout on Falcon-40B
- TRTLLM no Engine issue @frankfliu there is no engine given in properties in the test (since we don't provide), but MPI used
- lmi-dist some generation output is tensor other than python list

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
@@ -24,9 +24,6 @@ class TRTLLMRollingBatch(RollingBatch):
         :param properties: other properties of the model, such as decoder strategy
         """
         super().__init__(-1, **kwargs)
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/rolling_batch/trtllm_rolling_batch.py
```diff
@@ -24,9 +24,6 @@ def __init__(self, model_id_or_path, device, properties, **kwargs):
         :param properties: other properties of the model, such as decoder strategy
         """
         super().__init__(-1, **kwargs)
-        if properties.get("engine") != "MPI":
```
frankfliu, line 27: We can check `mpi_mode` instead of check engine.

We just need to manually set mpi_model in `PyModel`
```
properties.put("mpi_mode", "true");
``

PR #1365, opened 2023-12-05
[docker] Avoid JVM consume GPU memory

frankfliu:
## Description ##


xyang16 (approved):

lanking520:
can you apply to CU118 container as well? Or we abandon?

frankfliu:
This change is specific for LMI, if cu118 users might use java model, this is not necessary. If user use python model, they can manually set env `SERVING_OPTS=-Dai.djl.util.cuda.folk=true"

PR #1367, opened 2023-12-05
[python] Update lmi_dist warmup logic

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
@@ -12,8 +12,10 @@
 # the specific language governing permissions and limitations under the License.
 
 import os
+import logging
 from djl_python.rolling_batch.rolling_batch import RollingBatch, stop_on_any_exception, Token, FINISH_REASON_MAPPER
 from transformers import AutoConfig
+from lmi_dist.models import get_model
 from lmi_dist.utils.parameters import (
     NextTokenChooserParameters,
     StoppingCriteriaParameters,
@@ -75,7 +77,6 @@ class LmiDistRollingBatch(RollingBatch):
                 quantize = "bitsandbytes"
         if quantize is None and dtype == "int8":
             quantize = "bitsandbytes"
-        from lmi_dist.models import get_model
         self.model = get_model(
             model_id_or_path,
             revision=revision,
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -89,10 +90,41 @@ def _init_model(self, kwargs, model_id_or_path):
             self._warmup(**kwargs)
 
     def _warmup(self, **kwargs):
-        batch_size = int(self.properties.get("max_rolling_batch_size", 32))
-        max_batch_prefill_tokens = int(
-            self.properties.get("max_rolling_batch_prefill_tokens", -1))
-        self.model.warmup(batch_size, max_batch_prefill_tokens)
+        max_prefill_tokens = int(
+            self.properties.get("max_rolling_batch_prefill_tokens", 4096))
```
lanking520, line 94: the purpose of `max_rolling_batch_prefill_tokens` has changed. Do we still want to keep this concept? And why we need now to introduce a max_prefill_tokens, can we just set some small value like 512, 1024?

lanking520: and why we start to bring this to external, can this be done inside lmi-dist?

xyang16: This was external before.

lanking520: why we set to 4096 such small value?  Does that have impact to the total token we can support? Or on the other hand, if customer set this value to be very high, like 32000, will this impact model loading?

tests/integration/llm/prepare.py
```diff
@@ -773,6 +773,7 @@ def build_lmi_dist_model(model):
         )
     options = lmi_dist_model_list[model]
     options["engine"] = "MPI"
+    options["option.mpi_mode"] = "true"
```
lanking520, line 776: this is not needed? Given engine is MPI

xyang16: Because this change require this property: https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py#L42

xyang16: I got error with this.

PR #1377, opened 2023-12-10
allow git permission to aws assumed role

lanking520:
## Description ##

fix the lmi-dist deps build pipeline

### DIFF
--- a/.github/workflows/lmi-dist-deps-build.yml
+++ b/.github/workflows/lmi-dist-deps-build.yml
@@ -3,6 +3,10 @@ name: LMI-Dist dependency build
 on:
   workflow_dispatch:
 
+permissions:
+  id-token: write
+  contents: read
+
 jobs:
   create-runners-p4d:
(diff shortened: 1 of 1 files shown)


.github/workflows/lmi-dist-deps-build.yml
```diff
@@ -3,6 +3,10 @@ name: LMI-Dist dependency build
 on:
   workflow_dispatch:
 
+permissions:
```
frankfliu, line 6: What this permission is used for?

lanking520: permission is used for grant access to secuity tokens: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

frankfliu: Why we don't need this before?

frankfliu (approved):

PR #1385, opened 2023-12-12
[CI][fix] Sagemaker integration test cloudwatch metrics fix

sindhuvahinis:
## Description ##

Sagemaker metrics were not uploaded to cloud watch, because the github action inputs will be empty when the workflow is dispatched through schedule. 

Changes in this PR:
- Made `run_benchmark` an argument to sagemaker-endpoint-tests script. 
- Uploading metrics for MME also, (before it was only for SME)
- Metric name format will be `{name}-{engine_name}-{num_partitions}p-{instance_type}-metric`, for example `opt-1-3-b-deepspeed-2p-ml.g5.8xlarge-throughput`

Test:
- Tested MME and SME 1 test case in my ec2 machine.

### DIFF
--- a/tests/integration/llm/sagemaker-endpoint-tests.py
+++ b/tests/integration/llm/sagemaker-endpoint-tests.py
@@ -9,6 +9,11 @@ from sagemaker.utils import unique_name_from_base
 from argparse import ArgumentParser
 import numpy as np
 
+
+def boolean_arg(value):
+    return str(value).lower() == "true"
+
+
 parser = ArgumentParser(
     description=
     "This Script deploys a model with predefined configuration to a SageMaker Inference Endpoint"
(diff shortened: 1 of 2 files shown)


tests/integration/llm/sagemaker-endpoint-tests.py
```diff
@@ -256,13 +264,21 @@ def mme_test(name, image_type):
...
-            config.get("instance_type", DEFAULT_INSTANCE_TYPE),
+            instance_type,
             serializer=sagemaker.serializers.JSONSerializer(),
             deserializer=sagemaker.deserializers.JSONDeserializer())
-        for model in list(mme.list_models()):
+        for i, model in enumerate(list(mme.list_models())):
             outputs = predictor.predict(DEFAULT_PAYLOAD, target_model=model)
             print(outputs)
 
+            if run_benchmark:
```
siddvenk, line 274: do we convert `run_benchmark` to a bool anywhere? this would run if the value is "true" or "false", right (string values)?

sindhuvahinis: Good one, missed it. Handled it now.

lanking520 (approved):

siddvenk (approved):

PR #1396, opened 2023-12-13
[CI][fix] remove g5xl and introduce rolling batch in lmic

sindhuvahinis:
## Description ## 
- Removed the g5xl and removed huggingface testing. 
- Replaced huggingface testing with lmi dist test. 

## Testing 
Tested through running in our CI 
https://github.com/deepjavalibrary/djl-serving/actions/runs/7266944676 
Note that gptj failed because of lmi dist gptj is not supported in the above runner. `sharded is not supported for AutoModel` Lmidist work with non shared version though. So removed the gpt-j-lmi-dist for now.

### DIFF
--- a/.github/workflows/lmic_performance.yml
+++ b/.github/workflows/lmic_performance.yml
@@ -15,15 +15,6 @@ jobs:
   create-runners:
     runs-on: [self-hosted, scheduler]
     steps:
-      - name: Create new G5XL instance
-        id: create_gpu_xl
-        run: |
-          cd /home/ubuntu/djl_benchmark_script/scripts
-          token=$( curl -X POST -H "Authorization: token ${{ secrets.ACTION_RUNNER_PERSONAL_TOKEN }}" \
-          https://api.github.com/repos/deepjavalibrary/djl-serving/actions/runners/registration-token \
-          --fail \
-          | jq '.token' | tr -d '"' )
-          ./start_instance.sh action_lmic_g5 $token djl-serving
       - name: Create new G5 instance
         id: create_gpu
         run: |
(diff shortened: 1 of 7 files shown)


zachgk:
Is this working? I think you also need to specify what IAM Role/Account to assume and ensure that the role has the necessary permissions. If you are trying to use the newly created role (as opposed to the older IAM account), it doesn't have cloudwatch permissions

.github/workflows/lmic_performance.yml
```diff
@@ -15,42 +15,47 @@ jobs:
   create-runners:
```
lanking520, line 15: all of the runner should already have many permission to the serivce. Which specific one you are looking for?

sindhuvahinis: Cloudwatch put metric data permission

sindhuvahinis:
> Is this working? I think you also need to specify what IAM Role/Account to assume and ensure that the role has the necessary permissions. If you are trying to use the newly created role (as opposed to the older IAM account), it doesn't have cloudwatch permissions

@zachgk running it manually worked.

sindhuvahinis:
Will do some more changes on introducing rolling batch and removing the g5 xl machine and mark this ready for review.

tests/integration/llm/prepare.py
```diff
@@ -200,119 +200,136 @@
...
     },
     "gpt-neox-20b-fp16-deepspeed": {
         "option.task": "text-generation",
         "option.dtype": "fp16",
-        "option.engine": "deepspeed",
-        "option.model_id": "s3://djl-llm/gpt-neox-20b/"
+        "engine": "DeepSpeed",
+        "option.model_id": "s3://djl-llm/gpt-neox-20b/",
+        "option.parallel_loading": "true",
+        "option.rolling_batch": "deepspeed",
```
frankfliu, line 308: Does "auto" work for DS?

frankfliu (approved):

tests/integration/llm/prepare.py
sindhuvahinis: Yes, it works.

tosterberg (approved):

PR #1401, opened 2023-12-15
[docs] fix tnx n_positions description

tosterberg:
## Description ##

Update docs to better reflect the meaning of the config for option.n_positions.


serving/docs/lmi/configurations_large_model_inference_containers.md
```diff
@@ -96,7 +96,7 @@ If you are using Neuron container and engine set to Python, the following parame
...
-| option.n_positions             | No       | Input sequence length                                                                                                                                                                                                                                                                                                                                 | Default: `128`                                                                                 |
+| option.n_positions             | No       | Total sequence length, input sequence length + output sequence length.                                                                                                                                                                                                                                                                                | Default: `128`                                                                                 |
```
frankfliu, line 99: We need consider consolidate with max_rolling_batch_size when rolling batch is enabled.

frankfliu (approved):

serving/docs/lmi/configurations_large_model_inference_containers.md
tosterberg: Agreed - I will work on a PR to update the CX, and default behaviors.

PR #1404, opened 2023-12-15
[TNX][config] update rolling batch batch size behavior and docs

tosterberg:
## Description ##

This PR specifies the use of `batch_size` and `max_rolling_batch_size` in TNX properties. Specifically, this removes the need for `option.batch_size` for all but one case (batch size > 1 and streaming enabled). Tests and docs are updated to reflect the change.

One additional constraint was added to the properties, which is if `batch_size` > 1 and `rolling_batch` is enabled will now throw an error. This behavior should have existed previously, but due to being unable to differentiate `batch_size` and `option.batch_size` when setting properties, it was not detectable.

One warning that was previously thrown, was removed. Originally, a warning was thrown when `batch_size` > 1 and `enable_streaming` was enabled. Due to our inability to differentiate dynamic batching `batch_size` and `option.batch_size` this warning is erroneous.

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/tnx_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/tnx_properties.py
@@ -82,17 +82,18 @@ class TransformerNeuronXProperties(Properties):
         return rolling_batch
 
     @validator('batch_size')
-    def validate_batch_size(cls, batch_size, values):
+    def validate_batch_size(cls, batch_size: int, values) -> int:
         """
-        Transformer neuronx has both option.batch_size and batch_size.
-        option.batch_size is to compile the model with batch size, which cannot be
-        differentiated in neuronx handlers. Hence, just throwing a warning here.
+        Transformer neuronx supports batch_size, and max_rolling_batch_size.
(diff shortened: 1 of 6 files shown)


engines/python/setup/djl_python/properties_manager/tnx_properties.py
```diff
@@ -82,17 +82,18 @@ def validate_rolling_batch(cls, rolling_batch: str) -> str:
...
-        differentiated in neuronx handlers. Hence, just throwing a warning here.
+        Transformer neuronx supports batch_size, and max_rolling_batch_size.
+        The batch_size param is for dynamic batching, and max_rolling_batch_size is for rolling batch.
+        We validate here that the values are compatible and set for both compilation and inference.
         """
         if batch_size > 1:
-            if values['rolling_batch'] == RollingBatchEnum.disable and values[
-                    'enable_streaming'] != StreamingEnum.false:
-                logging.warning(
-                    "We cannot enable streaming for dynamic batching")
```
sindhuvahinis, line 95: So, now we support streaming with dynamic batch?

sindhuvahinis (approved):

PR #1406, opened 2023-12-16
[CI][IB] Parses template from S3 and local dir

zachgk:
This adds more possible source locations for a template besides http, specifically local files and s3://bucketName/key.

This can be used for several purposes. First, it can support debugging and template locally. Afterwards, it can allow more hidden or secure templates.

### DIFF
--- a/.github/workflows/instant_benchmark.yml
+++ b/.github/workflows/instant_benchmark.yml
@@ -122,6 +122,7 @@ jobs:
         run: |
           python3 record_benchmark.py --template template.json \
           --job ${{ matrix.job }} --instance ${{ github.event.inputs.instance }} \
+          --model models/test \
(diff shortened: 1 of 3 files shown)


frankfliu (approved):

.github/workflows/instant_benchmark.yml
```diff
@@ -67,6 +67,17 @@ jobs:
       - name: Clean env
         run: |
           yes | docker system prune -a --volumes
+      - name: install deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install awscli -y
```
lanking520, line 73: use pip install awscli

zachgk: Right now, we use the apt-get install in most places in CI with only a single exception using the pip install. It would be better to follow the convention in this PR and then change all usages to `pip install` in a following one

PR #1409, opened 2023-12-18
[Refactor] lmi dist rolling batch properties

sindhuvahinis:
## Description ##

Extracted the properties of lmi-dist rolling batch. 

Testing 
- Tested them in my EC2 machine. 
- Tested in CI using my local branch https://github.com/deepjavalibrary/djl-serving/actions/runs/7279838457

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
@@ -43,9 +43,8 @@ class LmiDistRbProperties(Properties):
     def validate_mpi_mode(cls, properties):
         if not properties.get("is_mpi") and int(
                 properties.get("tensor_parallel_degree", "1")) != 1:
-            raise ValueError(
-                f"Need mpi_mode to start lmi-dist RollingBatcher."
-                f"Try with engine=MPI in your serving.properties")
+            raise ValueError(f"Need mpi_mode to start lmi-dist RollingBatcher."
+                             f"Try with engine=MPI in your serving.properties")
         return properties
 
     @root_validator()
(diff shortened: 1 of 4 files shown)


engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
```diff
@@ -0,0 +1,85 @@
...
+# the specific language governing permissions and limitations under the License.
+import logging
+import os
+from enum import Enum
+from typing import Optional
+
+import torch
+from pydantic.class_validators import root_validator
+
+from djl_python.properties_manager.hf_properties import get_torch_dtype_from_str
```
tosterberg, line 21: Do we use this mapper anywhere else? If so it might be a good idea to move it out of a specific property implementation into a utils file.

engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
```diff
@@ -0,0 +1,85 @@
...
+                             f"Try with engine=MPI in your serving.properties")
+        return properties
+
+    @root_validator()
+    def validate_quantize(cls, properties):
+        if properties.get('quantize') is None:
+            if properties.get('dtype') == "int8":
+                properties['quantize'] = LmiDistQuantizeMethods.bitsandbytes
+        else:
+            os.environ["CUDA_MEMORY_FRACTION"] = "0.9"
```
tosterberg, line 56: I am somewhat aware of why we are setting this, so that we don't OOM during rolling batch. If we have some specific guidance on why we use it, when it should be used, and if we can ever remove the constraint and get that into a comment in here since we are limiting our memory management options with it.

tosterberg (approved):

xyang16 (approved):

engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
sindhuvahinis: Yes will do it in the next ones

PR #1442, opened 2024-01-03
Support peft models in LMI Dist

xyang16:
## Description ##


lanking520 (commented):
Could you add the new introduced member variable here? https://docs.djl.ai/docs/serving/serving/docs/lmi/configurations_large_model_inference_containers.html#lmi-dist Also, could we just determine wether the model is peft or not ourselves?

xyang16:
> Could you add the new introduced member variable here? https://docs.djl.ai/docs/serving/serving/docs/lmi/configurations_large_model_inference_containers.html#lmi-dist Also, could we just determine wether the model is peft or not ourselves?

I have updated the docs. I can't find a reliable way to determine if it's peft model yet.

siddvenk:
There might be a better way now, but we have auto detection of peft models in some of our handlers like here https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/deepspeed.py#L173.

If there is a configuration file `adapter_config.json` that we can load from PeftConfig.from_pretrained, we consider it a peft model (we assume all peft models are Lora models there)

lanking520:
@xyang16 could we just add this simple check somewhere to check for normal model or peft model?

https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/deepspeed.py#L174-L192

PR #1443, opened 2024-01-04
[LMIDist] Add test that calls `LmiDistRollingBatch` with python script

KexinFeng:
## Description ##

This adds a unit test that calls `LmiDistRollingBatch` with python script, for the following purpose:
1. Easiness of further development, e.g. speculative decoding
2. Use the greedy output text as benchmark, to prevent possible text quality degradation from model level modification
3. Easy demonstration and benchmarking
4. Easy debugging

## Next step ##
In the next step, this test can be added to integration test: .github/workflows/rolling_batch_integration.yml
which will require writing a new shell script for running `python3 file` in docker.

### DIFF
--- a/engines/python/setup/djl_python/tests/test_lmi_dist.py
+++ b/engines/python/setup/djl_python/tests/test_lmi_dist.py
@@ -45,11 +45,18 @@ class TestLmiDist(unittest.TestCase):
             # "codellama/CodeLlama-7b-hf"
         ]
         expected_text_30 = {
-            "TheBloke/Llama-2-13B-Chat-fp16": {1: 'Hello, my name is Dr. [Last Name] and I am a licensed clinical psychologist with over 10 years of experience working with children, ad',
-              2:'The president of the United States is the head of the executive branch and the highest-ranking official in the federal government. The president is elected by the people through the Electoral',
-              3: 'The capital of France is Paris.\n\nThe capital of Germany is Berlin.\n\nThe capital of Italy is Rome.\n\nThe capital of Spain is Madrid.',
-              4: 'The future of AI is not just about building smarter machines, but also about ensuring that these machines are used for the betterment of society.\n\nAs A',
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/tests/test_lmi_dist.py
```diff
@@ -0,0 +1,141 @@
+import unittest
```
frankfliu, line 1: Please add license header.

engines/python/setup/djl_python/tests/test_lmi_dist.py
```diff
@@ -0,0 +1,141 @@
...
+
+import torch
+import os, sys
+
+# script_directory = os.path.dirname(os.path.abspath(__file__))
```
frankfliu, line 6: remove commented out code

engines/python/setup/djl_python/tests/test_lmi_dist.py
```diff
@@ -0,0 +1,164 @@
...
+            # TODO: fix this. weight model.layers.0.self_attn.rotary_emb.inv_freq does not exist
+            # "TheBloke/Llama-2-7B-Chat-AWQ",
+            "TinyLlama/TinyLlama-1.1B-Chat-v0.6",
+            # TODO: fix this. weight model.layers.0.self_attn.rotary_emb.inv_freq does not exist
+            # "TinyLlama/TinyLlama-1.1B-python-v0.1",
+            # g5.12xlarge single gpu ok. But no way to clear the gpu memory after running llama-2-7b thus cause OOM
+            # "codellama/CodeLlama-7b-hf"
+        ]
+        expected_text_30 = {
+            "TheBloke/Llama-2-7B-Chat-fp16": {
```
lanking520, line 47: let's remove this then

lanking520: But it is just too big to test on a CPU machine. We could use a better or smaller model for evaluation and testing

lanking520 (approved):

PR #1450, opened 2024-01-04
add awscurl build pipeline

lanking520:
## Description ##

Add awscurl build

### DIFF
--- a/.github/workflows/awscurl-build.yml
+++ b/.github/workflows/awscurl-build.yml
@@ -1,0 +1,47 @@
+name: awscurl build
+
+on:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/awscurl-build.yml
```diff
@@ -0,0 +1,47 @@
...
+
+on:
+  workflow_dispatch:
+
+jobs:
+  create-runners:
+    runs-on: [ self-hosted, scheduler ]
+    steps:
+      - name: Create new Graviton instance
+        id: create_aarch64
```
frankfliu, line 11: Why need aarch64?

We should just update djl-serving publish pipeline in djl repo

PR #1459, opened 2024-01-05
[TRTLLM] Fix bug in handler

ydm-amazon:
## Description ##

Fixes bug in TensorRT-LLM handler that prevents some error messages from being displayed to user


lanking520 (commented):
self.rolling_batch_type this also can kill too

ydm-amazon:
> self.rolling_batch_type this also can kill too

Done

lanking520 (approved):

PR #1460, opened 2024-01-08
[TRTLLM] add enable kv cache reuse

lanking520:
## Description ##

Allow reusable KVCache


tests/integration/llm/prepare.py
```diff
@@ -725,6 +725,7 @@
     "mistral-7b": {
         "option.model_id": "s3://djl-llm/mistral-7b/",
         "option.tensor_parallel_degree": 4,
+        "option.enable_kv_cache_reuse": "true",
```
rohithkrn, line 728: this option needs to be added in the toolkit to get updated.

rohithkrn: my bad, it's already there.

rohithkrn (approved):

PR #1468, opened 2024-01-10
[RollingBatch] add customized rollingbatch

lanking520:
## Description ##

Support customized rollingbatch function

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -227,6 +227,8 @@ class RollingBatch(ABC):
             self.output_formatter = _jsonlines_output_formatter
         elif "none" == formatter:
             pass
+        elif callable(formatter):
+            self.output_formatter = formatter
         else:
             # TODO: allows to load custom formatter from a module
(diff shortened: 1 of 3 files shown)


sindhuvahinis (approved):

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -227,6 +227,8 @@ def __init__(self, **kwargs):
             self.output_formatter = _jsonlines_output_formatter
         elif "none" == formatter:
             pass
+        elif callable(formatter):
```
frankfliu, line 230: How we pass this from `serving.properties`? Where is the code load this callable?

lanking520: user cannot pass this through serving.properties, since bunch of the postprocessing logic is not descriptable. This option is just trying to make it easy for customer to extend from the base class

frankfliu: We should be able load the module by itself name

frankfliu (approved):

PR #1472, opened 2024-01-10
[CI] add 2 inf2 rolling batch tests

lanking520:
## Description ##

Support full rolling batch integration tests for INF2

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -497,6 +497,24 @@ transformers_neuronx_handler_list = {
         "option.tensor_parallel_degree": 2,
         "option.dtype": "bf16",
         "option.use_stable_diffusion": True
+    },
+    "llama-7b-rb": {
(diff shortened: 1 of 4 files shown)


tests/integration/llm/prepare.py
```diff
@@ -497,6 +497,24 @@
...
+        "option.load_split_model": True
+    },
+    "mistral-7b-rb": {
+        "option.model_id": "s3://djl-llm/mistral-7b/",
+        "option.tensor_parallel_degree": 4,
+        "option.n_positions": 512,
+        "option.max_rolling_batch_size": 4,
+        "option.rolling_batch": 'auto',
+        "option.model_loading_timeout": 2400,
+        "option.load_split_model": True
```
tosterberg, line 517: Is mistral also a split model, I think it is just the HF model.

tosterberg (approved):

tests/integration/llm/prepare.py
ydm-amazon: Yep, it's just the model (similar to Llama 7B)

PR #1478, opened 2024-01-11
[tnx] add gqa to properties

siddvenk:
## Description ##

GQA feature is in beta currently, and only applies to Llama and Mistral model types.

I think there are some issues on the neuronx side, specifically with the config shard-over-batch. That doesn't seem to work, so I'll raise an issue to them. 

GQA flags are ignored for non supported model types, and TNX validates and sets a fallback value in case the user-specified value doesn't work. I've tested with llama70b and gpt2 to validate that the configs work and are processed by neuron, but no correctness/benchmark testing is done


engines/python/setup/setup.py
```diff
@@ -57,7 +57,8 @@ def run(self):
 
     test_requirements = [
         'numpy', 'requests', 'Pillow', 'transformers', 'torch', 'einops',
-        'accelerate', 'sentencepiece', 'protobuf', 'yapf', 'pydantic==1.10.13'
+        'accelerate', 'sentencepiece', 'protobuf', 'yapf', 'pydantic==1.10.13',
+        'transformers_neuronx'
```
sindhuvahinis, line 61: Is the unit test cases not checked in yet?

lanking520: oh, why we add tnx as dependencies? Could we do mocking?

siddvenk: yea i guess we could mock. I wanted to avoid replicating the enum defined by transformers neuronx in the main code. It might be easiest to just replicate

sindhuvahinis (approved):

PR #1495, opened 2024-01-17
[trtllm handler] Add checks on input data

ydm-amazon:
## Description ##

Adds checks on input data to fix issue with custom preprocessing

### DIFF
--- a/engines/python/setup/djl_python/tensorrt_llm.py
+++ b/engines/python/setup/djl_python/tensorrt_llm.py
@@ -69,12 +69,21 @@ class TRTLLMService(object):
 
         return input_data, input_size, parameters, errors, batch
 
+
+def is_broken_input_data(input_data):
+    if len(input_data) > 0 and isinstance(input_data[0],
+                                          dict) and input_data != [{
+                                              'inputs': ''
+                                          }]:
+        return True
+    return False
+
     def inference(self, inputs):
         outputs = Output()
 
         input_data, input_size, parameters, errors, batch = self.parse_input(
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -69,12 +69,17 @@ def parse_input(self, inputs):
 
         return input_data, input_size, parameters, errors, batch
 
+def is_broken_input_data(input_data):
+    if len(input_data) > 0 and isinstance(input_data[0], dict) and input_data != [{'inputs': ''}]:
```
lanking520, line 73: why we are checking the last piece?

```
[{'inputs': ''}]
```

ydm-amazon: See quip docs for details

But long story short for all tokens except for first token, input_data is [{'inputs': ''}] no matter the actual request body

frankfliu: `[{'inputs': ''}]` is expected, why this treat as broken input data?

ydm-amazon: Exactly
It is expected so if the input_data is `not` equal to `[{'inputs': ''}]` but still contains a dictionary, then it is broken

ydm-amazon:
Ran gradle formatter to pass checks

ydm-amazon:
Fixing issue at toolkit level. Closing.

PR #1502, opened 2024-01-20
Align properties parsing to be similar to java

ydm-amazon:


serving/docker/partition/utils.py
```diff
@@ -113,5 +113,7 @@ def load_properties(properties_dir):
                 if line.startswith("#") or not line.strip():
                     continue
                 key, value = line.strip().split('=', 1)
+		key = key.strip()
```
frankfliu, line 116: Fix indentation issue

lanking520 (approved):

PR #1521, opened 2024-01-27
T5 support with python backend in TensorRT LLM

sindhuvahinis:
## Description ##

This PR is to support  t5 model with python backend of TRTLLM. 

serving.properties
```
option.model_id={t5-model}
option.entryPoint=djl_python.tensorrt_llm
option.rolling_batch=disable
```

There are a few things to be do, which will be done in other PRs.
1. To make JIT compilation for python backend
2. Infer entrypoint automatically for tensorrt llm
3. Log probs and other details for the generated text. 
4. Add integration test cases for t5 model

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/trt_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/trt_properties.py
@@ -14,7 +14,7 @@ from djl_python.properties_manager.properties import Properties, RollingBatchEnu
 from pydantic import validator
 
 TRT_SUPPORTED_ROLLING_BATCH_TYPES = [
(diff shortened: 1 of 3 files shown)


lanking520 (commented):
general thoughts, we should switch to dynamic batching for customer to use. We also need some CI testing to cover:

1. Basic inference run with T5 model
2. Some additional document on preparing the artifacts

lanking520:
as discussed, let's think about introducing the

```
option.max_dymanic_batch_size=<>
```
to avoid confusion. So customer either set

```
option.max_rollingbatch_size=<>
```
or
```
option.max_dynamic_batch_size=<>
```

for their inference

PR #1537, opened 2024-02-13
[Test] build test handler

lanking520:
## Description ##

Create an engine that everyone can simply iterate with and playing around without using model server

### DIFF
--- a/engines/python/setup/djl_python/test_model.py
+++ b/engines/python/setup/djl_python/test_model.py
@@ -13,13 +13,17 @@
 
 import logging
 import os
+import struct
 import sys
+from types import ModuleType
+from typing import List, Dict, Union
 
+from djl_python import PairList
 from .arg_parser import ArgParser
 from .inputs import Input
 from .outputs import Output
 from .np_util import to_nd_list
-from .service_loader import load_model_service
+from .service_loader import load_model_service, ModelService
 
 
 def create_request(input_files, parameters):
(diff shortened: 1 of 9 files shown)


engines/python/setup/djl_python/test_model.py
```diff
@@ -103,27 +131,85 @@ def create_npz_request(list, key: str = None) -> Input:
...
+    return kwargs
+
+
+class TestHandler:
+
+    def __init__(self,
+                 entry_point: Union[str, ModuleType],
+                 model_dir: str = None):
+        self.serving_properties = load_properties(model_dir)
+        self.serving_properties = update_properties_with_env_vars(
```
frankfliu, line 190: env var is default value, we should call update_properties_with_env_vars first, then load properties

engines/python/setup/djl_python/test_model.py
```diff
@@ -103,27 +131,85 @@ def create_npz_request(list, key: str = None) -> Input:
...
+            self.serving_properties)
+        if isinstance(entry_point, str):
+            os.chdir(model_dir)
```
frankfliu, line 193: Should we do this outside of TestHandler?

serving/docker/partition/trt_llm_partition.py
```diff
@@ -38,9 +37,9 @@ def update_kwargs_with_env_vars(kwargs):
     env_vars = os.environ
     for key, value in env_vars.items():
         if key.startswith("OPTION_"):
-            key = key[7:].lower()
-            if key == "entrypoint":
-                key = "entryPoint"
+            key = key.lower()
+            if key == "option.entrypoint":
```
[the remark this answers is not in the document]
lanking520: good catch, let me update this

rohithkrn:
can you add a readme file with instructions to use?

frankfliu (approved):

PR #1539, opened 2024-02-14
Updates vllm to 0.3.1

zachgk:
It also migrates to pydantic>2 as that is a new dependency with vllm 0.3.0. It uses the migration mode for now, which makes it easier to revert back to pydantic 1 if the state of dependencies change


engines/python/setup/djl_python/properties_manager/ds_properties.py
```diff
@@ -6,7 +6,7 @@
 import torch
 from typing import Optional, Any
 
-from pydantic import root_validator, validator, Field
+from pydantic.v1 import root_validator, validator, Field
```
frankfliu, line 9: Why not migrate to v2 directly?

lanking520 (approved):

engines/python/setup/djl_python/properties_manager/ds_properties.py
lanking520: v2 still have ongoing changes

lanking520: we have to keep maintaining it

frankfliu (approved):

serving/docker/deepspeed.Dockerfile
```diff
@@ -19,7 +19,7 @@ ARG torch_version=2.1.2
 ARG torch_vision_version=0.16.2
 # HF Deps
 ARG protobuf_version=3.20.3
-ARG transformers_version=4.36.2
+ARG transformers_version=4.37.2
 ARG accelerate_version=0.25.0
```
lanking520, line 23: acclerate may also need to update

PR #1547, opened 2024-02-19
[vLLM] add enforce eager as an option

lanking520:
## Description ##

This option will help to reduce CUDA graph GPU memory addition under some extreme scenarios


ydm-amazon (approved):

maaquib:
Wondering if we should also add [gpu-memory-utilization](https://github.com/vllm-project/vllm/blob/e433c115bce2bf27f7b1abdde7029566007d9eee/vllm/engine/arg_utils.py#L185-L191) flag to configure more memory for CUDAGraph [[link](https://github.com/vllm-project/vllm/blob/e433c115bce2bf27f7b1abdde7029566007d9eee/vllm/worker/model_runner.py#L670-L674)]

```
        logger.info("CUDA graphs can take additional 1~3 GiB memory per GPU. "
                    "If you are running out of memory, consider decreasing "
                    "`gpu_memory_utilization` or enforcing eager mode. "
                    "You can also reduce the `max_num_seqs` as needed "
                    "to decrease memory usage.")
```

lanking520:
@maaquib there are negative impact on this parameter since the general pagedattention will need to use extra GPU spaces to get tokens calculated. I can add this, but not recommend to change on this for particular use cases, it's hard to balance the moving variants of two factors

PR #1552, opened 2024-02-20
[CI][Deps build] upgrade torch to 2.1.2

lanking520:
## Description ##

change the order of build to make vllm all the way to the end.


ydm-amazon:
Is the order important because the speculative decoding depends on llm-awq?

lanking520:
@ydm-amazon no, it is more or less change the pytorch version, the title is misleading

ydm-amazon (approved):

PR #1554, opened 2024-02-21
[docker] map select tgi env vars to lmi env vars

siddvenk:
## Description ##

This PR adds support for mapping some TGI environment variables to equivalent LMI environment variables.

The set of environment variables may change based on more discussion, but this at least sets it up so we can add/remove from the list.

This change is only added to the deepspeed/trtllm containers, not NeuronX.

---
Tested locally with: ` docker run -it --runtime=nvidia -p 8080:8080 --shm-size=2g -e MODEL_ID=TheBloke/Llama-2-13b-fp16 -e SM_NUM_GPUS=max -e MAX_BATCH_SIZE=32 -e OPTION_MAX_ROLLING_BATCH_SIZE=64 deepjavalibrary/djl-serving:deepspeed`

Logs show:
```
Environment variables:
	HF_HUB_ENABLE_HF_TRANSFER: 1
	HF_HOME: /tmp/.cache/huggingface
	OPTION_MAX_ROLLING_BATCH_SIZE: 64
	OMP_NUM_THREADS: 1
	HF_MODEL_ID: TheBloke/Llama-2-13b-fp16
	OPTION_TENSOR_PARALLEL_DEGREE: max
	DJL_CACHE_DIR: /tmp/.djl.ai
```

### DIFF
--- a/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
+++ b/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
@@ -5,6 +5,16 @@ verlte() {
     [ "$1" = "$2" ] && return 1 || [ "$1" = "`echo -e "$1\n$2" | sort -V | head -n1`" ]
 }
 
+# Takes 2 environment variables: SM/TGI env var name, LMI env var name
+# If SM/TGI env var is set, and LMI env var is unset, set LMI env var to SM/TGI env var value.
+translateTGIToLMI() {
+  local tgiVal="${!1}"
+  local lmiVal="${!2}"
+  if [ -n "$tgiVal" ]; then
+    export "$2"=${lmiVal:-$tgiVal}
+  fi
+}
+
 # Follow https://docs.aws.amazon.com/sagemaker/latest/dg/inference-gpu-drivers.html
 if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
     CUDA_COMPAT_MAX_DRIVER_VERSION=$(readlink /usr/local/cuda/compat/libcuda.so.1 |cut -d'.' -f 3-)
@@ -21,6 +31,15 @@ if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
 else
     echo "Skip CUDA compat libs setup as package not found"
 fi
+
+# Convert select SM/TGI Environment Variables to LMI Equivalents
+translateTGIToLMI "HF_MODEL_QUANTIZE" "OPTION_QUANTIZE"
+# We use HF_TRUST_REMOTE_CODE in our properties parsing
+translateTGIToLMI "HF_MODEL_TRUST_REMOTE_CODE" "HF_TRUST_REMOTE_CODE"
+translateTGIToLMI "SM_NUM_GPUS" "TENSOR_PARALLEL_DEGREE"
+translateTGIToLMI "MAX_CONCURRENT_REQUESTS" "SERVING_JOB_QUEUE_SIZE"
+translateTGIToLMI "MAX_BATCH_PREFILL_TOKENS" "OPTION_MAX_ROLLING_BATCH_PREFILL_TOKENS"
+translateTGIToLMI "MAX_BATCH_SIZE" "OPTION_MAX_ROLLING_BATCH_SIZE"
 
 if [[ "$1" = "serve" ]]; then
     shift 1


serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -22,6 +30,14 @@ else
     echo "Skip CUDA compat libs setup as package not found"
 fi
 
+# Convert TGI Environment Variables to LMI Equivalents
+translateTGIToLMI "MODEL_ID" "HF_MODEL_ID"
```
lanking520, line 34: how about we map to `OPTION_MODEL_ID`

siddvenk: That makes sense. Looking through the code there are two things that stand out to me:
- We support HF_MODEL_ID and OPTION_MODEL_ID (though not in the same ways). Need to fix that
- We support TENSOR_PARALLEL_DEGREE and OPTION_TENSOR_PARALLEL_DEGREE. Need to consolidate

lanking520: I think we can stick with OPTION_MODEL_ID and also OPTION_TENSOR_PARALLEL_DEGREE. @frankfliu what is your thoughts on this. And gradually remove HF_MODEL_ID and TENSOR_PARALLEL_DEGREE usage in our code base.

siddvenk:
I have added some more mappings based on the SM examples from https://github.com/aws/amazon-sagemaker-examples/blob/main/inference/generativeai/huggingfacetgi/tgi-hosting/models.json.

serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -22,6 +32,17 @@ else
     echo "Skip CUDA compat libs setup as package not found"
 fi
 
+# Convert select SM/TGI Environment Variables to LMI Equivalents
+translateTGIToLMI "HF_MODEL_ID" "OPTION_MODEL_ID"
```
frankfliu, line 36: We already handled HF_MODEL_ID AND HF_TASK in our model server code

serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -22,6 +32,17 @@ else
...
+translateTGIToLMI "HF_MODEL_REVISION" "OPTION_REVISION"
+translateTGIToLMI "HF_MODEL_QUANTIZE" "OPTION_QUANTIZE"
+translateTGIToLMI "HF_TASK" "OPTION_TASK"
+translateTGIToLMI "HF_MODEL_TRUST_REMOTE_CODE" "TRUST_REMOTE_CODE"
+translateTGIToLMI "SM_NUM_GPUS" "OPTION_TENSOR_PARALLEL_DEGREE"
```
frankfliu, line 41:
```suggestion
translateTGIToLMI "SM_NUM_GPUS" "TENSOR_PARALLEL_DEGREE"
```

serving/docker/dockerd-entrypoint-with-cuda-compat.sh
frankfliu: I think we should keep HF_MODEL_ID, HF_TASK, AND HF_REVISION in our source code. They are not only used for LLM.

frankfliu: I don't think we should not encourage `OPTION_*` env var usage, it's more of a workaround. Those are per model settings, and should live if `serving.properties`

siddvenk:
Spoke with @frankfliu offline. @lanking520 here's what we discussed:
- we already support HF_MODEL_ID, HF_TASK, HF_REVISION in model server layer. we use those to create a `serving.properties` on the fly, so no change is needed here. I removed those from here
- we have logic using TENSOR_PARALLEL_DEGREE throughout model server that we would need to change if we set to OPTION_TENSOR_PARALLEL_DEGREE. Mapping to TENSOR_PARALLEL_DEGREE let's us reuse that logic now

lanking520 (approved):
agree for now. But I don't feel like we should maintain for two system on parsing envs. Ideally, we should just map everything to serving.properities and use 1 logic to map and maintain them...

siddvenk:
> agree for now. But I don't feel like we should maintain for two system on parsing envs. Ideally, we should just map everything to serving.properities and use 1 logic to map and maintain them...

I agree

frankfliu (approved):

PR #1556, opened 2024-02-21
[docker][TNX] Upgrade to 2.17.0 SDK

tosterberg:
## Description ##

Upgrade TNX container to 2.17.0 SDK


lanking520:
possibly upgrade optimum neuron as well

tosterberg:
> possibly upgrade optimum neuron as well

Optimum will follow separately - optimum changed the way that they save model artifacts and it will be a more involved PR.

PR #1558, opened 2024-02-22
type hints and comments for trtllm handler

ydm-amazon:
## Description ##

Trial commit for type hints and code clarity.

Only works for Python 3.9 and newer - older type hints style that is compatible with 3.8 is deprecated soon. Please let me know if compatibility with 3.8 is a must.

### DIFF
--- a/engines/python/setup/djl_python/tensorrt_llm.py
+++ b/engines/python/setup/djl_python/tensorrt_llm.py
@@ -46,15 +46,13 @@ class TRTLLMService(object):
         """
         Preprocessing function that extracts information from Input objects.
 
-        Args:
-            inputs (Input): a batch of inputs, each corresponding to a new request
+        :param inputs (Input): a batch of inputs, each corresponding to a new request
 
-        Returns:
-            input_data (list[str]): a list of strings, each string being the prompt in a new request
-            input_size (list[int]): a list of ints being the size of each new request
-            parameters (list[dict]): parameters pertaining to each request
-            errors (dict): a dictionary mapping int indices to corresponding error strings if any
-            batch (list): a list of Input objects contained in inputs (each one corresponds to a request)
+        :return input_data (list[str]): a list of strings, each string being the prompt in a new request
+        :return input_size (list[int]): a list of ints being the size of each new request
+        :return parameters (list[dict]): parameters pertaining to each request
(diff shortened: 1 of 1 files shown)


lanking520:
python 3.9 is good enough. The only concern is inside Amazon there are still couple of teams using 3.8

engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -110,9 +141,15 @@ def inference(self, inputs):
 _service = TRTLLMService()
 
 
-def handle(inputs: Input):
+def handle(inputs: Input) -> Optional[Output]:
```
lanking520, line 144: why output is optional? We have to have Output object

ydm-amazon: inputs, theoretically, could be None in an error case

```
    if inputs.is_empty():
        # initialization request
        return None
```

lanking520: None is a special treatment we deal at the back, I don't feel like making Optional is a good idea,

Does

```
Union[None, Output] = Optional[Output]
```
?

ydm-amazon: Yes, the two are the same thing. Optional is just syntactic sugar

ydm-amazon: It's fine to remove the Optional - there won't be an error thrown since I'm most likely doing type checking in another way that doesn't take information directly from type hints.

ydm-amazon:
Added a separate commit so the two commits can be compared.

I changed the format of the comments to be the same as the currently existing format in other parts of djl-python. Previously, I was using pytorch format by default.

lanking520 (approved):

engines/python/setup/djl_python/tensorrt_llm.py
lanking520: @ydm-amazon one good thing of Union is that it support 3.8

ydm-amazon: Oh, what I meant by not supporting 3.8 is choosing to not do

`from typing import List` and using `List` as a type hint

and instead just using `list` as a type hint. This is a feature introduced in 3.9, but the former way is either already deprecated or soon to be deprecated

PR #1560, opened 2024-02-22
[serving] Allows configure log level at runtime

frankfliu:
## Description ##


serving/docs/management_api.md
```diff
@@ -239,3 +239,17 @@ curl "http://localhost:8080/models?limit=2&next_page_token=0"
   ]
 }
 ```
+
+### Configure server logging
+
+`POST /server/logging`
+
+* level - set logging level: TRACE, DEBUG, INFO, WARN, ERROR, FATAL, OFF
```
siddvenk, line 247: should we document that if no level is specified, INFO logging level will be set?

frankfliu: done

siddvenk (approved):

PR #1567, opened 2024-02-23
Stream returns only after putting placeholder finishes

zachgk:
This should hopefully resolve a flaky failure in testAsyncInference()


serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
```diff
@@ -317,27 +317,30 @@ void runJob(
             pending.setCode(202);
             pending.addProperty(X_NEXT_TOKEN, nextToken);
             pending.addProperty(X_CUSTOM_ATTRIBUTES, X_NEXT_TOKEN + '=' + nextToken);
-            cache.put(nextToken, pending);
```
frankfliu, line 320: I didn't realize `cache.put()` returns a Future

frankfliu (approved):

PR #1568, opened 2024-02-23
[docs] add deployment guide section for deploying model to endpoint

siddvenk:
## Description ##

Adds guide for deployment (section 4 in deployment guide).

Doing this a bit out of order, but the remaining pieces in the deployment guide are:
- picking your container and backend
- configurations
- benchmarking

### DIFF
--- a/serving/docs/lmi_new/deployment_guide/deploying-your-endpoint.md
+++ b/serving/docs/lmi_new/deployment_guide/deploying-your-endpoint.md
@@ -1,6 +1,150 @@
 # Deploying your model on a SageMaker Endpoint
 
-TODO
+We recommend using the SageMaker Python SDK to deploy your model on SageMaker. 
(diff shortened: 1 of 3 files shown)


serving/docs/lmi_new/deployment_guide/deploying-your-endpoint.md
```diff
@@ -1,6 +1,140 @@
...
+)
+
+# Make a prediction with your endpoint
+outputs = predictor.predict({
+    "inputs": "The meaning of life is", 
+    "parameters": {"do_sample": True, "max_new_tokens": 256}
+})
+```
+
+## Configuration - environment variables
```
rohithkrn, line 86: Do we recommend one approach vs other between serving.properties and env var? I personally feel env vars is simple and we should put that at the top. (and we should get rid of `SERVING_LOAD_MODELS` env var requirement)

lanking520 (approved):

PR #1570, opened 2024-02-25
[serving] Adds model event listener

frankfliu:
## Description ##

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/util/ModelServerListener.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/util/ModelServerListener.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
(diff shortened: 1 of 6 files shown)


wlm/src/main/java/ai/djl/serving/wlm/util/ModelServerListener.java
```diff
@@ -0,0 +1,92 @@
...
+package ai.djl.serving.wlm.util;
+
+import ai.djl.Device;
+import ai.djl.serving.wlm.Adapter;
+import ai.djl.serving.wlm.ModelInfo;
+
+import java.nio.file.Path;
+
+/** An interface that represent a model server event listener. */
+public interface ModelServerListener {
```
zachgk, line 22: Do we want to create a `ModelServerListenerAdapter` to provide empty definitions (like TrainingListenerAdapter)?

zachgk (approved):

wlm/src/main/java/ai/djl/serving/wlm/util/ModelServerListener.java
frankfliu: Added `ModelServerListenerAdapter`

PR #1574, opened 2024-02-26
[docs] add readme for user guides, remove unused user guides

siddvenk:
## Description ##

Adding a README for the user guides to provide an overview of the information covered. Also deleted unused user guides

### DIFF
--- a/serving/docs/lmi_new/user_guides/README.md
+++ b/serving/docs/lmi_new/user_guides/README.md
@@ -1,0 +1,23 @@
+# LMI Backend User Guides
+
+LMI provides backend specific user guides that cover the following topics:
+
+* Model Artifact Structure 
+  * All backends support standard HuggingFace Transformers Pretrained artifacts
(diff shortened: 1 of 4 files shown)


serving/docs/lmi_new/user_guides/README.md
```diff
@@ -0,0 +1,21 @@
+# LMI Backend User Guides
+
+LMI provides backend specific user guides that cover the following topics:
```
zachgk, line 3: Needs newline

In markdown, lines without a newline separating them are interpreted as being in the same paragraph. So this puts the text line and the bullet points in the same paragraph. Some systems will parse it right regardless, but mkdocs does not

siddvenk: good to know - the markdown preview in github does parse it how i intended. I'll do a pass on the other docs we have added and see if this needs to be fixed elsewhere (separate PR)

lanking520 (approved):

PR #1578, opened 2024-02-28
[docs] add configuration doc to deployment guide

siddvenk:
This PR adds an updated configuration doc to the deployment guide. It is intended to be a comprehensive configuration guide in the context of LMI.

It walks through the important configurations, what they mean, and how to set them.

It also covers the same configurations we had in our previous doc, but in a list form rather than table form. The table form is harder to parse compared to the list form in my opinion, but let me know if you disagree.

It also covers the environment variable translation at the end. Most of the focus is on the `serving.properties` format.

---
This is intended to reflect the state of V8. I know we are working on many improvements to minimized the required set of configs, but I am not confident in recommending auto currently because it can easily fall back to the slow huggingface accelerate backend, and it cannot be used to select vllm.

For the `HF_MODEL_ID` experience, I think that is better served in a separate doc. This doc is intended for advanced users that want a comprehensive overview.


serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+
+If you are deploying a model hosted on the HuggingFace Hub, you must specify the `option.model_id=<hf_hub_model_id>` configuration.
+When using a model directly from the hub, we recommend you also specify the model revision (commit hash) via `option.revision=<commit hash>`.
+Since model artifacts are downloaded at runtime from the Hub, using a specific revision ensures you are using a model compatible with package versions in the runtime environment.
+Open Source model artifacts on the hub are subject to change at any time, and these changes may cause issues when instantiating the model (the model may require a newer version of transformers than what is available in the container). 
+If a model provides custom model (*modeling.py) and custom tokenizer (*tokenizer.py) files, you need to specify `option.trust_remote_code=true` to load and use the model.
+
+If you are deploying a model hosted in S3, `option.model_id=<s3 uri>` should be the s3 object prefix of the model artifacts.
+Alternatively, you can deploy the model artifacts alongside your `serving.properties` configuration file via the `model.tar.gz` artifact and not specify `option.model_id`.
```
lanking520, line 53: for LMI specifically, we do not expect the model to be a `tar.gz`, SageMaker during the runtime will uncompress for us. We can say Sagemaker model object saved into tar.gz. But even today, SM is supporting uncompressed model loading, tar.gz is no longer a requirement

serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+
+LMI expects the following two configurations to determine which backend to use:
+
+* `engine`. The options are `Python` and `MPI`, which dictates how we launch the Python processes
+* `option.rolling_batch`. The available options depend on which container you use:
+  * The DeepSpeed Container provides `auto`, `deepspeed`, `lmi-dist`, `vllm`
+  * The TensorRT-LLM Container provides `auto`, `tensorrtllm`
+  * The Transformers NeuronX Container provides `auto`
+
+We recommend using `engine=MPI` and `option.rolling_batch=auto` unless you want to use a specific backend.
```
lanking520, line 66: MPI + vLLM will not work, we need to make sure we tell the engines of choice when they choose engine

serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+
+* Model Server level configurations. These configurations do not have a prefix (e.g. `job_queue_size`)
+* Engine/Backend level configurations. These configurations have a `option.` prefix (e.g. `option.model_id`)
+
+Since LMI is built using the DJLServing model server, all DJLServing configurations are available in LMI.
+You can find a list of these configurations [here](https://docs.djl.ai/docs/serving/serving/docs/configurations_model.html#python-model-configuration).
+
+The following list of configurations is intended to highlight the relevant configurations for LMI:
+
+* `engine`: The runtime engine of the code. For LMI, the options are `MPI` or `Python`.
```
lanking520, line 93: I would prefer to  keep the table, better for managing

serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+
+We will now cover the components of a minimal starting configuration. This minimal configuration will look like:
+
+```
+# use standard python engine, or mpi aware python engine
+engine=<Python|MPI>
+# where the model artifacts are stored
+option.model_id=<hf_hub_model_id|s3_uri>
+# which inference library to use
+option.rolling_batch=<auto|vllm|lmi-dist|deepspeed|tensorrtllm>
```
ydm-amazon, line 36: For TensorRT-LLM rolling batch, the option should be `trtllm` instead of `tensorrtllm`

serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+# how many gpus to shard the model across with tensor parallelism
+option.tensor_parallelism=<max|number between 1 and number of gpus available>
```
ydm-amazon, line 38: option.tensor_parallelism_degree

serving/docs/lmi_new/deployment_guide/configurations.md
```diff
@@ -1,6 +1,189 @@
...
+SageMaker will extract the contents of the `model.tar.gz` archive to the default directory LMI uses to load and serve models.
+
+## Inference Library Configuration
+
+LMI expects the following two configurations to determine which backend to use:
+
+* `engine`. The options are `Python` and `MPI`, which dictates how we launch the Python processes
+* `option.rolling_batch`. The available options depend on which container you use:
+  * The DeepSpeed Container provides `auto`, `deepspeed`, `lmi-dist`, `vllm`
+  * The TensorRT-LLM Container provides `auto`, `tensorrtllm`
```
ydm-amazon, line 63: Same as in line 36

serving/docs/lmi_new/deployment_guide/configurations.md
siddvenk: Last I tested with `Model` object in sm pysdk, they still hadn't supported uncompressed artifacts. I agree with you, so let me test and validate with pysdk

serving/docs/lmi_new/deployment_guide/configurations.md
siddvenk: I'll provide a static mapping here instead for the supported combinations

serving/docs/lmi_new/deployment_guide/configurations.md
siddvenk: good catch - will update here and elsewhere

serving/docs/lmi_new/deployment_guide/configurations.md
siddvenk: good catch

serving/docs/lmi_new/deployment_guide/configurations.md
siddvenk: I think I'll just remove this portion for now (tar.gz related content). SM PySDK doesn't support uncompressed model loading even though SM platform does

lanking520 (approved):

PR #1593, opened 2024-03-01
[serving] Adds promtheus metrics support

frankfliu:
## Description ##


zachgk:
Is there a reason we can't add the prometheus support as a plugin?

frankfliu:
We can add as a plugin, but the `log4j2.xml` reference to `prometheus` is problematic if the plugin is not there.
The reason add prometheus as a different project is purely because complier flag issue. log4j annotation processor will generate compiler warning and cause error with "-Werr" flag.

prometheus/src/main/java/ai/djl/prometheus/MetricExporter.java
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Writes prometheus metrics to {@code OutputStream}.
+     *
+     * @param os the {@code OutputStream} to write
+     * @param set the filter names
+     * @throws IOException throws if failed to write
+     */
+    public static void export(OutputStream os, Set<String> set) throws IOException {
+        if (!Boolean.parseBoolean(Utils.getEnvOrSystemProperty("SERVING_PROMETHEUS"))) {
+            os.write(
```
zachgk, line 42: Can we make this return an exception and therefore have the HTTP API return a failure error code in this case?

frankfliu: Fixed

zachgk (approved):

PR #1603, opened 2024-03-06
[ci] Fix deprecation error in management-console

tosterberg:
## Description ##

CI build failures pointing to a deprecation in the Charset, updated to use StandardCharsets


tosterberg:
Covered by https://github.com/deepjavalibrary/djl-serving/pull/1601

PR #1604, opened 2024-03-06
Support chat template

xyang16:
## Description ##


engines/python/setup/djl_python/properties_manager/hf_properties.py
```diff
@@ -50,6 +50,7 @@ class HuggingFaceProperties(Properties):
     quantize: Optional[HFQuantizeMethods] = None
     low_cpu_mem_usage: Optional[bool] = False
     disable_flash_attn: Optional[bool] = True
+    apply_chat_template: Optional[bool] = False
```
lanking520, line 53: can we do auto detect?

engines/python/setup/djl_python/huggingface.py
```diff
@@ -481,6 +482,8 @@ def wrap_text_generation_pipeline(self, hf_pipeline):
         def wrapped_pipeline(inputs, *args, **kwargs):
             model = hf_pipeline.model
             tokenizer = hf_pipeline.tokenizer
+            if self.hf_configs.apply_chat_template:
+                inputs = tokenizer.apply_chat_template(inputs, tokenize=False)
```
lanking520, line 486: could we check if request is chat API style and apply chat template automatically?

engines/python/setup/djl_python/properties_manager/hf_properties.py
xyang16: Done

engines/python/setup/djl_python/huggingface.py
xyang16: Done

serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
```diff
@@ -131,14 +133,23 @@ protected void handleRequest(
                                 });
                 break;
             case "invocations":
-                handleInvocations(ctx, req, decoder, null);
+                if (!Boolean.parseBoolean(Utils.getenv("MESSAGES_API_ENABLED"))) {
```
siddvenk, line 136: this shouldn't be `!`, right? if the environment variable is true we should handle as chat completion?

siddvenk:
where do we enforce the actual chat completions schema? Does that still need to be implemented?

serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
```diff
@@ -131,14 +133,23 @@ protected void handleRequest(
                                 });
                 break;
             case "invocations":
-                handleInvocations(ctx, req, decoder, null);
+                if (Boolean.parseBoolean(Utils.getenv("MESSAGES_API_ENABLED"))) {
+                    handleChatCompletions(ctx, req, decoder, null);
```
zachgk, line 137: Can we move these changes to a plugin? This seems like another prediction format for compatibility, so it is better if we keep it external from our core design to keep DJL Serving clean

siddvenk:
Notes from review:
* Support chat across all backends (huggingface handler, trt handler, neuron handler)
* Create a default chat template that matches openai spec (messages object)
  * Apply this template by default
  * Is our fallback reject the request with bad input format, or attempt to use tokenizer default chat template?
* We need the response object format as well
* Ideally let's move the Java code to a plugin

engines/python/setup/djl_python/properties_manager/chat_properties.py
```diff
@@ -0,0 +1,75 @@
+from typing import Optional, Union, List, Dict
+
+from pydantic.v1 import BaseModel, Field, validator, root_validator
+
+
+class ChatProperties(BaseModel):
```
siddvenk, line 6: I think having a strongly typed structure makes sense. I haven't tested this specifically, but does using Pydantic here introduce more overhead (since it's running per request) compared to using just a regular class or dictionary?

I don't think we need to address it in this PR, but we should evaluate other implementations. For this release i think we should mark our chat completions support as experimental.

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -105,6 +106,79 @@ def get_content_type_from_output_formatter(output_formatter: Union[str,
...
+                    "logprob": token.log_prob,
+                    "token": token.text
+                }]
+            },
+            "finish_reason": details.get("finish_reason", None)
+        }
+        response2 = {
+            "choices": [choice2],
+            "usage": {
+                "completion_tokens": details["generated_tokens"]
```
siddvenk, line 144: Can we also add `prompt_tokens` and `total_tokens`?

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -132,6 +206,7 @@ def __init__(
         :param adapter: list of adapters
         :param output_formatter: output formatter function (for example,
             _json_output_formatter, _jsonlines_output_formatter, or user provided function
+        :param output_formatter: output formatter
```
siddvenk, line 209: can remove

engines/python/setup/djl_python/huggingface.py
```diff
@@ -72,6 +74,8 @@
     "QUESTION_ANS": AutoModelForQuestionAnswering,
 }
 
+OPENAI_CHAT_TEMPLATE = "{% for message in messages %}{% if message['role'] == 'user' %}{{ ' ' }}{% endif %}{{ message['content'] }}{% if not loop.last %}{{ '  ' }}{% endif %}{% endfor %}{{ eos_token }}"
```
siddvenk, line 77: I know I suggested using a default template initially, but I think it will cause more issues than solve. I think we should remove this template, and use the tokenizer provided chat template. If no chat template is provided, we should throw an error to the user indicating they must use a model/tokenizer with a chat template

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
xyang16: Done

engines/python/setup/djl_python/huggingface.py
xyang16: OK

lanking520:
you can add more test here: https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/tests/test_test_model.py#L30 just to capture the tokenizer + chat_completion type of messaging without involving GPUs

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
xyang16: Added

engines/python/setup/djl_python/tests/test_rolling_batch.py
```diff
@@ -237,13 +235,10 @@ def get_tokenizer(self):
 
         req = Request(132,
                       "This is a wonderful day",
-                      parameters={
-                          "max_new_tokens": 256,
-                          "details": True
-                      },
+                      parameters={"max_new_tokens": 256},
+                      details=True,
```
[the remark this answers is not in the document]
sindhuvahinis: Yes it would break the UX. So details won't be part of parameters anymore?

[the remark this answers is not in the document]
siddvenk: this doesn't break the CX for the user. We do parse details from parameters, but we're just passing it in separately now to the Request rather than reading it from parameters within the request

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -336,9 +428,13 @@ def get_new_requests(self,
                 params = parameters[i] if i < len(parameters) else {}
                 adapter = adapters[i] if adapters is not None and i < len(
                     parameters) else None
+                details = params.pop("details", False)
```
siddvenk, line 431: @lanking520 @sindhuvahinis this is where details are being parsed from the parameters the user sends

sindhuvahinis: Oh, got it.

lanking520: could we have this UX tested somewhere? https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/tests/test_test_model.py#L30 maybe here?

xyang16: Added test

xyang16: details experience doesn't change, it still can be passed from parameters

siddvenk (approved):

PR #1609, opened 2024-03-07
[Partition] Remove redudant model splitting, Improve Input Model Parsing

a-ys:
## Description ##

This PR includes 2 changes

1. Removed a redundant call to save_pretrained_split in TNXModelLoader partition(). Previously it was called in partition() and then called again in load_inf2_model_from_disk(). We retain only the call in load_inf2_model_from_disk().
     - This was tested with llama-2-7b, reduced partitioning time from 13 min to 9 min (inf2.48xlarge, 6 neuron cores exposed to container, compiler cache populated for both runs)

3. Improved the error handling of the partition PropertiesManager. Previously, this checked if any .bin files existed in the model artifacts, if not, partitioning would fail. Now, we check for .bin OR .safetensors files, allowing users to pass in models that only have .safetensors files without erros.

cc: @tosterberg

### DIFF
--- a/engines/python/setup/djl_python/neuron_utils/model_loader.py
+++ b/engines/python/setup/djl_python/neuron_utils/model_loader.py
@@ -301,7 +301,6 @@ class TNXModelLoader(ModelLoader):
         self.model_config.save_pretrained(save_path)
         self.model = self.load_hf_model()
         self.load_path = self.get_load_path()
-        save_pretrained_split(self.model, self.load_path)
         self.model = self.load_inf2_model_from_disk()
         shutil.copytree(self.load_path, self.split_model_path)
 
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/neuron_utils/model_loader.py
```diff
@@ -301,7 +301,6 @@ def partition(self, save_path: str, **kwargs):
         self.model_config.save_pretrained(save_path)
         self.model = self.load_hf_model()
         self.load_path = self.get_load_path()
-        save_pretrained_split(self.model, self.load_path)
```
tosterberg, line 304: I think we still want to save this in the partition method, lets move it from the self.load_path as the save_pretrained_split, to instead self.split_model_path. I do think this also means that we will need to change our copy behavior on the docker partition side. We want the checkpoint directory to have the split model weights directory, the original model config.json, and the generation_config.json if it exists.

tosterberg: To further qualify this, we want it to be run if we have not already run this in the load_inf2_model_from_disk method.

tosterberg (approved):

PR #1615, opened 2024-03-08
[Build] Pins datasets version to fix pip dependency resolution failure

maaquib:
## Description ##

See #1614


tosterberg (approved):
CI failure for the continuous build is not a blocker, it is just looking for the 0.24 snapshot which doesn't exist anymore. Release pipeline and artifacts should be fine for building the release containers.

PR #1616, opened 2024-03-08
Single model adapter API

zachgk:
This adds a shortcut for the adapter API for single model use case. In addition to being able to call it through `/models/{modelName}/adapters`, it can also be called through `/adapters` if this is unambiguous. This use case may be run into in cases like SageMaker SME.

### DIFF
--- a/serving/docs/adapters_api.md
+++ b/serving/docs/adapters_api.md
@@ -16,7 +16,6 @@ This is an extension of the [Management API](management_api.md) and can be acces
 ### Register an adapter
 
 `POST /models/{modelName}/adapters`
-`POST /adapters` (only for single model use case)
 
 * name - The adapter name.
 * src - The adapter src. It currently requires a file, but eventually an id or URL can be supported depending on the model handler.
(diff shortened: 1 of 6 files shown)


serving/docs/adapters_api.md
```diff
@@ -16,6 +16,7 @@ This is an extension of the [Management API](management_api.md) and can be acces
 ### Register an adapter
 
 `POST /models/{modelName}/adapters`
+`POST /adapters` (only for single model use case)
```
frankfliu, line 19: I don't think we need make SM specific API public

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
```diff
@@ -36,14 +36,16 @@
 public class AdapterManagementRequestHandler extends HttpRequestHandler {
 
     static final Pattern ADAPTERS_PATTERN = Pattern.compile("^/models/[^/^?]+/adapters([/?].*)?");
+    static final Pattern ADAPTERS_SINGLE_PATTERN = Pattern.compile("^/adapters([/?].*)?");
```
frankfliu, line 39: Can you using single regex?

frankfliu (approved):

serving/docs/adapters_api.md
zachgk: I don't like leaving undocumented features, but I moved this into a comment at the bottom of the page where it is more easily ignored

PR #1617, opened 2024-03-08
[lmi] update auto engine logic for vllm and lmi-dist

siddvenk:
## Description ##

This PR introduces a new component `LMIConfigRecommender` that serves to configure LMI models with proper configuration. To start, it will set the rolling_batch, engine, and tp_degree of LMI models.

This is intended to replace the current `inferLMIEngine` implementation. It doesn't do so completely since there is still some logic there for non LMI non TorchServe Python models.

This separates the LMI configuration decisions into a standalone class, which can be used in a plugin/event listener. There are additional things we can configure in the future here like entryPoint to remove the LMI specific logic that exists in some other places


serving/docker/deepspeed.Dockerfile
```diff
@@ -65,6 +65,7 @@ ENV SAFETENSORS_FAST_GPU=1
 ENV NCCL_BLOCKING_WAIT=0
 ENV NCCL_ASYNC_ERROR_HANDLING=1
 ENV TORCH_NCCL_AVOID_RECORD_STREAMS=1
+ENV SERVING_FEATURES=vllm,lmi-dist
```
lanking520, line 68: last time when we introduce this feature, our bring your own `model.py` no longer works. Let's add a small bring your own model.py test and validate it

siddvenk: I'll test with a custom model.py - was the issue that using a custom handler did not work at all due to this environment variable? or was there some other functionality that broke?

siddvenk: I tested with a custom model.py (I copied the huggingface.py handler to a model.py) and it works.

But I also think that this logic here shouldn't be used if user provides a custom model.py. We don't inspect the custom model.py so it's hard to determine whether any of the configurations we set here actually make sense in that context. But since most people providing a custom model.py are doing so for pre/post processing logic (just extending our handler in a sense), I guess this is fine.

frankfliu: Last time, it was a bug in detecting `rolling_batch`, it's not related to `model.py`, it's just a mismatch between `model.py` implementation vs detected rolling_batch mode

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -45,67 +45,101 @@ public final class LmiUtils {
...
             // TODO: Move this from hardcoded to deduced in PyModel
             prop.setProperty("option.entryPoint", "djl_python.stable-diffusion");
-            engineName = "DeepSpeed";
-        } else if (isDeepSpeedRecommended(modelType)) {
-            engineName = "DeepSpeed";
-        } else if (features != null && features.contains("trtllm")) {
             engineName = "MPI";
+        } else if (isTrtLLM(rollingBatch, features)) {
+            engineName = "MPI";
+        } else if ("lmi-dist".equals(rollingBatch) || isLMIDistRecommended(modelType, features)) {
```
lanking520, line 94: regarding to the order. Could we do model specific? Each LMI version will have a recommended model like MPT for LMI-Dist and LLAMA for vLLM?

siddvenk: Yea that makes sense - to start we can have map from model type to backend. Maybe in the future (not sure if this will be inside or outside model server) we can further tune the recommendation based on some other factors like input/output max

siddvenk: updated to a model level mapping.

I do want to change the structure here a bit, since now we're doing more than inferring the engine. Need to think about it a a bit more and can follow up - we're likely going to have more configs we want to auto set upfront if not provided

serving/docker/deepspeed.Dockerfile
siddvenk: I see. If the bug was in detecting rolling batch from the custom model.py, that seems like a model.py bug rather than a potential bug here

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -45,67 +45,98 @@ public final class LmiUtils {
...
+        } else if ("auto".equals(rollingBatch)) {
+            rollingBatch = inferRollingBatchForModel(modelType, features);
+        }
+
+        if ("lmi-dist".equals(rollingBatch)) {
+            engineName = "MPI";
+        } else if ("vllm".equals(rollingBatch)) {
             engineName = "Python";
+            // workaround for vllm TP
+            String tpDegree = prop.getProperty("option.tensor_parallel_degree");
```
lanking520, line 103: option. field should take higher priority than environment variable. Customer could also potentially set `OPTION_TENSOR_PARALLEL_DEGREE` did the ENV parsing happening before this logic @frankfliu ?

lanking520 (approved):
LGTM, just 1 comments

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
siddvenk: yea i have this fix ready in local - I have moved around some things as well to make it a bit cleaner. It will still live in LMIUtils for now, but should be easier to separate into a plugin after this

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.Map;
+import java.util.Properties;
+
+/** A utility class to auto configure LMI model properties. */
+public final class LmiConfigRecommender {
+
+    private static final Map<String, String> MODEL_TO_ROLLING_BATCH =
+            Map.ofEntries(
+                    Map.entry("falcon", "lmi-dist"),
+                    Map.entry("gpt-neox", "lmi-dist"),
```
lanking520, line 27: gpt neox can run on vLLM well

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.Map;
+import java.util.Properties;
+
+/** A utility class to auto configure LMI model properties. */
+public final class LmiConfigRecommender {
+
+    private static final Map<String, String> MODEL_TO_ROLLING_BATCH =
+            Map.ofEntries(
+                    Map.entry("falcon", "lmi-dist"),
```
lanking520, line 26: falcon can also run on vllm well. We can revisit this once Rubikon Engine start to come

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -0,0 +1,102 @@
...
+
+    static void configure(Properties lmiProperties, LmiUtils.HuggingFaceModelConfig modelConfig) {
+        setRollingBatch(lmiProperties, modelConfig);
+        setEngine(lmiProperties);
+        setTensorParallelDegree(lmiProperties);
+    }
+
+    private static void setRollingBatch(
+            Properties lmiProperties, LmiUtils.HuggingFaceModelConfig modelConfig) {
+        String rollingBatch = lmiProperties.getProperty("option.rolling_batch", "auto");
```
lanking520, line 59: what is the current behavior of rolling_batch auto under Python engine? I think it comes to HF Accelerate

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -45,67 +44,42 @@ public final class LmiUtils {
...
-        } else if (isDeepSpeedRecommended(modelType)) {
-            engineName = "DeepSpeed";
-        } else if (features != null && features.contains("trtllm")) {
-            engineName = "MPI";
-        } else {
-            engineName = "Python";
-        }
-        logger.info("Detected engine: {}, modelType: {}", engineName, modelType);
-        return engineName;
+        LmiConfigRecommender.configure(prop, modelConfig);
```
lanking520, line 57: did we ensure we only enter here when customer provide something like option.model_id and we know it is LLM. This is just to prepare for use case on us to support something like

```
HF_MODEL_ID=bge-word-embedding
```

word embedding model doesn't need TP=max, it should be TP1 instead. We can change it in the LMI_config chapter. If model is identified as LLM, then we do TP=max

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
siddvenk: Yes, it uses HF Accelerate with the Scheduler rolling batch.

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
siddvenk: We only enter this recommender if we determine that we are using a "Python" model, and we can fetch the HuggingFace Model Config. While not perfect, that seems like a pretty reasonable requirement to enter this `auto-configure` behavior.

For embedding models, those will also have a HuggingFace Model Config and maybe that contains some information we can use (like architecture or task) to execute separate behavior here. We might also want to do some more analysis here (such as checking model architecture is CausalLM or Seq2SeqLM). We can probably follow up here once we have a better grasp on how to separate generation use-case and embedding use-case

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
siddvenk: Python/auto is also what is used for TNX in that container.

lanking520 (approved):

PR #1624, opened 2024-03-13
[UX] sampling with vllm

sindhuvahinis:
## Description ##

This PR 
- Makes greedy the default for vllm, [before this PR random sampling was the default for vllm]
- Introduces num_beams should enable beam search. [Even for the beam search, the generated sequence will be 1. It will be greater than 1 only if `n` > 1. `n` in equivalent to `num_return_sequences` in huggingface] 
- Makes default value of max_new_tokens to be 30. [Before this PR, it was 16]

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
@@ -86,15 +86,19 @@ class VLLMRollingBatch(RollingBatch):
 
         :return: The same parameters dict, but with VLLM style parameter names.
         """
-        parameters.pop('do_sample', None)
+        parameters["max_tokens"] = parameters.pop("max_new_tokens", 30)
         if "seed" in parameters.keys():
             parameters["seed"] = int(parameters["seed"])
-        if "max_new_tokens" in parameters.keys():
-            parameters["max_tokens"] = parameters.pop("max_new_tokens")
+        if not parameters.pop('do_sample', False):
+            # if temperature is zero, vLLM does greedy sampling
+            parameters['temperature'] = 0
         if "stop_sequences" in parameters.keys():
             parameters["stop"] = parameters.pop("stop_sequences")
         if "ignore_eos_token" in parameters.keys():
             parameters["ignore_eos"] = parameters.pop("ignore_eos")
+        if "num_beams" in parameters.keys():
+            parameters["best_of"] = parameters.pop("num_beams")
+            parameters["use_beam_search"] = True
         return parameters
 
     @stop_on_any_exception
--- a/serving/docs/lmi/user_guides/lmi_input_output_schema.md
+++ b/serving/docs/lmi/user_guides/lmi_input_output_schema.md
@@ -71,7 +71,6 @@ Apart from these common parameters, there are other parameters that are specific
 ```
 DeepSpeedRollingBatchParameters : {
     'typical_p' : float (default= 1.0), 
-    'seed' : integer (default = 0),
     'stop_sequences' : list (default = None),
     'truncate' : integer (default = None),
 }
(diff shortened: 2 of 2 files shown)


serving/docs/lmi/lmi_input_output_schema.md
```diff
@@ -68,8 +70,7 @@ Apart from these common parameters, there are other parameters that are specific
 
 ```
 DeepSpeedRollingBatchParameters : {
-    'typical_p' : float (default= 1.0),
-    'do_sample' : boolean (default = False), 
+    'typical_p' : float (default= 1.0), 
     'seed' : integer (default = 0),
```
siddvenk, line 74: should we remove seed from here since it is marked as a common parameter?

engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
```diff
@@ -86,15 +86,19 @@ def translate_vllm_params(self, parameters: dict) -> dict:
 
         :return: The same parameters dict, but with VLLM style parameter names.
         """
-        parameters.pop('do_sample', None)
+        parameters["max_tokens"] = parameters.pop("max_new_tokens", 30)
```
siddvenk, line 89: why 30? is this the same default for other backends as well?

engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
```diff
@@ -86,15 +86,19 @@ def translate_vllm_params(self, parameters: dict) -> dict:
...
         if "seed" in parameters.keys():
             parameters["seed"] = int(parameters["seed"])
-        if "max_new_tokens" in parameters.keys():
-            parameters["max_tokens"] = parameters.pop("max_new_tokens")
+        if not parameters.pop('do_sample', False):
```
siddvenk, line 92: is do_sample also an alias for sampling with some default temperature in other backends? if so, are the default values for temperature when do_sample is set in parity across backends?

lanking520: +1, we probably want to define the do sample behavior

sindhuvahinis: Yes, do_sample means sampling here. But the default for temperature for greedy in other backends is 1.0.

And vllm seems to do sampling only when temperature is less than 1.0. I guess, they go for the logic that settign temp=0, means disabling the randomness, so they choose greedy when temp is set to 0. 

vLLM chooses Sampling as their default method for choosing next token. But all other frameworks has Greedy as their default. So to unify our handlers behavior, here we are trying to make the sampling method in parity. But yes, default temp value would not be the same as other backends.

serving/docs/lmi/lmi_input_output_schema.md
sindhuvahinis: Done

engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
siddvenk: I see. It can be in another PR, or it might not be feasible at all, but if the default sampling behavior is different across backends when `do_sample=True`, that could cause confusion when users switch between backends. I wonder if it's possible (or woth it) to try to unify the behavior when do_sample=True across all backends (meaning same temperature/other params set to same value)

siddvenk (approved):

engines/python/setup/djl_python/rolling_batch/vllm_rolling_batch.py
sindhuvahinis: Agreed, Will write an internal quip doc for it.

davidthomas426: vllm implements greedy sampling differently internally, but it just uses temperature of 0 as the API instead of adding a separate parameter for it.

* https://github.com/vllm-project/vllm/blob/fb96c1e98c05ffa35dd48416f68e88edb2f9eb34/vllm/sampling_params.py#L235-L236
* https://github.com/vllm-project/vllm/blob/fb96c1e98c05ffa35dd48416f68e88edb2f9eb34/vllm/model_executor/layers/sampler.py#L396-L448

PR #1631, opened 2024-03-18
[tnx] add support for chat completions api

tosterberg:
## Description ##

Adding support for input in the form of chat completion.

Larger possible discussion point. Do we want to do this at the model server api endpoint level `http://127.0.0.1:8080/v1/chat/completions` or similar? Additional consideration, would we want to make middleware for translating input matching some of the other common api's into our inputs + parameters on invocations interface?


tosterberg:
Converting to draft to integrate with  https://github.com/deepjavalibrary/djl-serving/pull/1604.

PR #1632, opened 2024-03-18
[tnx] bump optimum version

tosterberg:
## Description ##

Update optimum to v0.0.20 and separate our generation model code from optimums. This matches the behavior from 0.0.16 optimum and allows us to use the same adapter for TNX and Optimum loaded models.

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_modeling.py
+++ b/engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_modeling.py
@@ -1,0 +1,282 @@
+#!/usr/bin/env python
+#
+# Copyright 2023 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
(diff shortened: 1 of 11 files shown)


engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_modeling.py
```diff
@@ -0,0 +1,284 @@
...
+                 model: torch.nn.Module,
+                 config: "PretrainedConfig",
+                 model_path: Union[str, "Path", "TemporaryDirectory"],
+                 generation_config: Optional[GenerationConfig] = None,
+                 **kwargs) -> None:
+        super().__init__(model, config)
+        self.model_path = model_path
+        self.cur_len = 0
+        self.checkpoint_dir = None
+        self.compiled_dir = None
```
sindhuvahinis, line 60: Where are these checkpoint_dir and compiled_dir used?

tosterberg: Good catch - they are not now that the optimummodelforcausallm doesn't handle saving. I will remove those two attributes.

sindhuvahinis (approved):

PR #1633, opened 2024-03-18
Supports vLLM LoRA adapters

zachgk:
This adds support for vLLM LoRA adapters with the adapters API.


serving/docs/lmi/user_guides/vllm_user_guide.md
```diff
@@ -93,4 +99,9 @@ If you encounter an issue with a `Pass Through` configuration, it is likely an i
...
+| option.enable_loras                      | \>= 0.27.0  | Pass Through	       | This config enables support for LoRA adapters.                                                                                                                                                                                                                                                                                                                                                       | Default: `false`	      |
+| option.max_loras                         | \>= 0.27.0  | Pass Through	       | This config determines the maximum number of LoRA adapters to cache on GPU. All others will be evicted by LRU cache to memory or disk.                                                                                                                                                                                                                                                               | Default: `1`	          |
```
lanking520, line 103: if defaults is 1, eseentially customer are not really doing batch inference right? Since at runtime we need to load each of adapter back to GPU

engines/python/src/main/java/ai/djl/python/engine/PyPredictor.java
```diff
@@ -72,10 +72,11 @@ public List<O> batchPredict(List<I> inputs) throws TranslateException {
             int size = inputs.size();
             if (size == 1) {
                 Output output;
-                if (isRollingBatch) {
-                    output = rollingBatch.addInput((Input) first, timeout);
+                Input input = (Input) first;
+                if (isRollingBatch && !input.getProperties().containsKey("handler")) {
```
lanking520, line 76: trying to understand the logic here. Are we saying we are not supporting customer handler?

engines/python/src/main/java/ai/djl/python/engine/Connection.java
```diff
@@ -362,9 +362,14 @@ protected void encode(ChannelHandlerContext ctx, Input msg, ByteBuf out) {
...
-                ByteBuffer bb = content.valueAt(i).toByteBuffer();
-                out.writeInt(bb.remaining());
-                out.writeBytes(bb);
+                BytesSupplier supplier = content.valueAt(i);
+                if (supplier != null) {
+                    ByteBuffer bb = supplier.toByteBuffer();
+                    out.writeInt(bb.remaining());
+                    out.writeBytes(bb);
+                } else {
+                    out.writeInt(0);
```
lanking520, line 371: trying to know in which scenario we will have suppler to be null

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -267,24 +273,26 @@ def reset(self):
         self.req_id_counter = 0
 
     @abstractmethod
-    def inference(self, input_data, parameters):
+    def inference(self, input_data, parameters, adapters=None):
```
lanking520, line 276: do we have scenarios where some input doesn't need adapter and some needs. How do we get the order of input data with adapters?

serving/docs/lmi/user_guides/vllm_user_guide.md
zachgk: Yeah. 1 was their default, but I changed it to 4 which makes more sense for how we are using it.

tests/integration/llm/client.py
```diff
@@ -831,6 +839,8 @@ def test_handler_rolling_batch(model, model_spec):
     seq_length = 100
     params = {"do_sample": True, "max_new_tokens": seq_length, "details": True}
     req["parameters"] = params
+    if "adapters" in spec:
+        req["adapters"] = spec.get("adapters")[0]
```
siddvenk, line 843: This will use the same adapter for all requests, right? Do we want to test with multiple adapters across concurrent requests?

zachgk: I want to get this merged for now to unblock SAs, but I am planning to have an additional PR with some more testing later

siddvenk: sounds good - i wanted to make sure i wasn't missing something

PR #1640, opened 2024-03-19
add customer required field

lanking520:
## Description ##

Sending all details information to output formatter just for further processing. Making tokenizer reachable from rollingbatch class.

@sindhuvahinis @siddvenk

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/lmi_dist_v2_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/lmi_dist_v2_rolling_batch.py
@@ -85,6 +85,9 @@ class LmiDistRollingBatch(RollingBatch):
         self.request_cache = OrderedDict()
         super().reset()
 
+    def get_tokenizer(self):
+        return self.engine.preprocessor.tokenizer
+
     def translate_lmi_dist_params(self, parameters: dict):
         """
         Helper function to convert DJL Serving parameter names to parameter names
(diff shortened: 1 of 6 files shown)


siddvenk (approved):

engines/python/setup/djl_python/rolling_batch/lmi_dist_v2_rolling_batch.py
```diff
@@ -79,6 +79,9 @@ def reset(self) -> None:
         self.request_cache = OrderedDict()
         super().reset()
 
+    def get_tokenizer(self):
+        return self.engine.tokenizer.tokenizer
```
rohithkrn, line 83: this has to be `self.engine.preprocessor.tokenizer`. `engine` does not have `tokenizer` attribute.

lanking520: @rohithkrn does the same applies to vLLM or Lmi-dist changed the location?

rohithkrn: no, this is specific to `lmi-dist`.

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -163,13 +164,15 @@ def set_next_token(self,
         self.step_token_number = len(
             next_token.id) if next_token.id[0] != -1 else -1
         details = {}
+        # making detailed information captured for each token generation
+        if self.token_cache is not None:
+            details["finish_reason"] = finish_reason
+            details["tokens"] = self.token_cache
+            details["generated_tokens"] = len(self.token_cache)
+            details["input_text"] = self.input_text
```
anusharaghavan, line 172: Can we send the entire payload instead of just input_text. We need to access other input parameters like maxtokens to calculate usage metrics along with input text

PR #1644, opened 2024-03-20
[python] Move output formatter to request level

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -94,6 +94,11 @@ TASK_TO_MODEL = {
     "text2text-generation": AutoModelForSeq2SeqLM
 }
 
+OUTPUT_FORMATTER_TO_CONTENT_TYPE = {
+    "json": "application/json",
+    "jsonlines": "application/jsonlines",
+}
+
 
 class DeepSpeedService(object):
 
@@ -446,12 +451,10 @@ class DeepSpeedService(object):
                     idx += 1
 
                 formatter = parameters[i].get("output_formatter")
-                if formatter == "json":
-                    outputs.add_property(f"batch_{i}_Content-Type",
-                                         "application/json")
-                elif formatter == "jsonlines":
-                    outputs.add_property(f"batch_{i}_Content-Type",
-                                         "application/jsonlines")
+                if formatter in OUTPUT_FORMATTER_TO_CONTENT_TYPE:
+                    outputs.add_property(
+                        f"batch_{i}_Content-Type",
+                        OUTPUT_FORMATTER_TO_CONTENT_TYPE[formatter])
 
             return outputs
         if is_streaming_enabled(self.properties.enable_streaming):
--- a/engines/python/setup/djl_python/deepspeed.py
+++ b/engines/python/setup/djl_python/deepspeed.py
@@ -20,8 +20,6 @@
 # }
 
 import logging
-import os
-import json
 import torch
 from transformers import (AutoConfig, PretrainedConfig, AutoTokenizer,
                           AutoModelForCausalLM, AutoModelForSeq2SeqLM,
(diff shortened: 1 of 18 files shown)


siddvenk:
Some of the logic in parse_input is duplicated across handlers, but I think that is ok for now. I'm looking into consolidating this method into a single one across all handlers, so i can try to refactor that separately.

siddvenk (approved):

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -94,6 +94,11 @@
     "text2text-generation": AutoModelForSeq2SeqLM
 }
 
+OUTPUT_FORMATTER_TO_CONTENT_TYPE = {
```
frankfliu, line 97: Can we move this constant to rolling_batch.py?

engines/python/setup/djl_python/huggingface.py
```diff
@@ -308,12 +313,10 @@ def inference(self, inputs):
...
-                if formatter == "json":
-                    outputs.add_property(f"batch_{i}_Content-Type",
-                                         "application/json")
-                elif formatter == "jsonlines":
-                    outputs.add_property(f"batch_{i}_Content-Type",
-                                         "application/jsonlines")
+                if formatter in OUTPUT_FORMATTER_TO_CONTENT_TYPE:
+                    outputs.add_property(
+                        f"batch_{i}_Content-Type",
+                        OUTPUT_FORMATTER_TO_CONTENT_TYPE[formatter])
```
frankfliu, line 319: The formatter might not be a str type, better use a function

engines/python/setup/djl_python/deepspeed.py
xyang16: Done

engines/python/setup/djl_python/huggingface.py
xyang16: Done

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -128,8 +128,6 @@ def initialize(self, properties: dict):
                 "max_seq_len": int(properties.get("max_tokens", 1024)),
                 "tokenizer": self.tokenizer
             }
-            if "output_formatter" in properties:
```
frankfliu, line 132: Why we remove this? Should the value in `serving.properties` be the default value?

engines/python/setup/djl_python/deepspeed.py
```diff
@@ -441,9 +442,13 @@ def inference(self, inputs: Input):
                                 batch_index=i)
                     idx += 1
 
-            content_type = self.rolling_batch.get_content_type()
-            if content_type:
-                outputs.add_property("content-type", content_type)
+                formatter = parameters[i].get("output_formatter")
+                content_type = get_content_type_from_output_formatter(
```
siddvenk, line 446: could we make content-type a property of the request object that we determine based on the output_formatter passed in, and then return it as part of the result from the inference call?

engines/python/setup/djl_python/deepspeed.py
xyang16: This is setting output_formatter in kwargs to initialize DeepSpeedRollingBatch, but we moved output_formatter out of DeepSpeedRollingBatch to request level. So it's not needed.

engines/python/setup/djl_python/deepspeed.py
xyang16: Yes it's already returned in the response in the property.

frankfliu (approved):

engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -130,6 +128,11 @@ def parse_input(self, inputs):
                         raise ValueError(
                             "In order to enable dynamic batching, all input batches must have the same parameters"
                         )
+
+                if not "output_formatter" in param:
+                    param[
+                        "output_formatter"] = self.properties.output_formatter
```
a-ys, line 134: Hi @xyang16, self.properties is not defined for TransformersNeuronXService here. Apologies if this is already planned and I am missing context.

PR #1651, opened 2024-03-21
[ci] Updating lmi-dist ci tests for rubikon-engine

maaquib:
## Description ##

Following changes have been made
* Removes bitsandbytes test case, since we no longer support it
* Adds speculative decoding test case
* Adds tests for Gemma-7B and Santacoder

- If this change is a backward incompatible change, why must this change be made? - We're dropping support for bitsandbytes quantization
- Interesting edge cases to note here - N/A


.github/workflows/rolling_batch_integration.yml
```diff
@@ -299,14 +299,34 @@ jobs:
             serve -m test=file:/opt/ml/model/test/
           python3 llm/client.py lmi_dist octocoder
           docker rm -f $(docker ps -aq)
-      - name: Test gpt-neox-20b-bitsandbytes
+      - name: Test speculative-llama-13b
         working-directory: tests/integration
         run: |
           rm -rf models
-          python3 llm/prepare.py lmi_dist gpt-neox-20b-bitsandbytes
+          echo -en "SPECULATIVE=true" > docker_env
```
lanking520, line 306: what is this for? This is new to me

lanking520 (commented):
could you also remove the speculative decoding test in vLLM?

.github/workflows/rolling_batch_integration.yml
maaquib: In the current implementation the patch is applied when this env variable is set. @chen3933 to confirm

lanking520 (approved):

serving/docker/deepspeed.Dockerfile
```diff
@@ -20,7 +20,7 @@ ARG torch_vision_version=0.16.2
 ARG pydantic_version=2.6.1
 # HF Deps
 ARG protobuf_version=3.20.3
-ARG transformers_version=4.38.1
+ARG transformers_version=4.39.0
 ARG accelerate_version=0.27.2
```
lanking520, line 24: 0.28.0

maaquib:
Test run: https://github.com/deepjavalibrary/djl-serving/actions/runs/8382943920

davidthomas426 (approved):
Is it ok for our CI tests to use HF model ids directly rather than s3 links?

Other than that, LGTM.

lanking520:
> Is it ok for our CI tests to use HF model ids directly rather than s3 links?
> 
> Other than that, LGTM.

There is a problem with HF id. Firstly it is not very stable. Secondly, sometimes the download speed is very long and cause timeout (flaky tests)

lanking520:
@maaquib could you also just replace the lmi-dist handler with the new one in this PR, so we can just cover in CI testing?

maaquib:
> could you also add tests for vLLM as well, just to serve as baseline?

Added gemma and starcoder2 to vllm

rohithkrn (approved):

tests/integration/llm/client.py
```diff
@@ -373,7 +384,15 @@ def get_model_name():
         "batch_size": [1, 4],
         "seq_length": [256],
         "tokenizer": "mistralai/Mistral-7B-v0.1"
-    }
+    },
+    "llama-7b-unmerged-lora": {
```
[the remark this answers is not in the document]
maaquib: It's should not be blocking. I have not added it to the `.github/workflows/llm_integration.yml` yet. We can do that once the lora plumbing in the engine is done.

maaquib:
Build: https://github.com/deepjavalibrary/djl-serving/actions/runs/8395244232

PR #1657, opened 2024-03-22
[tnx] fix output formatter as param implementation

tosterberg:
## Description ##

Fix for output formatter to params feature for inferentia

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx.py
+++ b/engines/python/setup/djl_python/transformers_neuronx.py
@@ -144,9 +144,9 @@ class TransformersNeuronXService(object):
                             "In order to enable dynamic batching, all input batches must have the same parameters"
                         )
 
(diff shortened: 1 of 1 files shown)


xyang16 (approved):

engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -144,9 +144,9 @@ def parse_input(self, inputs):
                             "In order to enable dynamic batching, all input batches must have the same parameters"
                         )
 
-                if not "output_formatter" in param:
-                    param[
-                        "output_formatter"] = self.properties.output_formatter
+                    if not "output_formatter" in param:
```
xyang16, line 147: It should be unindented one upper level

xyang16 (changes requested):

PR #1660, opened 2024-03-22
[docs][lmi] update landing page sample notebook links

siddvenk:
## Description ##

Adds notebook links for the completed sample notebooks in the gen ai hosting examples repo.

Not all of the models have been completed yet, but for the ones that have I have included the links here. I have removed models from the list that don't have examples so users don't have bad links that don't take them anywhere.


sindhuvahinis (approved):

sindhuvahinis:
https://github.com/aws-samples/sagemaker-genai-hosting-examples - Is this the one we check in our notebooks or djl-demo?

siddvenk:
That repo is going to be housing the "golden" examples. Written by SAs, it will be model focussed.

I think we can continue using djl-demo for notebooks we write, and then share those with SAs to be converted into notebooks in that other repo

PR #1665, opened 2024-03-25
[TRTLLM] Add entrypoint for SM Neo AOT compilation

ethnzhng:
## Description ##

This PR is introducing an additional path in the entrypoint of the TensorRT-LLM LMI container, which will enable the container to be used in the context of [SageMaker Neo](https://aws.amazon.com/sagemaker/neo/) for AOT compilation. 

- This change only adds an alternate AOT partition script which must be specifically enabled during `docker run`, and thus does not impact serving or any default behavior

### DIFF
--- a/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
+++ b/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
@@ -32,6 +32,14 @@ else
     echo "Skip CUDA compat libs setup as package not found"
 fi
 
+if [ -n "$SM_NEO_EXECUTION_CONTEXT" ]; then
+  echo "SageMaker Neo execution context detected"
+  /usr/bin/python3 /opt/djl/partition/sm_neo_trt_llm_partition.py
+  exit_code=$?
+  echo "TensorRT-LLM compilation exited with code $exit_code"
+  exit $exit_code
+fi
+
 # Convert select SM/TGI Environment Variables to LMI Equivalents
 translateTGIToLMI "HF_MODEL_QUANTIZE" "OPTION_QUANTIZE"
 # We use HF_TRUST_REMOTE_CODE in our properties parsing
--- a/serving/docker/partition/sm_neo_trt_llm_partition.py
+++ b/serving/docker/partition/sm_neo_trt_llm_partition.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python
+#
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+import json
+import logging
+import os
+import sys
+import traceback
+
+from utils import load_properties
+from tensorrt_llm_toolkit import create_model_repo
+
+# TODO: Merge the functionality of this file into trt_llm_partition.py
+# so all TRT-LLM partitioning is unified
+
+DJL_SERVING_OPTION_PREFIX = "option."
+
+
+class InputConfiguration(Exception):
+    """Raise when SageMaker Neo interface expectation is not met"""
+
+
+class CompilationFatalError(Exception):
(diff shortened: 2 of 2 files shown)


serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -32,6 +32,14 @@ else
     echo "Skip CUDA compat libs setup as package not found"
 fi
 
+if [ -n "$SM_NEO_EXECUTION_CONTEXT" ]; then
```
lanking520, line 35: So customer only provide an env and by default we enter this path to compile with Neo? Do we want to check what is inside this env vars?

ethnzhng: In practice, customers won't set this env var themselves. It will be set by the Neo compilation host when running the container. We could explicitly require it to be set to "1" if you think that would be preferable

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -0,0 +1,163 @@
...
+        else:
+            raise InputConfiguration(
+                "Neither compiler_flags nor serving.properties found. Please either:"
+                " \nA) specify `compiler_flags` in the CompilerOptions field of SageMaker Neo or CreateCompilationJob API, or"
+                " \nB) include a `serving.properties` file along with your model files."
+                " \nFor info on valid `compiler_flags` fields and values for TensorRT-LLM, see Neo documentation:"
+                " https://docs.aws.amazon.com/sagemaker/latest/dg/neo-troubleshooting.html"
+                " \nFor `serving.properties` configuration, see"
+                " https://docs.djl.ai/docs/serving/serving/docs/lmi/user_guides/trt_llm_user_guide.html"
+                " Note that Neo requires that the `serving.properties` file is placed in the"
```
nskool, line 145: `Neo` -> `SageMaker Neo`?

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -0,0 +1,163 @@
...
+        compilation_error_file) = get_neo_env_vars()
+
+        # Neo requires that serving.properties is in same dir as model files
+        serving_properties = load_properties(input_model_directory)
+        compiler_flags = get_neo_compiler_flags(compiler_options)
+        kwargs = {}
+        
+        if compiler_flags is not None:
+            # If set, prefer Neo CompilerOptions flags over LMI serving.properties
+            logging.info(f"Using CompilerOptions from Neo")
```
nskool, line 123: `Neo` -> `SageMaker Neo` ?

While this log-line mentions `CompilerOptions from Neo` will be used, It will also be useful to explicitly mention that options from serving.properties will be overridden and those from Neo will be used.

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -0,0 +1,163 @@
...
+            verify_neo_compiler_flags(compiler_flags)
+            kwargs = compiler_flags
+        elif len(serving_properties) > 0:
+            # Else, if present, use LMI serving.properties options
+            logging.info(f"Using compiler options from serving.properties file")
+
+            djl_serving_tag = "option."
```
nskool, line 130: Nit: make this a global variable at the beginning of the file for better visibility.

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -0,0 +1,163 @@
...
+import sys
+import traceback
+
+from utils import load_properties
+from tensorrt_llm_toolkit import create_model_repo
+
+# TODO: Merge the functionality of this file into trt_llm_partition.py
+# so all TRT-LLM partitioning is unified
+
+class InputConfiguration(Exception):
```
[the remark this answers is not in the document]
ethnzhng: Neo Service has a preexisting list of specific exception names that they capture & forward, this is one of them and the most fitting

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -0,0 +1,163 @@
...
+        tb_exc = traceback.format_exc()
+        logging.error(f"Failed to write error file: {tb_exc}")
+
+
+def get_neo_env_vars():
+    """
+    Get environment variables required by the Neo interface
+    """
+    try:
+        compiler_options = os.environ.get("COMPILER_OPTIONS")
```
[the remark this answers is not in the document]
ethnzhng: No, `COMPILER_OPTIONS` should always be passed in to the container like the other flags, since this is how the customers specify using TRT-LLM for Neo in the first place. I will change it to match the others

lanking520 (approved):

nskool (approved):

PR #1667, opened 2024-03-25
[lmi][lcnc] fallback to accelerate backend when non text-generation m…

siddvenk:
…odel is provided

## Description ##

Fallback to accelerate for non text-generation model architectures.

From my limited testing, non text-generation architectures are not supported for rolling batch. For these models, we fall back to hf accelerate with dynamic batching.

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
@@ -15,12 +15,17 @@ package ai.djl.serving.wlm;
 import ai.djl.util.Utils;
 import ai.djl.util.cuda.CudaUtils;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 
 /** A utility class to auto configure LMI model properties. */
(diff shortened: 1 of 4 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -46,6 +51,9 @@ public final class LmiConfigRecommender {
                     Map.entry("qwen2", "vllm"),
                     Map.entry("stablelm", "vllm"));
 
+    private static final Set<String> OPTIMIZED_TASK_ARCHITECTURES =
+            Set.of("ForCausalLM", "LMHeadModel", "ForConditionalGeneration");
```
lanking520, line 55: https://huggingface.co/THUDM/chatglm3-6b/blob/main/config.json#L5 this is going to break

siddvenk: ugh, that's good to know. didn't know about this use-case

siddvenk: i've added auto_map to fix this use-case

lanking520 (approved):

PR #1670, opened 2024-03-26
[fix] restrict per request streaming to rolling batch use-cases

siddvenk:
## Description ##

Fix an issue with "stream" being populated for dynamic batch use-cases.

We support per request streaming only for rolling batch (I think it may be possible with dynamic batching, but only in the case where we serve 1 request at a time, so it's probably not worth it).

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx.py
+++ b/engines/python/setup/djl_python/transformers_neuronx.py
@@ -159,7 +159,6 @@ class TransformersNeuronXService(object):
                     if "output_formatter" not in param:
                         param[
                             "output_formatter"] = self.config.output_formatter
-                    param["stream"] = input_map.pop("stream", False)
                     if first or self.rolling_batch:
                         parameters.append(param)
                         first = False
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/transformers_neuronx.py
```diff
@@ -189,6 +188,8 @@ def parse_input(self, inputs):
 
                 if not "output_formatter" in param:
                     param["output_formatter"] = self.config.output_formatter
+                if self.rolling_batch:
+                    param["stream"] = input_map.pop("stream", False)
```
xyang16, line 192: Also needed for trtllm

siddvenk: trtllm is only rolling batch right now, but i know @sindhuvahinis is adding support for dynamic batch there. 

Sindhu, should i add the non rolling batch condition in trtllm as well? Not sure how the changes you will be raising will look, so maybe you can add it in your PR as well?

sindhuvahinis: Yeah will add it in my PR.

xyang16 (approved):

PR #1672, opened 2024-03-27
Increase test timeout for TRT-LLM compilation and tokenizer fix

nskool:
## Description ##

1. The new compilation workflow for TRT-LLM takes longer to compile due to two-step process: convert_checkpoint and 
trtllm-build.
2. Currently setting the retry value such that it increases wait time by 4-5 min. This is a bit generous. And if this addresses some of the timeout issues, will be reduced later to a more reasonable value.
3. Also changes some tokenizers for models to use llama, since awscurl fails for them due to the absence of `tokenizer.json`

### DIFF
--- a/tests/integration/launch_container.sh
+++ b/tests/integration/launch_container.sh
@@ -38,8 +38,7 @@ elif [[ "$platform" == *"inf2"* ]]; then # inf2: pytorch-inf2-24 24 will be the 
   if [[ $devices -gt 1 ]]; then
     is_llm=true
   fi
-  for ((i=0; i<$devices; i++))
-  do
+  for ((i = 0; i < $devices; i++)); do
     host_device+=" --device /dev/neuron${i}"
   done
 fi
@@ -95,6 +94,7 @@ else
     --network="host" \
     ${model_path:+-v ${model_path}:/opt/ml/model:ro} \
     -v ${PWD}/logs:/opt/djl/logs \
+    -v ~/.djl.ai/:/tmp/.djl.ai \
     -v ~/.aws:/home/djl/.aws \
     -v ~/sagemaker_infra/:/opt/ml/.sagemaker_infra/:ro \
     ${env_file} \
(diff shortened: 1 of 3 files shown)


tests/integration/launch_container.sh
```diff
@@ -95,6 +94,7 @@ else
     --network="host" \
     ${model_path:+-v ${model_path}:/opt/ml/model:ro} \
     -v ${PWD}/logs:/opt/djl/logs \
+    -v ~/.djl.ai/:/tmp/.djl.ai \
```
siddvenk, line 97: we can remove this - i don't think we want to use the cache across all tests/model

tests/integration/llm/client.py
```diff
@@ -550,13 +550,13 @@ def get_model_name():
         "max_memory_per_gpu": [22.0],
         "batch_size": [1, 4],
         "seq_length": [256],
-        "tokenizer": "baichuan-inc/Baichuan2-7B-Base"
+        "tokenizer": "TheBloke/Llama-2-13B-fp16"
```
siddvenk, line 553: we can omit this I'd say - without a tokenizer we'll report word level metrics instead of token level. I'd prefer that instead of using a tokenizer that can produce different results

nskool: Sounds good, I've removed it for now. I'm unsure how we use the awscurl results today, but we might want to revisit not using the tokenizer for some models.

tests/integration/launch_container.sh
nskool: Makes sense, removed

siddvenk (approved):

lanking520 (approved):

tests/integration/launch_container.sh
```diff
@@ -112,17 +111,21 @@ set +x
...
-  echo "extra sleep for 2 min on LLM models"
-  total=60
+  total_retries=60
   if [[ "$platform" == *"inf2"* ]]; then
-    total=80
+    total_retries=80
+  fi
+  if [[ "$platform" == *"trtllm"* ]]; then
+    total_retries = 100
+    echo "extra sleep of 5 min for trtllm compilation"
```
lanking520, line 122: this is not correct, we are not sleeping for 5min, instead we will keep retrying for 1500s, and that means 27min

PR #1673, opened 2024-03-27
[CI] Fixes vllm unmerged LoRA tests

zachgk:
Example: https://github.com/deepjavalibrary/djl-serving/actions/runs/8422220967/job/23061144377

The issue was that the download was through a symlink using the huggingface cache, but that cache was not in the container. This removes the symlink usage

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -928,6 +928,7 @@ def write_model_artifacts(properties,
             os.makedirs(os.path.join(adapters_path, adapter_name),
                         exist_ok=True)
             snapshot_download(adapter_id,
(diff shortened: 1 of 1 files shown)


rohithkrn (approved):

tests/integration/llm/prepare.py
```diff
@@ -928,6 +928,7 @@ def write_model_artifacts(properties,
             os.makedirs(os.path.join(adapters_path, adapter_name),
                         exist_ok=True)
             snapshot_download(adapter_id,
+                              local_dir_use_symlinks=False,
```
rohithkrn, line 931: curious why is this an issue, the local path should contain symlink to the cache right?

PR #1674, opened 2024-03-27
support streaming for chat completions, refactor chat parsing to stan…

siddvenk:
## Description ##

This PR does two things:
- separates the chat request parsing into a standalone method
- support streaming output via jsonlines_chat formatter

Tested locally with vllm/lmi-dist/trtllm

Note:
- While chat completions is supported via dynamic batching, the response we return is in our "standard" response format instead of the chat format. That is a separate issue and will be fixed in another PR

### DIFF
--- a/engines/python/setup/djl_python/chat_completions/__init__.py
+++ b/engines/python/setup/djl_python/chat_completions/__init__.py
@@ -1,0 +1,12 @@
+#!/usr/bin/env python
+#
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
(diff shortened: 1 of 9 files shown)


lanking520 (approved):

engines/python/setup/djl_python/chat_completions/__init__.py
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env python
+#
+# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
lanking520, line 3: 2024

PR #1678, opened 2024-03-27
[TrtLLM] Support JIT compilation and dynamic batch for TrtLLM python backend

sindhuvahinis:
## Description ##

T5 handler support is in another PR. Separated my changes into multiple PR for easy reviewing https://github.com/deepjavalibrary/djl-serving/pull/1680

This is for TensorRT-LLM T5 python backend
- Does JIT compilation of T5 models
- Automatically sets dynamic batch if not provided.
- Automatically sets engine=MPI and mpi_mode=True

NOTE: User has to disable rolling_batch in order for us to set dynamic batch and engine=mpi with these changes. We could do disable rolling batch ourselves, but did not do that way because, if TRTLLM release C++ backend support for T5 in between our releases, user could upgrade their TRTLLM in requirements.txt and enable rolling batch and test them. But open to suggestions here.

Testing:
Tested with flan-t5-xl in my EC2 machine.

My serving.properties
```
option.model_id=/opt/ml/model/t5/hf-model
option.rolling_batch=disable
option.entryPoint=djl_python.tensorrt_llm
```

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -78,7 +78,6 @@ public class PyModel extends BaseModel {
                     "Python engine does not support dynamic blocks");
         }
         String entryPoint = null;
-        boolean isTrtLlmBackend = false;
         if (options != null) {
             logger.debug("options in serving.properties for model: {}", modelName);
             for (Map.Entry<String, ?> entry : options.entrySet()) {
@@ -121,9 +120,6 @@ public class PyModel extends BaseModel {
                     case "entryPoint":
                         entryPoint = value;
                         break;
-                    case "rolling_batch":
-                        isTrtLlmBackend = "trtllm".equals(value);
-                        break;
                     case "parallel_loading":
                         parallelLoading = Boolean.parseBoolean(value);
                         break;
@@ -158,6 +154,7 @@ public class PyModel extends BaseModel {
             entryPoint = Utils.getenv("DJL_ENTRY_POINT");
             if (entryPoint == null) {
                 Path modelFile = findModelFile(prefix);
+                String features = Utils.getEnvOrSystemProperty("SERVING_FEATURES");
                 // find default entryPoint
                 String engineName = manager.getEngine().getEngineName();
                 if (modelFile != null) {
@@ -167,7 +164,7 @@ public class PyModel extends BaseModel {
                 } else if ("nc".equals(manager.getDevice().getDeviceType())
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
-                } else if (isTrtLlmBackend) {
+                } else if ("trtllm".equals(features)) {
                     entryPoint = "djl_python.tensorrt_llm";
                 } else if (pyEnv.getInitParameters().containsKey("model_id")) {
                     entryPoint = "djl_python.huggingface";
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
@@ -56,20 +56,27 @@ public final class LmiConfigRecommender {
 
     private LmiConfigRecommender() {}
 
-    static void configure(Properties lmiProperties, LmiUtils.HuggingFaceModelConfig modelConfig) {
-        setRollingBatch(lmiProperties, modelConfig);
-        setEngine(lmiProperties);
+    static void configure(
+            ModelInfo<?, ?> modelInfo,
+            Properties lmiProperties,
+            LmiUtils.HuggingFaceModelConfig modelConfig) {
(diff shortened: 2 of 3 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -56,34 +56,43 @@ public final class LmiConfigRecommender {
...
     }
 
     private static void setRollingBatch(
-            Properties lmiProperties, LmiUtils.HuggingFaceModelConfig modelConfig) {
+            Properties lmiProperties,
+            LmiUtils.HuggingFaceModelConfig modelConfig,
+            String features) {
         // If dynamic batch is enabled, we don't enable rolling batch.
         if (Integer.parseInt(lmiProperties.getProperty("batch_size", "1")) > 1) {
             return;
```
siddvenk, line 76: In setEngine, we are seeing if rolling_batch==disable, do we need to set the value here explicitly to disable? I think i would prefer that well, since we may change the default in the handlers from disable->auto at some point

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -107,6 +123,27 @@ private static void setTensorParallelDegree(Properties lmiProperties) {
         lmiProperties.setProperty("option.tensor_parallel_degree", tpDegree);
     }
 
+    private static void setDynamicBatch(
+            Properties lmiProperties,
+            LmiUtils.HuggingFaceModelConfig modelConfig,
+            ModelInfo<?, ?> modelInfo,
+            String features) {
+        // TODO TrtLLM python backend: Change it once TrtLLM supports T5 with inflight batching.
+        if ("t5".equals(modelConfig.getModelType())
```
siddvenk, line 132: nit: we're using this same condition in a couple places, maybe move to a single method `isT5TrtLlm`

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
sindhuvahinis: Yeah, Right now, in this changes, for t5, we ask the user to explicitly set the rolling batch to disable. So I guess, the UX would be like if the user specifies batch_size then we disable rolling batch automatically. If they don't we try to do rolling batch based on config type. 

But for t5, we enable dynamic batch automatically. In that case, we need to disable rolling batch here?

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
sindhuvahinis: Done

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
siddvenk: In our properties manager, we currently default to disabled rolling batch https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/properties_manager/properties.py#L49.

We have been discussing making the default "auto", so rolling batch is enabled by default. If that were the case, would the logic here still work? For T5, if user doesn't explicitly disable rolling batch, but sets a dynamic batch size, is there a chance both could be enabled? Do we explicitly set rolling_batch=disable somewhere when it's t5 trtllm?

I don't expect us to update this default behavior for this release, so it's not a big deal now. But if we update the default rolling batch value to auto before trt enables rolling batch for t5, could that cause issues down the line?

engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -167,7 +168,7 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
                 } else if ("nc".equals(manager.getDevice().getDeviceType())
                         && pyEnv.getTensorParallelDegree() > 0) {
                     entryPoint = "djl_python.transformers_neuronx";
-                } else if (isTrtLlmBackend) {
+                } else if ("trtllm".equals(features)) {
```
lanking520, line 171: why we cannot use the old param?

sindhuvahinis: Because that param checks whether rolling_batch == trtllm. Might as well, remove the param. 

```
case "rolling_batch":
      isTrtLlmBackend = "trtllm".equals(value);
       break;
 ```    
                    
https://github.com/deepjavalibrary/djl-serving/pull/1678/files#diff-244860c398daaa2ff4bc328b98c867ccce93643b1f65c4f1024bf6bdd5a7beecR124

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
sindhuvahinis: As discussed offline, for JIT (LCNC) experience, automatically disabling rolling batch if its t5 and also setting the entrypoint

lanking520 (approved):

siddvenk (approved):

PR #1680, opened 2024-03-27
[TrtLLM] Python backend support for T5 model

sindhuvahinis:
## Description ##
JIT compilation and others changes in another PR. Separated them for easy review. https://github.com/deepjavalibrary/djl-serving/pull/1678

Changes in this PR:
- Handler support to handle inference with python backend. tensorrt_llm_toolkit recognizes whether it is a python backend based on "use_python_backend" parameter sent with init_inference. 
- If rolling_batch is disabled, then the handler will automatically assume it is python backend and inits the inference. If the model is not supported by python backend, tensorrt_llm_toolkit will throw the error saying model is not supported.

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/properties.py
+++ b/engines/python/setup/djl_python/properties_manager/properties.py
@@ -81,8 +81,9 @@ class Properties(BaseModel):
         batch_size = int(batch_size)
         if batch_size > 1:
             if not is_rolling_batch_enabled(
-                    values['rolling_batch']) and is_streaming_enabled(
-                        values['enable_streaming']):
+                    values.get('rolling_batch', RollingBatchEnum.disable)
+            ) and is_streaming_enabled(
+                    values.get('enable_streaming', StreamingEnum.false)):
                 raise ValueError(
                     "We cannot enable streaming for dynamic batching")
         return batch_size
--- a/engines/python/setup/djl_python/properties_manager/trt_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/trt_properties.py
@@ -10,11 +10,12 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-from djl_python.properties_manager.properties import Properties, RollingBatchEnum, StreamingEnum
+from djl_python.properties_manager.properties import Properties, RollingBatchEnum
 from pydantic.v1 import validator
 
 TRT_SUPPORTED_ROLLING_BATCH_TYPES = [
-    RollingBatchEnum.auto.value, RollingBatchEnum.trtllm.value
+    RollingBatchEnum.auto.value, RollingBatchEnum.trtllm.value,
+    RollingBatchEnum.disable.value
 ]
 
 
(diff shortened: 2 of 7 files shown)


engines/python/setup/djl_python/properties_manager/trt_properties.py
```diff
@@ -24,11 +25,6 @@ class TensorRtLlmProperties(Properties):
     def validate_rolling_batch(cls, rolling_batch: str) -> str:
         rolling_batch = rolling_batch.lower()
 
-        if rolling_batch == RollingBatchEnum.disable.value:
```
siddvenk, line 27: should we keep this in, but allow for t5 model to bypass? for every other model we require rolling batch, right?

engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -163,4 +218,7 @@ def handle(inputs: Input) -> Output:
         # initialization request
         return None
 
-    return _service.inference(inputs)
+    if _service.is_rolling_batch_enabled:
+        return _service.rolling_batch_inference(inputs)
```
siddvenk, line 222: I like this much better, i've wanted to do this for the other handlers as well to clean up the inference methods.

engines/python/setup/djl_python/properties_manager/trt_properties.py
sindhuvahinis: Yeah, but we need to look for a config file, and it can be in huggingface format or engine format, those logic exists in the tensorrt_llm_toolkit. We probably can't do it at the properties pydantic classes, maybe in handler. One thing we can do is to look for trtllm_python_backend property we populate in LmiConfigRecommender maybe.

engines/python/setup/djl_python/properties_manager/properties.py
```diff
@@ -81,8 +81,9 @@ def validate_batch_size(cls, batch_size, values):
         batch_size = int(batch_size)
         if batch_size > 1:
             if not is_rolling_batch_enabled(
-                    values['rolling_batch']) and is_streaming_enabled(
-                        values['enable_streaming']):
+                    values.get('rolling_batch', RollingBatchEnum.disable)
+            ) and is_streaming_enabled(
+                    values.get('enable_streaming', StreamingEnum.false)):
```
lanking520, line 86: should we deprecate enable_streaming?

sindhuvahinis: Yeah. Will do that in a separate PR.

engines/python/setup/djl_python/properties_manager/trt_properties.py
sindhuvahinis: Now checking this and throwing error in handler. (Not in properties)

engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -97,7 +108,37 @@ def parse_input(
...
+            model_config = self._get_config(properties)
+            if model_config.model_type in self.PYTHON_BACKEND_SUPPORTED_MODELS:
+                self.model = tensorrt_llm_toolkit.init_inference(
+                    self.trt_configs.model_id_or_path,
+                    **properties,
+                    use_python_backend=True)
+            else:
+                raise ValueError(
+                    f"You cannot disable rolling batch if its not any of these models"
+                    f" {self.PYTHON_BACKEND_SUPPORTED_MODELS}. Please enable it with auto or tensorrt "
```
nskool, line 138: I think there's no value `tensorrt` for rolling batch - https://github.com/deepjavalibrary/djl-serving/blob/3e948d8307505b3e9b59dd6a1da3143136a5fef3/engines/python/setup/djl_python/properties_manager/properties.py#L19. `trtllm`

sindhuvahinis: Done

lanking520 (approved):

siddvenk (approved):

PR #1683, opened 2024-03-27
[TRTLLM] Add env var for Neo cache

ethnzhng:
## Description ##

This PR is adding an additional environment variable to be recognized by the SageMaker Neo partition script, and passed into the toolkit. It also adds `graphviz` to the TRT-LLM Dockerfile.

### DIFF
--- a/serving/docker/partition/sm_neo_trt_llm_partition.py
+++ b/serving/docker/partition/sm_neo_trt_llm_partition.py
@@ -63,7 +63,8 @@ def get_neo_env_vars():
         raise InputConfiguration(
             f"SageMaker Neo environment variable '{exc.args[0]}' expected but not found"
             f" \nRequired env vars are: 'COMPILER_OPTIONS', 'SM_NEO_INPUT_MODEL_DIR',"
-            f" 'SM_NEO_COMPILED_MODEL_DIR', 'SM_NEO_COMPILATION_ERROR_FILE'")
+            f" 'SM_NEO_COMPILED_MODEL_DIR', 'SM_NEO_COMPILATION_ERROR_FILE', 'SM_NEO_CACHE_DIR'"
+        )
 
 
 def get_neo_compiler_flags(compiler_options):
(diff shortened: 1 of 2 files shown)


serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -156,6 +158,7 @@ def main():
 
         try:
             kwargs["trt_llm_model_repo"] = compiled_model_directory
+            kwargs["cache_root"] = neo_cache_dir
```
nskool, line 161: 'cache_root' is generic, and in the future if the cache_root for neo were to behave differently than another cache_root, it would conflict. 

Can we rename to something that is neo-specific?

Lokiiiiii: How about `neo_cache_root` ? I am open to suggestions !

nskool: It seems `NEO_CACHE_DIR` was chosen as an env variable that the service will set. If we could re-use the same i.e. `neo_cache_dir`, that will be good, but if not `neo_cache_root` seems good enough to me

nskool (approved):

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -64,7 +64,7 @@ RUN mv *.deb djl-serving_all.deb || true
 
 # Install OpenMPI and other deps
 ARG DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && apt-get install -y g++ wget unzip openmpi-bin libopenmpi-dev libffi-dev git-lfs rapidjson-dev && \
+RUN apt-get update && apt-get install -y g++ wget unzip openmpi-bin libopenmpi-dev libffi-dev git-lfs rapidjson-dev graphviz && \
```
nskool, line 67: Nit: why is the apt-version of the package necessary if we are also pip installing it?

Lokiiiiii: It is a system package with a python interface. The 2 need to be installed separately.

lanking520: ok... We can install this

lanking520 (approved):

PR #1688, opened 2024-03-28
[trtllm] support HF_TRUST_REMOTE_CODE in trtllm partition script

siddvenk:
## Description ##

HF_TRUST_REMOTE_CODE does not work for our trt-llm partition use-case. Adding the mapping to the partition script.


serving/docker/partition/trt_llm_partition.py
```diff
@@ -29,6 +29,10 @@ def create_trt_llm_repo(properties, args):
             kwargs[key] = value
     kwargs['trt_llm_model_repo'] = args.trt_llm_model_repo
     kwargs["tensor_parallel_degree"] = args.tensor_parallel_degree
+    trust_remote_code = os.environ.get("HF_TRUST_REMOTE_CODE",
+                                       "false").lower() == "true"
```
sindhuvahinis, line 33: Nit: Maybe we could move this to update_kwargs_with_env_vars.

sindhuvahinis (approved):

serving/docker/partition/trt_llm_partition.py
siddvenk: will do, i didn't even see that method!

PR #1697, opened 2024-03-29
[TRTLLM] T5 python backend log_probs support

sindhuvahinis:
## Description

To support log_probs in TensorRT-LLM in details for T5.

### DIFF
--- a/engines/python/setup/djl_python/tensorrt_llm.py
+++ b/engines/python/setup/djl_python/tensorrt_llm.py
@@ -12,6 +12,7 @@
 # the specific language governing permissions and limitations under the License.
 
 import os
+import torch
 import logging
 import tensorrt_llm_toolkit
 from tensorrt_llm_toolkit.utils import utils as toolkit_utils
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -29,6 +30,68 @@
 from djl_python.properties_manager.properties import is_rolling_batch_enabled
 
 
+def _get_value_based_on_tensor(value, index=None):
```
lanking520, line 33: could we fuse all python backend interaction into a single class (file)? Just trying to keep the handler clean

sindhuvahinis: Sure, we could do that. Same as stable diffusion handlers.

sindhuvahinis: Sure, we could do that. Same as stable diffusion handlers.

engines/python/setup/djl_python/utils.py
```diff
@@ -0,0 +1,61 @@
+import logging
+from djl_python.inputs import Input
+from djl_python.encode_decode import encode, decode
+from djl_python.chat_completions.chat_utils import is_chat_completions_request, parse_chat_completions_request
+
+
+def parse_input(
```
lanking520, line 7: Could we reuse this across handlers?

lanking520 (approved):

PR #1706, opened 2024-03-29
Update lmi-dist docs v9

rohithkrn:
## Description ##

Update lmi-dist docs v9.
Also updated relevant parts of vllm docs


maaquib:
LGTM. Just curious where we want to add `adapter_ids` and `adapter_names` for lora configs

rohithkrn:
> LGTM. Just curious where we want to add `adapter_ids` and `adapter_names` for lora configs

that's a good point. I think they should go into the configuration tables.

lanking520 (approved):

rohithkrn:
synced up with @zachgk. Seems like in v9, we don't actually expose `adapter_ids` and `adapter_names`. They are used only in tests. Instead users will have to package according to - https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docs/adapters.md#adapters-local-directory-recommended.

PR #1709, opened 2024-03-30
[python] Update chat properties

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/chat_completions/chat_properties.py
+++ b/engines/python/setup/djl_python/chat_completions/chat_properties.py
@@ -12,7 +12,7 @@
 # the specific language governing permissions and limitations under the License.
 from typing import Optional, Union, List, Dict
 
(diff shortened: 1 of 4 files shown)


lanking520 (approved):

engines/python/setup/djl_python/chat_completions/chat_properties.py
```diff
@@ -12,7 +12,7 @@
 # the specific language governing permissions and limitations under the License.
 from typing import Optional, Union, List, Dict
 
-from pydantic.v1 import BaseModel, Field, validator, root_validator
+from pydantic import BaseModel, Field, validator, root_validator
```
frankfliu, line 15: Are we ready to upgrade pydantic?

xyang16: @frankfliu Sorry I just saw your comment, we are installing pydantic 2.6 in the docker, I didn't upgrade it.

PR #1714, opened 2024-04-01
[fix] Fix logprobs format in chat completions

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -131,27 +131,26 @@ def _json_chat_output_formatter(token: Token, first_token: bool,
         logprobs = None
         parameters = details.get("parameters", {})
         if parameters.get("logprobs"):
-            logprobs = [
-                {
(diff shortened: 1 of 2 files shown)


siddvenk (approved):

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -131,27 +131,26 @@ def _json_chat_output_formatter(token: Token, first_token: bool,
...
-                } for t in details.get("tokens", [])
-            ]
+            logprobs = {
+                "content": [{
+                        "token":
+                            t.get("text"),
+                        "logprob":
+                            t.get("log_prob"),
+                        "bytes":
+                            (b := [ord(c)
```
siddvenk, line 141: will `ord` ever throw an exception? what happens if there is no utf-8 representation for the character?

xyang16: This is from the output of the model, and detokenized from token_id. I guess will be very unlikely.

PR #1716, opened 2024-04-01
[docs][lmi] update guidance on advanced configurations

siddvenk:
## Description ##

Updating the guidance for our advanced configurations. The feedback from PMs was that our previous documentation didn't help users take action.

The new guidance is as follows:
```
There are two types of advanced configurations: `LMI`, and `Pass Through`.
`LMI` configurations are processed by LMI and translated into configurations that DeepSpeed uses.
`Pass Through` configurations are passed directly to the backend library. These are opaque configurations from the perspective of the model server and LMI.
We recommend that you file an [issue](https://github.com/deepjavalibrary/djl-serving/issues/new?assignees=&labels=bug&projects=&template=bug_report.md&title=) for any issues you encounter with configurations.
For `LMI` configurations, if we determine an issue with the configuration, we will attempt to provide a workaround for the current released version, and attempt to fix the issue for the next release.
For `Pass Through` configurations it is possible that our investigation reveals an issue with the backend library.
In that situation, there is nothing LMI can do until the issue is fixed in the backend library.
```


serving/docs/lmi/deployment_guide/configurations.md
```diff
@@ -108,7 +108,7 @@ The following list of configurations is intended to highlight the relevant confi
...
+| option.dtype                  | The data type you plan to cast the model weights to. If not provided, LMI will use the model's default data type.                                                                                                                                                                                                                                 | `fp16`                                                                                                                                                                       | `fp32`, `fp16`, `bf16` (only on G5/P4/P5 or newer instance types), `int8` (only in lmi-dist)                                                                                                                                                             |
```
lanking520, line 111: we will use fp16 as default

siddvenk: updated

lanking520 (approved):

PR #1718, opened 2024-04-02
[TRTLLM Python backend]Fix the output format for client side batching in dynamic batch 

sindhuvahinis:
## Description ##

If the user sends, client side batch then the result is json array. If the user sends single input, then we send back json object. 

Examples:
Request : `"inputs":["The new movie that got Oscar this year"]`
Response : 
```
[
  {
    "generated_text":"The new movie that got Oscar this year is The Shape of Water."
  }
]
```

Request: `"inputs":"The new movie that got Oscar this year"`
Response:
```
{
  "generated_text":"The new movie that got Oscar this year is The Shape of Water."
}
```

### DIFF
--- a/engines/python/setup/djl_python/tensorrt_llm_python.py
+++ b/engines/python/setup/djl_python/tensorrt_llm_python.py
@@ -10,7 +10,7 @@ from djl_python.properties_manager.trt_properties import TensorRtLlmProperties
 from djl_python.encode_decode import encode
 from djl_python.inputs import Input
 from djl_python.outputs import Output
-from djl_python.utils import parse_input
+from djl_python.utils import parse_input_with_client_batch
 
 
 def _get_value_based_on_tensor(value, index=None):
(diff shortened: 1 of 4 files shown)


lanking520:
do we want to support client side batching? Maybe just server side is good enough

engines/python/setup/djl_python/tensorrt_llm.py
```diff
@@ -50,7 +50,7 @@ def inference(self, inputs: Input) -> Output:
         """
         outputs = Output()
 
-        input_data, input_size, parameters, errors, batch = parse_input(
+        input_data, input_size, parameters, errors, batch, _ = parse_input(
```
frankfliu, line 53: This change break backward compatibility

sindhuvahinis:
> do we want to support client side batching? Maybe just server side is good enough

We already do support. This PR just addresses the output format. In our CI, we also test client side batching. Today's CI would fail because the format is wrong.

engines/python/setup/djl_python/tensorrt_llm.py
sindhuvahinis: Do you mean for people using parse_input in their custom handlers?

sindhuvahinis: As discussed offline, will introduce another function, not to break the backward compatibility of the users.

lanking520 (approved):

PR #1726, opened 2024-04-02
[Neo] Create neo_utils file for shared functions

a-ys:
## Description ##

Separating out some functions that are shared between Neo Neuron and TRT-LLM partition scripts. Neo Neuron partition script is currently WIP.

cc @ethnzhng


serving/docker/partition/neo_utils.py
```diff
@@ -0,0 +1,63 @@
+"""
```
zachgk, line 1: Can you rename this file to be `sm_neo_utils.py` to match the naming convention of `sm_neo_trt_llm_partition.py`

a-ys: Done

zachgk (approved):

PR #1727, opened 2024-04-03
[lmi] always configure lmi model instead of just when engine is missing

siddvenk:
## Description ##

This PR refactors the LMI infer engine into a more general LMI configure method that should run for all LMI models, not just when the engine isn't specified. 

The goal of this PR is to serve as the single area where we can:
- set MPI mode
- recommend rolling batch/backend depending on the model architecture
- set entry point
- set tp degree
- maybe more configs


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -92,25 +93,18 @@ private static void setRollingBatch(
         lmiProperties.setProperty("option.rolling_batch", rollingBatch);
     }
 
-    private static void setEngine(
+    private static void setMPIMode(
```
frankfliu, line 96:
```suggestion
    private static void setMpiMode(
```

frankfliu, line 96: See: https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

siddvenk:
Not really sure what's going on with the windows test, i'll look into it

frankfliu (approved):

PR #1732, opened 2024-04-04
[T5] Add integration test cases

sindhuvahinis:
## Description ##

Added integration test cases for python backend of TRTLLM. For both pre-compiled model and JIT compilation.

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -896,6 +896,12 @@ trtllm_handler_list = {
         "option.max_rolling_batch_size": 32,
         "option.output_formatter": "jsonlines"
     },
+    "flan-t5-xxl": {
+        "engine": "MPI",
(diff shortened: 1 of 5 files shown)


tests/integration/llm/prepare.py
```diff
@@ -896,6 +896,12 @@
         "option.max_rolling_batch_size": 32,
         "option.output_formatter": "jsonlines"
     },
+    "flan-t5-xxl": {
+        "engine": "MPI",
+        "option.model_id": "s3://djl-llm/flan-t5-xxl-trtllm-compiled/",
```
lanking520, line 901: you need to version it, please add 0.8.0 as part of the key. Just like the pre-compiled falcon model

sindhuvahinis: Done

lanking520 (approved):

PR #1744, opened 2024-04-05
[tnx] add safetensor loading for aot compilation

tosterberg:
## Description ##

Add in the support for safetensors loading to the aot partition methods, add a compile only option for just creating neff files, and add model artifacts schema selection to configuration properties.

### DIFF
--- a/engines/python/setup/djl_python/neuron_utils/model_loader.py
+++ b/engines/python/setup/djl_python/neuron_utils/model_loader.py
@@ -18,17 +18,14 @@ import shutil
 import logging
 import tempfile
 import importlib
-from typing import Optional
 from abc import ABC, abstractmethod
 from transformers import AutoModelForCausalLM
 from transformers_neuronx import NeuronAutoModelForCausalLM
 from transformers_neuronx.config import NeuronConfig, QuantizationConfig, ContinuousBatchingConfig
-from djl_python.rolling_batch.neuron_rolling_batch import GenerationStrategy
+from djl_python.properties_manager.tnx_properties import TnXGenerationStrategy, TnXModelSchema
 from transformers_neuronx.module import save_pretrained_split
-from djl_python.neuron_utils.utils import NeuronXModelAdapter
+from djl_python.neuron_utils.utils import NeuronXModelAdapter, get_neuronxcc_version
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/neuron_utils/model_loader.py
```diff
@@ -231,29 +212,31 @@ def load_hf_model(self) -> "PreTrainedModel":
...
+            f"Saving INF2 model to {self.split_model_path} as split model...")
         save_pretrained_split(self.model, self.split_model_path)
 
     def set_model_format(self):
         model_path = os.path.join(os.getcwd(), self.config.model_id_or_path)
         if os.path.isdir(model_path):
-            safetensors_index = os.path.join(model_path,
-                                             "model.safetensors.index.json")
-            self.safetensors_format = os.path.isfile(safetensors_index)
+            self.safetensors_format = any(
```
siddvenk, line 230: good fix - i wonder if we need to do this anywhere else? not for this PR, but would we good to check the other backends too

engines/python/setup/djl_python/properties_manager/tnx_properties.py
```diff
@@ -51,6 +50,18 @@ class TnXModelLoaders(str, Enum):
...
+class TnXModelSchema(str, Enum):
+    legacy = "legacy"
+    optimum = "optimum"
+    safetensors = "safetensors"
+    compile_only = "compile_only"
+
+
+class TnXGenerationStrategy(str, Enum):
+    continuous_batching = "continuous_batching"
+    naive_rolling_batch = "naive_rolling_batch"
```
siddvenk, line 62: what's the difference between these two?

siddvenk (approved):

engines/python/setup/djl_python/properties_manager/tnx_properties.py
tosterberg: Difference is that continuous batching uses the exposed KV cache implementation in transformers neuronx, and the naive implementation recalculates the prefill for all items in the batch every prefill step. It was the old workaround for not having KV cache available, but there are some models that do not support continuous batching so I brought it in as fallback support.

PR #1750, opened 2024-04-08
[lcnc] make default rolling batch size 256 for vllm, lmi-dist

siddvenk:
## Description ##

This is a workaround to set the default rb size for vllm and lmi-dist to 256. 
I would like to capture this in the LmiConfigRecommender.configure api, but that will rely on https://github.com/deepjavalibrary/djl-serving/pull/1727

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
@@ -144,6 +144,26 @@ public final class LmiConfigRecommender {
         }
     }
 
+    static void setRollingBatchSize(Properties lmiProperties) {
+        if (lmiProperties.containsKey("option.max_rolling_batch_size")) {
+            return;
+        }
+        String rollingBatch = lmiProperties.getProperty("option.rolling_batch");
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/deepspeed.py
```diff
@@ -125,7 +125,7 @@ def initialize(self, properties: dict):
                 )
             kwargs = {
                 "max_batch_size":
-                int(properties.get("max_rolling_batch_size", 4)),
+                int(properties.get("max_rolling_batch_size", 32)),
```
maaquib, line 128: This might cause OOMs for existing use-cases since our deepspeed has a 2D KVCache implementation

maaquib (changes requested):

engines/python/setup/djl_python/deepspeed.py
siddvenk: oh interesting, the default rb size on the java size was 32 before, does the python side for ds restrict it further if we set this lower?

siddvenk: reverted this

lanking520 (approved):

maaquib (approved):

PR #1764, opened 2024-04-11
Migrate to pydantic v2

sindhuvahinis:
## Description ##

Migrating to pydantic v2 from v1 for all of our properties manager.


siddvenk:
can you also include the chat properties? it's in a different directory but i think it also uses pydantic v1

sindhuvahinis:
> can you also include the chat properties? it's in a different directory but i think it also uses pydantic v1

Good catch. Included that too.

sindhuvahinis:
This PR is getting very big. Maybe will divide this one/two handlers at a time and raise multiple PRs for easier review.

PR #1772, opened 2024-04-15
[serving] Fixes HF_MODEL_ID conflict with djl:// URL issue

frankfliu:
## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/ModelServer.java
+++ b/serving/src/main/java/ai/djl/serving/ModelServer.java
@@ -65,6 +65,7 @@ import java.nio.file.Paths;
 import java.security.GeneralSecurityException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
 import java.util.Properties;
@@ -331,14 +332,11 @@ public class ModelServer {
         String loadModels = configManager.getLoadModels();
         Path modelStore = configManager.getModelStore();
         if (loadModels == null || loadModels.isEmpty()) {
-            if (modelStore == null) {
-                return;
-            }
             loadModels = "ALL";
         }
 
         ModelManager modelManager = ModelManager.getInstance();
-        List<String> urls = new ArrayList<>();
+        Set<String> urls = new HashSet<>();
         if ("NONE".equalsIgnoreCase(loadModels)) {
             // to disable load all models from model store
             return;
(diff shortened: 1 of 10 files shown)


serving/src/main/java/ai/djl/serving/ModelServer.java
```diff
@@ -512,6 +498,16 @@ private static void printHelp(String msg, Options options) {
     }
 
     private String createHuggingFaceModel(String modelId) throws IOException {
+        if (modelId.startsWith("djl://") || modelId.startsWith("s3://")) {
```
siddvenk, line 501: Couple Questions here:

1. For the s3:// case, this will cause a regression. Our infer entryPoint logic relies on option.model_id being set to infer the huggingface entrypoint. I'd say that's a separate issue though, and should be addressed separately. Not sure if this will cause any issues in our CI (maybe in the LCNC tests)

2. Do we need to support djl://* urls under HF_MODEL_ID? I'm confused about this

frankfliu: 1. `createHuggingFaceModel` is only used at model startup time, the logic here basically make it equivalent to: `djl-serving -m s3://...", the model will be downloaded in `modelDir` and no model_id will be set. When `inferEngine()` is called, `config.json` should already in local folder, and `getHuggingFaceModelConfig()` also looking in `modelDir` first. So this should just work
2. We support text embedding model, and already support `option.model_id=djl://..." in serving.properties. the easiest way to demo our text embedding in Jupyter notebook is to use  `HF_MODEL_ID`

siddvenk: For 1: Right, I understand this but it does mean that this line no longer works for this case https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/src/main/java/ai/djl/python/engine/PyModel.java#L169. The issue is that in PyModel, the logic should be updated and this change will cause us to enter the exception path. I tested this with the PR locally to make sure, and it does cause a "regression" in behavior, but I don't like calling it a regression since the issue is in PyModel infer entry point logic

For 2: Got it, makes sense

frankfliu: Which case will fail?

I'm aware of the follow case may failed to detect entryPoint, only with serving.property, no engine, and no `model.py`:
```
option.model_id=s3://...
```

Even there is `model.py` in S3 bucket, we will still use `huggingface.py`

siddvenk: This case:
```
HF_MODEL_ID=s3://<model>
```

in the lmi container will fail. The fallback logic for entrypoint will not set option.entrypoint=djl_python.huggingface

siddvenk (approved):

PR #1779, opened 2024-04-16
[lmi] infer entry point for lmi models in LmiConfigRecommender

siddvenk:
## Description ##

Adds infer entry point logic to LmiConfigRecommender, and removes the duplicate logic from PyModel.

This will also address a couple of issues:
- this will now infer the correct entrypoint when djl-serving is only provided with raw model files (previously this did not work)
- this will now infer the correct entrypoint when HF_MODEL_ID=<s3_uri>. This did work previously, but some of the previous refactor around HF_MODEL_ID to fix other issues exposed some of the issues in the PyModel entry point logic

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -166,7 +166,8 @@ public class PyModel extends BaseModel {
                     entryPoint = "djl_python.transformers_neuronx";
                 } else if ("trtllm".equals(features)) {
                     entryPoint = "djl_python.tensorrt_llm";
-                } else if (pyEnv.getInitParameters().containsKey("model_id")) {
+                } else if (pyEnv.getInitParameters().containsKey("model_id")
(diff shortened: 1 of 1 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -153,6 +155,26 @@ private static void setRollingBatchSize(Properties lmiProperties) {
                 "option.max_rolling_batch_size", String.valueOf(rollingBatchSize));
     }
 
+    private static void setEntryPoint(
```
frankfliu, line 158: If there is `model.py` or $model_name.py, we should pick it up, set a `entryPoint` here will disable custom model use case.

siddvenk: oh right, if .py file exists in model dir but it's not specified in the option. will fix

frankfliu (approved):

PR #1780, opened 2024-04-16
[0.27.0-DLC] update user guide for TRTLLM

lanking520:
## Description ##

Update userguide for TRTLLM


serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
```diff
@@ -0,0 +1,125 @@
+# TensorRT-LLM finding max_num_tokens tutorial
+
+From TRTLLM 0.8.0, we introduced `max_num_tokens` build option to simplify the compilation of TRTLLM model as well as supporting all kinds of traffic pattern.
```
frankfliu, line 3: We should use LMI 0.27.0, it's hard for user to figure the version mapping.

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
```diff
@@ -0,0 +1,125 @@
...
+From LMI 0.27.0 (TRTLLM 0.8.0), we introduced `max_num_tokens` build option to simplify the compilation of TRTLLM model as well as supporting all kinds of traffic pattern.
+This max number of tokens represent the shared tokens across the TRTLLM framework regardless of the batch size or context length.
+It is similar to pagedAttention tokens (e.g max_batched_tokens) settings in vLLM. If you set this value in LMI, it is equivalent to the followings:
+
+```
+Serving prperties: option.max_num_tokens=50000
+ENV: OPTION_MAX_NUM_TOKENS=50000
+```
+
+It is the same as:
```
siddvenk, line 12: I feel like this is a bit confusing, since 4096 * 256 > 50000. They're not really the same

I think we should call out that max_input_len, max_output_len, and max_rolling_batch_size are now used as client controls (limits the types of requests clients can send, and the maximum number in the batch). Without any of these restrictions specified, we basically allow any composition of requests as long as we do not exceed 50000 tokens total across all requests in a single batch

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
```diff
@@ -0,0 +1,125 @@
...
+
+You can override `model_id` to
+- local path like `/opt/djl/large_store/<my_downloaded_model_folder>`
+- HF model id like `TinyLlama/TinyLlama-1.1B-Chat-v1.0`
+
+`tensor_parallel_degree` means the number of gpu you want to use to parallel computation.
+The larger value you set, the bigger max_token the model will be able to support. Finally, just run with:
+
+```
+python3 finder.py
```
siddvenk, line 120: Not for this PR, but have we considered including this as an optional step in the compilation process?

It would be great if we could trigger this automatically (if user requests, or we could enable by default) as part of the compilation process

siddvenk:
We need to revisit this doc as well https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docs/lmi/deployment_guide/backend-selection.md.

It was written for v8 container where the differences in traffic patterns is used to recommend containers. Not needed for this PR, but needs to be addressed as well.

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
siddvenk: If user does not specify `max_num_tokens`:
- do we automatically set `max_num_tokens` to `batch_size * (max_input_len + max_output_len)`? I assume we also have default values for each of those in case user does not specify?

If user only specifies `max_num_tokens`, do we have defaults for batch_size, input/output len as well that would constrain the types of requests clients can send?

ydm-amazon: Good question, if user does not specify max_num_tokens, max_num_tokens is not automatically set - we then use max_batch_size, input len, and output len. Those have defaults - I believe it is batch size 16, input 1024, output 512. That would definitely set limits on the input and output length of the request

lanking520: If user not specify max_num_tokens, we simply not setting them and fallback to the old ways. User will going to suffer from the input/output and batch size limitation. I think this is a good call out, we can try to make a max_num_tokens calculated if customer specify the remaining three. So we can always support batch size 256

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
```diff
@@ -0,0 +1,125 @@
...
+
+Before finding the precise number, you can simply try the following formula by yourself to estimate:
+
+```
+max_num_tokens = (avg_input + avg_output) * avg_concurrency
+```
+And then try this number with LMI solution and see if it works. 
+If not working, go with a binary search approach by reducing halves until you get the number you want.
+
+We also created a simple function to allow you do this with LMI container.
```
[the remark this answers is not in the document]
lanking520: SageMaker Notebook, they can also initiate a container in

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
lanking520: This takes some time to finish, so that's why we didn't automate this

serving/docs/lmi/tutorials/trtllm_finding_max_num_tokens_tutorial.md
```diff
@@ -0,0 +1,132 @@
...
+ENV: OPTION_MAX_NUM_TOKENS=50000
+```
+
+It is the same as:
+
+```
+option.max_num_tokens=50000
+option.max_input_len=4096
+option.max_output_len=4096
+option.max_rolling_batch_size=256
```
siddvenk, line 20: I think we need to call out that these configs are independent of the max_num_token configs. 

These configs are purely used for validation/generation limits/batch size restriction, and do not mean the server can handle 256 batch with 4096 total tokens each

lanking520: addressed with a comments below

siddvenk (approved):

ydm-amazon (approved):

PR #1781, opened 2024-04-16
[serving] Refactor rolling batch detection logic

frankfliu:
## Description ##

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
@@ -12,6 +12,7 @@
  */
 package ai.djl.serving.wlm;
 
+import ai.djl.util.Ec2Utils;
 import ai.djl.util.Utils;
 import ai.djl.util.cuda.CudaUtils;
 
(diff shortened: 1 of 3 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -80,15 +81,24 @@ private static void setRollingBatch(
...
         } else if (!isTextGenerationModel(modelConfig)) {
             // Non text-generation use-cases are not compatible with rolling batch
             rollingBatch = "disable";
-        } else if (isVllmEnabled(features) && isLmiDistEnabled(features)) {
-            rollingBatch = MODEL_TO_ROLLING_BATCH.getOrDefault(modelConfig.getModelType(), "auto");
-        } else if (LmiUtils.isTrtLlmRollingBatch(lmiProperties)) {
+        } else if (isLmiDistEnabled(features)
+                && "lmi-dist".equals(MODEL_TO_ROLLING_BATCH.get(modelType))) {
+            rollingBatch = "lmi-dist";
+        } else if (isLmiDistEnabled(features)
```
siddvenk, line 93: `isVllmEnabled`

wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -80,15 +81,24 @@ private static void setRollingBatch(
...
+                && "vllm".equals(MODEL_TO_ROLLING_BATCH.get(modelType))) {
+            rollingBatch = "vllm";
+        } else if (isTrtLlmEnabled(features)) {
             rollingBatch = "trtllm";
+        } else if (Ec2Utils.isSageMaker()) {
```
lanking520, line 98: Why sagemaker makes a difference here?

siddvenk: This will break the neuron case

frankfliu: I added vllm,lmi-dist to neuron docker features, "vllm" or "lmi-dist" will be used as rolling batch in neuron case.

siddvenk: This will still break until we update the handler to support those https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/properties_manager/tnx_properties.py#L77. 

But, I'm confused why vllm and lmi-dist are added as features in the neuron container. Will vllm and lmi-dist map to the tnx rolling batch implementation in neuronx? Why not just expose `SERVING_FEATURES=tnx` in neuron, and check for `tnx` here to enable rolling batch in neuron?

Is this because we'll have vllm support for neuronx, and since lmi-dist is compatible with lmi-dist this should just work?

siddvenk (approved):

PR #1788, opened 2024-04-18
Refactor parse_input for handlers

sindhuvahinis:
## Description ##

- DeepSpeed handler is not refactored, since it is going to be deprecated. 

Tested vllm, huggingface and tensorrt llm handlers with gpt2 model.

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -23,7 +23,7 @@ from transformers import (
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
 from peft import PeftConfig, PeftModel, PeftModelForCausalLM
 
-from djl_python.encode_decode import encode, decode
+from djl_python.encode_decode import encode
 from djl_python.inputs import Input
 from djl_python.outputs import Output
 from djl_python.streaming_utils import StreamingUtils
@@ -31,7 +31,7 @@ from djl_python.rolling_batch.rolling_batch import get_content_type_from_output_
 
 from djl_python.properties_manager.properties import StreamingEnum, is_rolling_batch_enabled, is_streaming_enabled
 from djl_python.properties_manager.hf_properties import HuggingFaceProperties
-from djl_python.chat_completions.chat_utils import is_chat_completions_request, parse_chat_completions_request
+from djl_python.utils import parse_input_with_formatter, InputFormatConfigs
 
 ARCHITECTURES_2_TASK = {
     "TapasForQuestionAnswering": "table-question-answering",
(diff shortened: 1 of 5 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -208,75 +208,20 @@ def load_stopping_criteria_list(self, stop_sequence):
...
-
-        self.adapters = adapters if found_adapters else None
-
-        return input_data, input_size, parameters, errors, batch
+    def parse_input(
+            self, inputs: Input
+    ) -> tuple[list[str], list[int], list[dict], dict, list]:
+        is_rolling_batch = is_rolling_batch_enabled(
+            self.hf_configs.rolling_batch)
+        parsed_input = parse_input_with_client_batch(
```
siddvenk, line 216: I wonder if we can do this during initialization rather than per request.

We should know at the initialization phase whether we're using dynamic batch or rolling batch, and can have separate methods/objects like DynamicBatchInputParser, RollingBatchInputParser for example.

I also wonder if we should change the signature to return the ParsedInput object here? That could be useful for users trying to override this implementation, where they don't have to worry about which components of ParseInput to return and that interface stays relatively consistent for them over time

engines/python/setup/djl_python/huggingface.py
```diff
@@ -208,81 +213,32 @@ def load_stopping_criteria_list(self, stop_sequence):
...
-            if not isinstance(adapters_per_item, list):
-                adapters_per_item = [adapters_per_item]
-
-            if adapters_per_item:
-                found_adapters = True
-            else:
-                # inference with just base model.
-                adapters_per_item = [""] * len(_inputs)
+    def parse_input(
+        self, inputs: Input, tokenizer, output_formatter
```
siddvenk, line 217: where are tokenizer and output_formatter used in this method?

engines/python/setup/djl_python/utils.py
```diff
@@ -13,24 +15,37 @@ class ParsedInput:
     errors: dict
     batch: list
     is_client_side_batch: list = field(default_factory=lambda: [])
+    adapters: list = None
+    found_adapters: bool = False
+
+
+@dataclass
+class InputFormatter:
```
siddvenk, line 23: Is the plan to extend the functionality of this class to do formatting? If not, I think we should change the name to something like InputFormatConfigs or RequestFormatConfigs

sindhuvahinis: I am not clear of the use case to extend this yet to have multiple formats yet. I like InputFormatConfigs, will change to this.

engines/python/setup/djl_python/huggingface.py
sindhuvahinis: These two are, now not needed in all the handlers. Used the same signature for all the handlers for backward compatibility.

engines/python/setup/djl_python/utils.py
sindhuvahinis: Done

siddvenk (approved):

PR #1798, opened 2024-04-22
[fix] match serving option with properties for draft model download

tosterberg:
## Description ##

Fix the download option consumption for s3 downloads for draft models.

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
@@ -1130,11 +1130,11 @@ public final class ModelInfo<I, O> extends WorkerPoolConfig<I, O> {
     }
 
     private void downloadDraftModel() throws ModelException, IOException {
-        String draftModelId = prop.getProperty("option.speculative_draft_model");
+        String draftModelId = prop.getProperty("option.draft_model_id");
         if (draftModelId != null && draftModelId.startsWith("s3://")) {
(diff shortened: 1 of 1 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -1130,11 +1130,11 @@ private void runS3cmd(String src, String dest) throws ModelException {
     }
 
     private void downloadDraftModel() throws ModelException, IOException {
-        String draftModelId = prop.getProperty("option.speculative_draft_model");
```
lanking520, line 1133: We should not rename this, since it breaks the current experience os SD

lanking520 (changes requested):
Please check internally with the team if you plan to rename a parameter

wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
tosterberg: Sorry, I meant to raise this in draft. Current speculative decoding experience through the handler won't download an s3 model. I brought this up for discussion, we can either rename if this hasn't been used yet (https://github.com/deepjavalibrary/djl-serving/pull/1693), or we add functionality that follows the properties naming for `option.draft_model_id`.

tosterberg: I found it in `lmi_dist`, missed it the first time, so I am going to close this and rethink how to unify the CX for these properties and make sure they function the same.

PR #1801, opened 2024-04-22
set user-agent for hf requests

siddvenk:
## Description ##

set user agent for hf requests


wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -176,6 +176,7 @@ public static URI generateHuggingFaceConfigUri(ModelInfo<?, ?> modelInfo, String
             String hfToken = Utils.getenv("HF_TOKEN");
             configUri = URI.create("https://huggingface.co/" + modelId + "/raw/main/config.json");
             HttpURLConnection configUrl = (HttpURLConnection) configUri.toURL().openConnection();
+            configUrl.setRequestProperty("User-Agent", "curl/7.8.6");
```
frankfliu, line 179: We wait for HF response, let defer this change.

PR #1803, opened 2024-04-23
[trt][lmi] always set mpi mode for trt container

siddvenk:
## Description ##


sindhuvahinis (approved):

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -54,17 +54,22 @@ private LmiUtils() {}
 
     static void configureLmiModel(ModelInfo<?, ?> modelInfo) throws ModelException {
         HuggingFaceModelConfig modelConfig = getHuggingFaceModelConfig(modelInfo);
+        Properties prop = modelInfo.getProperties();
         if (modelConfig == null) {
-            // Not a LMI model
+            // Precompiled models may not have config, set mpi mode when trtllm
+            // TODO: Include TRT/Neuron compiled models in this configure flow
```
frankfliu, line 60: We should save our own metadata file for precompiled model when we compile it.

lanking520 (commented):
Ignore my comments, we are not really setting MPI in the tests

PR #1813, opened 2024-04-24
replace peft with open source version

lanking520:
## Description ##

update peft wheel


rohithkrn:
We should make sure to add in the release notes that this is a backward incompatible change.

rohithkrn (approved):

PR #1819, opened 2024-04-26
[docs][lmi] update backend selection guide to recommend lmi-dist over…

siddvenk:
… vllm

## Description ##

Update the container selection document to recommend lmi-dist over vllm. Also updates the image.

This change is really just a replacement of lmi-dist wherever vllm was. I think we need to reevaluate when we recommend lmi-dist and trtllm since the traffic pattern evaluation does not seem as relevant anymore.


lanking520 (approved):
please also backport to 0.27.0 branch

PR #1826, opened 2024-04-26
[tnx] version bump Neuron SDK and Optimum

tosterberg:
## Description ##

Bump versions for Neuron SDK and optimum-neuron. Additionally fixed a pip install bug in the dockerfile, and added NO_OMP_NUM_THREADS=true for pytorch based multithreading.

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/Connection.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/Connection.java
@@ -186,6 +186,10 @@ class Connection {
             // TODO: re-map logic device once neuron fixed bug
             pyEnv.addEnv("NEURON_RT_VISIBLE_CORES", visibleCores);
             logger.info("Set NEURON_RT_VISIBLE_CORES={}", visibleCores);
+
+            String neuronThreads = getNeuronThreads(tensorParallelDegree);
+            pyEnv.addEnv("OMP_NUM_THREADS", neuronThreads);
+            logger.info("Set OMP_NUM_THREADS={}", neuronThreads);
         }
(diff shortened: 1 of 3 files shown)


engines/python/src/main/java/ai/djl/python/engine/Connection.java
```diff
@@ -186,6 +186,10 @@ static String[] getPythonStartCmd(PyEnv pyEnv, Model model, int workerId, int po
             // TODO: re-map logic device once neuron fixed bug
             pyEnv.addEnv("NEURON_RT_VISIBLE_CORES", visibleCores);
             logger.info("Set NEURON_RT_VISIBLE_CORES={}", visibleCores);
+
+            String neuronThreads = getNeuronThreadsPerCore(tensorParallelDegree);
+            pyEnv.addEnv("OMP_NUM_THREADS", neuronThreads);
```
frankfliu, line 191: Add a comment to the neuron doc about OMP_NUM_THREADS

tosterberg: I was able to get the `NO_OMP_NUM_THREADS=true` to work so I went with that. I did however add a section to the tnx docs referencing the TNX release notes around possible race conditions when using MME configs. I added a note on how to solve that problem if it is an issue.

tosterberg: Updated after our discussions - I double checked the inf2 cpu cores and we can safely set the threads to be 2x the tensor parallel degree since the inf2.xlarge has 2 neuron cores and 4 cpu cores.

frankfliu (approved):

PR #1841, opened 2024-04-29
[CI] Fix test LLM client

zachgk:
## Description ##

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -1165,4 +1165,4 @@ def run(raw_args):
 
 
 if __name__ == "__main__":
-    run(sys.argv)
+    run(sys.argv[1:])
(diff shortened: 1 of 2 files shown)


nskool (approved):

tests/integration/llm/client.py
```diff
@@ -1165,4 +1165,4 @@ def run(raw_args):
 
 
 if __name__ == "__main__":
-    run(sys.argv)
+    run(sys.argv[1:])
```
lanking520, line 1168: does this work with

```
parser.parse_args(args=raw_args)
```
?

zachgk: That is what it is fixing. The sys.argv includes the executable `client.py` which is being removed here. Passing it to `parse_args` will have it treat the `client.py` as the first argument

PR #1842, opened 2024-04-30
Stop the model server when model download fails

sindhuvahinis:
## Description ##

If the error occurs at model initialization is failed, we handle the exception at CompletableFuture `exceptionally`, to wait for 3 seconds and then stop the model server. Then we before even checking whether model server is stopped or not, we bind to the 8080 address, and then we check for model server is stopped or not. 

This causes problems with SageMaker, since it receives the ping HTTP Status as 200, they mark the endpoint as `InService`.

### DIFF
--- a/serving/src/main/java/ai/djl/serving/ModelServer.java
+++ b/serving/src/main/java/ai/djl/serving/ModelServer.java
@@ -459,6 +459,9 @@ public class ModelServer {
                                     });
             if (configManager.waitModelLoading()) {
                 f.join();
+                if (stopped.get()) {
+                    throw new BadWorkflowException("Model server is stopped.");
+                }
             }
             startupModels.add(modelName);
         }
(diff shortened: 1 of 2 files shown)


serving/src/main/java/ai/djl/serving/ModelServer.java
```diff
@@ -459,6 +459,9 @@ private void initModelStore() throws IOException, BadWorkflowException {
                                     });
             if (configManager.waitModelLoading()) {
                 f.join();
+                if (stopped.get()) {
```
frankfliu, line 462: We actually don't need this.

We can do:
```
            CompletableFuture<Void> f = modelManager.registerWorkflow(workflow);
            f.exceptionally(...)
```

sindhuvahinis: Done.

lanking520 (approved):

frankfliu (approved):

PR #1861, opened 2024-05-02
[vllm, lmi-dist] Support input log_probs

sindhuvahinis:
## Description ##

Support input log_probs in our API. 

If user sends `"decoder_input_details": true` and `details: true`, prefill in details is sent. 
If user sends `decoder_input_details:true` and `details: false`, prefill in details is still sent, but other values would be null. 

Testing:
Tested with vllm gpt2 model, with both  json and jsonline output formatter.

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -71,8 +71,12 @@ def _json_output_formatter(token: Token, first_token: bool, last_token: bool,
                 "finish_reason": details.get("finish_reason", None),
                 "generated_tokens": details.get("generated_tokens", None),
                 "inputs": details.get("inputs", None),
-                "tokens": details.get("tokens", None)
+                "tokens": details.get("tokens", None),
             }
+
+            if "prompt_tokens_details" in details:
+                final_dict["prefill"] = details.get("prompt_tokens_details")
+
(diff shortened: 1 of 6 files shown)


engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -97,8 +101,11 @@ def _jsonlines_output_formatter(token: Token, first_token: bool,
             final_dict["details"] = {
                 "finish_reason": details.get("finish_reason", None),
                 "generated_tokens": details.get("generated_tokens", None),
-                "inputs": details.get("inputs", None)
+                "inputs": details.get("inputs", None),
             }
+            if "prompt_tokens_details" in details:
```
lanking520, line 106: Is this following TGI streaming use case?

sindhuvahinis: Tried TGI container 2.0 

```
data:{"error":"Input validation error: `decoder_input_details` == true is not supported when streaming tokens","error_type":"validation"}
```

lanking520: ok, then we can define the way we like

lanking520 (approved):

lanking520 (commented):
Could we also update our API schema documentation?

sindhuvahinis:
> Could we also update our API schema documentation?

Done

PR #1866, opened 2024-05-03
add TGI compat feature for rollingbatch

lanking520:
## Description ##

Add TGI compatibility support

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -12,11 +12,13 @@
 # the specific language governing permissions and limitations under the License.
 import json
 import logging
+import os
 import time
 from abc import ABC, abstractmethod
 from typing import List, Union, List, Callable, Optional
 
 FINISH_REASON_MAPPER = ["length", "eos_token", "stop_sequence"]
+TGI_COMPAT = False
 
 
 class Token(object):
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/tests/test_test_model.py
```diff
@@ -54,7 +55,8 @@ def test_with_env(self):
         envs = {
             "OPTION_MODEL_ID": "NousResearch/Nous-Hermes-Llama2-13b",
             "SERVING_LOAD_MODELS": "test::MPI=/opt/ml/model",
-            "OPTION_ROLLING_BATCH": "auto"
+            "OPTION_ROLLING_BATCH": "auto",
+            "TGI_COMPAT": "true"
```
frankfliu, line 59: Can we use SERVING_TGI_COMPAT?

lanking520: Why SERVING?

frankfliu: This should be a global server configuration and can be configure in `config.properties`

sindhuvahinis:
If the env name is finalized, every thing else, looks good to me.

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -445,6 +451,8 @@ def __init__(self, **kwargs):
         self.waiting_steps = kwargs.get("waiting_steps", None)
         self.current_step = 0
         self.default_output_formatter = kwargs.get("output_formatter", None)
+        global TGI_COMPAT
+        TGI_COMPAT = kwargs.get("tgi_compat", "false").lower() == 'true'
```
frankfliu, line 455:
```suggestion
        self.tgi_compat = kwargs.get("tgi_compat", "false").lower() == 'true'
```

lanking520: There is no way to pass this value back to the serializer. Unless we make them to be part of the Request class

frankfliu (approved):

PR #1870, opened 2024-05-03
[T5] TRTLLM python repo model check

sindhuvahinis:
## Description ##

To check whether model is already converted or not for TRTLLM python repo. In TrtLLM 0.10.0, Triton support for T5 is getting released. Once it is released, we will be removing TRTLLM python support in our handlers and toolkit.

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -594,7 +594,12 @@ trtllm_handler_list = {
         "engine": "MPI",
         "option.model_id": "s3://djl-llm/flan-t5-xxl-trtllm-compiled/v0.8.0/",
         "option.rolling_batch": "disable",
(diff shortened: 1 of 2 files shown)


tests/integration/llm/prepare.py
```diff
@@ -594,7 +594,12 @@
         "engine": "MPI",
         "option.model_id": "s3://djl-llm/flan-t5-xxl-trtllm-compiled/v0.8.0/",
```
lanking520, line 595: 0.9.0?

sindhuvahinis: Good catch. Updated and uploaded s3 s3://djl-llm/flan-t5-xxl-trtllm-compiled/v0.9.0/

lanking520 (approved):

PR #1875, opened 2024-05-04
Add better token handling under hazardous condition

lanking520:
## Description ##

Reach better handling at problematic points

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch.py
@@ -358,6 +358,9 @@ class Request(object):
             details_dict["inputs"] = self.input_text
             details_dict["parameters"] = self.original_params
             details_dict["prompt_tokens"] = len(self.input_ids)
+        # Special handling for error case
+        elif finish_reason == "error":
+            details_dict["finish_reason"] = finish_reason
         generated_text = self.full_text_prefix
         if last_token:
             generated_text = generated_text + ''.join(self.generated_tokens)
@@ -407,19 +410,14 @@ def stop_on_any_exception(func):
             return func(self, *args, **kwargs)
         except Exception:
             logging.exception("Rolling batch inference error")
-            err = {
-                "data": "",
-                "last": True,
-                "step_token_num": 0,
-                "code": 424,
-                "error": ""
-            }
-            results = []
-            for i in range(
-                    len(self.active_requests) + len(self.pending_requests)):
-                results.append(err)
+            for request in self.active_requests:
+                token = Token([-1], "", -1, None)
+                request.set_next_token(token,
+                                       last_token=True,
+                                       finish_reason="error")
+            response = self.postprocess_results()
             self.reset()
-            return results
+            return response
 
     return try_catch_handling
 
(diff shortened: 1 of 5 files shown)


engines/python/setup/djl_python/tests/test_test_model.py
```diff
@@ -140,3 +147,56 @@ def test_with_env_chat(self):
...
+        serving_properties = {
+            "engine": "Python",
+            "rolling_batch": "auto",
+            "model_id": model_id
+        }
+        result = handler.inference_rolling_batch(
+            inputs, serving_properties=serving_properties)
+        for key, value in result.items():
+            final_dict = json.loads(value)
+            self.assertEqual(final_dict["details"]["finish_reason"], 'error')
```
rohithkrn, line 179: test is failing here

frankfliu (approved):

rohithkrn (approved):

rohithkrn:
I was doing some tests on my side. With current container when streaming error occurs, the response will be
```
1714866779400: {"token": {"id": [338], "text": "is", "log_prob": -0.3300172984600067}}
{"token": {"id": [263], "text": " a", "log_prob": -1.0825505554676056}}
{"token": {"id": [2107], "text": " great", "log_prob": -2.0450351238250732}}
{"token": {"id": [1342], "text": " example", "log_prob": -0.7814879417419434}}
{"error":"","code":424}
```

After this PR, the response will be
```
1714867941864: {"token": {"id": [338], "text": "is", "log_prob": -0.3300172984600067}}
{"token": {"id": [2000], "text": " called", "log_prob": -2.8891883194446564}}
{"token": {"id": [1346], "text": " "", "log_prob": -2.0550572872161865}}
{"token": {"id": [2177], "text": "Par", "log_prob": -0.09167909622192383}}
{"token": {"id": [-1], "log_prob": -1}, "generated_text": "is called "Par"}
```

look at the last line. So that way this change can break backwards compatibility in terms how users interpret response.

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -407,19 +407,14 @@ def try_catch_handling(self, *args, **kwargs):
...
-                "code": 424,
-                "error": ""
-            }
-            results = []
-            for i in range(
-                    len(self.active_requests) + len(self.pending_requests)):
-                results.append(err)
+            for request in self.active_requests:
+                token = Token([-1], "", -1, None)
+                request.set_next_token(token,
```
rohithkrn, line 412: Do you think we should also set error response code? https://docs.aws.amazon.com/sagemaker/latest/APIReference/API_runtime_InvokeEndpointWithResponseStream.html#sagemaker-runtime_InvokeEndpointWithResponseStream-response-ModelStreamError

engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -407,19 +410,14 @@ def try_catch_handling(self, *args, **kwargs):
...
+                                       last_token=True,
+                                       finish_reason="error")
```
rohithkrn, line 417: could we also add `repr(exception)` to the finish reason? something like `finish_reason=f"error: repr(e)" to let the users know what the error is?

lanking520: for some customers, they do not want to expose error log in stack traces, those info can be used by hacker to intercept the content inside

PR #1887, opened 2024-05-08
[CI] remove redundant pipeline and adding retry

lanking520:
## Description ##

- Remove 1 pipeline and merge with the other
- Adding retry logic to failed docker command


tests/integration/remove_container.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+function fail {
+  echo $1 >&2
+  exit 1
+}
+
+function retry {
+  local n=1
```
nskool, line 9: Nit - if we want 5 retries, I believe this should be `local n=0`

nskool (approved):

PR #1889, opened 2024-05-08
Add multi-token support

ymwangg:
## Description ##

Per-token text is not supported. This may needs some changes on vllm/lmi_dist side.
Per-token cum_logprob is not supported, but should be easy to add.


engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
```diff
@@ -39,15 +39,29 @@ def update_request_cache_with_output(request_cache: OrderedDict,
...
+            logprobs[token_id].logprob
+            for token_id, logprobs in zip(new_token_ids, new_logprobs_list)
+        ]
+    else:
+        new_logprobs = [None] * len(new_token_ids)
+
+    request_cache[request_id]["token_ids"] = new_token_ids
+    request_cache[request_id]["logprobs"] = new_logprobs
+    request_cache[request_id]['cumulative_logprob'] = seq_output.cumulative_logprob
+    request_cache[request_id]["text"] = seq_output.text
```
lanking520, line 61: will we have duplicated text reporting for SD?

lanking520 (changes requested):
Per discussion, we will fix the internal issue at engine layer to merge this PR. Will lift the block once that is done

engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
ymwangg: It should work now.

lanking520 (approved):

PR #1891, opened 2024-05-08
[Benchmark] dataset preparation

lanking520:
## Description ##

Prepare dataset per guidance

### DIFF
--- a/tests/integration/llm/dataset_prep.py
+++ b/tests/integration/llm/dataset_prep.py
@@ -2,21 +2,46 @@ import urllib.request
 import json
 
 dataset_urls = {
-    "gsm8k": "https://raw.githubusercontent.com/ymwangg/vllm-test/main/dataset/gsm8k.jsonl",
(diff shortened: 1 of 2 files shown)


tests/integration/llm/dataset_prep.py
```diff
@@ -0,0 +1,83 @@
...
+
+def humaneval_parsing(row: dict, parameters):
+    return {"inputs": row["prompt"], "parameters": parameters, "stream": True}
+
+
+def mbpp_parsing(row: dict, parameters):
+    return {"inputs": row["prompt"], "parameters": parameters, "stream": True}
+
+
+def mtbench_parsing(row: dict, parameters):
```
frankfliu, line 60: We don't need add parameters in dataset any more, we can pass parameters in awscurl command now.

lanking520: customer can choose to override if needed. Those are the good defaults from science

frankfliu (approved):

PR #1893, opened 2024-05-09
[IB] remove engine from the settings

lanking520:
## Description ##

Remove engine from the settings

### DIFF
--- a/tests/integration/benchmark/nightly/g5-12xl.txt
+++ b/tests/integration/benchmark/nightly/g5-12xl.txt
@@ -5,7 +5,6 @@ ENGINE={vllm,lmi-dist}
 [container]
 deepjavalibrary/djl-serving:lmi-nightly
 [serving_properties]
-engine=Python
 option.rolling_batch=$ENGINE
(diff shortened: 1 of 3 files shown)


tests/integration/benchmark/nightly/g5-12xl.txt
```diff
@@ -5,7 +5,6 @@ ENGINE={vllm,lmi-dist}
 [container]
 deepjavalibrary/djl-serving:lmi-nightly
 [serving_properties]
-engine=Python
```
frankfliu, line 8: Why removing it?

It doesn't hurt and make it easy to distinguish from future rust engine

lanking520: this is wrong today. For LMI-Dist, we will need MPI engine

frankfliu: We shouold automatically add `option.mpi_model=true`, if not then it's a bug

lanking520: it is added

PR #1897, opened 2024-05-10
Refactor huggingface and rolling batch handlers

sindhuvahinis:
### Testing

Tested in my EC2 local machine. Built a container and ran the CI by pushing the image in my private dockerhub account. 
https://github.com/deepjavalibrary/djl-serving/actions/runs/9019241325/job/24781785296

In the CI, one of the test is failing, but can't reproduce it in my local EC2 machine.

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -15,22 +15,25 @@ import os
 import re
 
 import torch
-from transformers import (
-    pipeline, Pipeline, Conversation, AutoModelForCausalLM,
-    AutoModelForSeq2SeqLM, AutoTokenizer, AutoConfig,
-    AutoModelForSequenceClassification, AutoModelForTokenClassification,
-    AutoModelForQuestionAnswering, StoppingCriteria, StoppingCriteriaList)
+from transformers import (pipeline, Pipeline, Conversation,
+                          AutoModelForCausalLM, AutoModelForSeq2SeqLM,
+                          AutoTokenizer, AutoModelForSequenceClassification,
+                          AutoModelForTokenClassification,
+                          AutoModelForQuestionAnswering, StoppingCriteria,
+                          StoppingCriteriaList)
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
-from peft import PeftConfig, PeftModel, PeftModelForCausalLM
+from peft import PeftModel, PeftModelForCausalLM
 
 from djl_python.encode_decode import encode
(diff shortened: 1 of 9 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -172,8 +150,7 @@ def initialize(self, properties: dict):
                 self.load_stopping_criteria_list(properties["stop_sequence"])
 
         self.input_format_configs = InputFormatConfigs(
-            is_rolling_batch=is_rolling_batch_enabled(
-                self.hf_configs.rolling_batch),
+            is_rolling_batch=False,
```
frankfliu, line 153: Why this always false?

engines/python/setup/djl_python/huggingface.py
```diff
@@ -593,9 +497,14 @@ def handle(inputs: Input):
...
     Default handler function
     """
+    global _service
     if not _service.initialized:
+        properties = inputs.get_properties()
+        if "rolling_batch" in properties and properties.get(
+                "rolling_batch") != RollingBatchEnum.disable.value:
+            _service = RollingBatchService()
         # stateful model
-        _service.initialize(inputs.get_properties())
```
frankfliu, line 598: Basically, you made the RollingBatch class a top level entryPoint
Let's have a design discussion about this decision.

sindhuvahinis: Sure. Not really a top level entryPoint. RollinBatch does not have `handle` method in it. And the UX still does not change, entryPoint would still be `djl_python.huggingface` for all the vllm, lmi-dist, scheduler rolling batch entryPoint.

engines/python/setup/djl_python/rolling_batch/rolling_batch_service.py
```diff
@@ -0,0 +1,145 @@
...
+    "GPTNeoXForCausalLM",
+    "T5ForConditionalGeneration",
+    "LlamaForCausalLM",
+    "FalconForCausalLM",
+    "MPTForCausalLM",
+    "GPTBigCodeForCausalLM",
+}
+
+
+def get_rolling_batch_class_from_str(rolling_batch_type: str, is_mpi: bool,
```
lanking520, line 29: does TRTLLM also falls into this category?

sindhuvahinis: This PR does not cover it. But it should be easy to re-use this code for trtllm as well.

PR #1907, opened 2024-05-15
Updates default management port

zachgk:
This updates the default management port to be 8081 to match the documented behavior in
https://docs.djl.ai/docs/serving/serving/docs/configurations_global.html#configure-listening-port.


lanking520 (approved):

frankfliu:
We'd better update the document. the default behavior is more convenient for user.

PR #1909, opened 2024-05-15
[ci] Add text embedding integration test

xyang16:
## Description ##

### DIFF
--- a/tests/integration/embedding/client.py
+++ b/tests/integration/embedding/client.py
@@ -1,0 +1,235 @@
+import sys
+import requests
+import argparse
+import subprocess as sp
+import logging
+import re
+import os
+import math
+import json
+import shutil
+from random import randrange
+
+logging.basicConfig(level=logging.INFO)
+
+
+def get_model_name():
+    endpoint = f"http://127.0.0.1:8080/models"
+    res = requests.get(endpoint).json()
+    return res["models"][0]["modelName"]
+
+
+onnx_model_spec = {
+    "bge-base-en-v1.5": {
+        "max_memory_per_gpu": [1.0, 1.0, 1.0, 1.0],
+        "batch_size": [1, 2, 4, 8],
+    },
+    "bge-reranker": {
+        "max_memory_per_gpu": [2.0, 2.0, 2.0, 2.0],
+        "batch_size": [1, 2, 4, 8],
+    }
+}
+
+
+def check_worker_number(desired):
+    model_name = get_model_name()
+    endpoint = f"http://127.0.0.1:8080/models/{model_name}"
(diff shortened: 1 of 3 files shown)


tests/integration/embedding/client.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def get_model_name():
+    endpoint = f"http://127.0.0.1:8080/models"
+    res = requests.get(endpoint).json()
+    return res["models"][0]["modelName"]
+
+
+onnx_model_spec = {
+    "bge-base-en-v1.5": {
```
frankfliu, line 23: Can we use a tiny model?

tests/integration/embedding/client.py
```diff
@@ -0,0 +1,237 @@
+import sys
```
frankfliu, line 1: Can we re-use existing client.py?

tests/integration/embedding/client.py
lanking520: I think it is ok, not too big to me

tests/integration/embedding/client.py
lanking520: you can import some methods if needed

tests/integration/embedding/client.py
xyang16: BAAI/bge-small-en-v1.5?

tests/integration/embedding/client.py
xyang16: The llm/client is much different. There's no output tokens, etc. So that's why I put it in another client.py.

tests/integration/embedding/client.py
```diff
@@ -0,0 +1,237 @@
...
+        for i, d in enumerate(data):
+            with open(os.path.join(dataset_dir, f"prompt{i}.txt"), "w") as f:
+                f.write(json.dumps(d))
+        command_data = f"--dataset {dataset_dir}"
+    else:
+        json_data = json.dumps(data)
+        command_data = f"-d '{json_data}'"
+    command = (f"./awscurl -c {concurrency} "
+               f"-N {num_run} -X POST {endpoint} --connect-timeout 120 "
+               f"-H {headers} {command_data} -P -t")
```
lanking520, line 84: you don't need `-t` here right?

tests/integration/embedding/client.py
frankfliu: can we just use `from huggingface_hub import InferenceClient` directly?

frankfliu: We should ensure our output is compatible with TEI

frankfliu (approved):

lanking520 (approved):

PR #1913, opened 2024-05-15
[IB] support benchmark matrix

lanking520:
## Description ##

Supporting benchmark handle in matrix

### DIFF
--- a/tests/integration/instant_benchmark.py
+++ b/tests/integration/instant_benchmark.py
@@ -29,6 +29,10 @@ parser.add_argument("--instance",
                     required=False,
                     type=str,
                     help="The current instance name")
+parser.add_argument("--record",
+                    required=False,
+                    type=str,
+                    help="Place to record metrics")
 
 parser.add_argument("--job", required=False, type=str, help="The job string")
 args = parser.parse_args()
@@ -110,6 +114,7 @@ def parse_raw_template(url, override_container):
     commandline = []
     requirements = []
     vars = []
+    benchmark_vars = []
     info = None
     while iterator < len(lines):
         if '[test_name]' == lines[iterator]:
@@ -148,6 +153,12 @@ def parse_raw_template(url, override_container):
                     lines[iterator]):
                 vars.append(lines[iterator])
                 iterator += 1
+        elif '[benchmark_vars]' == lines[iterator]:
+            iterator += 1
+            while iterator < len(lines) and not is_square_bracket(
+                    lines[iterator]):
+                benchmark_vars.append(lines[iterator])
+                iterator += 1
         elif '[info]' == lines[iterator]:
             info = []
             iterator += 1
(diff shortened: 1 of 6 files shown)


lanking520:
https://github.com/deepjavalibrary/djl-serving/actions/runs/9101724608

tests/integration/instant_benchmark.py
```diff
@@ -174,8 +185,14 @@ def parse_raw_template(url, override_container):
...
             mul_results = multiply_template_with_vars(name, cur_result, vars)
+            # each of the replicated deployment options
             for r in mul_results.values():
-                r['awscurl'] = r['awscurl'].encode().hex()
+                replicated_awscurl = multiply_template_with_vars(
+                    '', {'awscurl': cur_result['awscurl']}, benchmark_vars)
+                for option in replicated_awscurl.keys():
+                    replicated_awscurl[option] = replicated_awscurl[option][
+                        'awscurl'].encode().hex()
+                r['awscurl'] = replicated_awscurl
```
zachgk, line 195: You also need to clear the benchmark_vars below

tests/integration/record_benchmark.py
```diff
@@ -46,55 +51,61 @@
...
-        },
-    ]
+    benchmark_vars = data["benchmark_vars"] if data["benchmark_vars"] else ""
+    metric_name = lambda n: (f"lmi_{data['instance']}_{esc(data['image'])}"
+                             f"_{esc(job_name)}_{esc(benchmark_vars)}_{n}")
+    metric_data = []
+    for metric, unit in cloudwatch_report_schema.items():
+        if metric in data.keys():
+            metric_data.append({
+                'MetricName': metric_name(metric),
```
zachgk, line 100: There is some fudging of the metric names here. We have names in the cloudwatch_report_schema, json produced by awscurl, data object, and log lines parsed in `data_from_client` from `benchmark.log`. These might differ, but we should try to get them to all match. Right now, the `benchmark.log` will differ from the cloudwatch upload and the awscurl ones, while the data object could have either

Maybe the quickest is to make the names from `benchmark.log` match the ones in the cloudwatch_report_schema/awscurl. This should align all the names. If you do this, you will also need to update the tables in the cloudwatch dashboard as the metrics they are displaying are being renamed

tests/integration/record_benchmark.py
```diff
@@ -46,55 +51,61 @@
 
 data = {}
```
zachgk, line 52: If we are using local `data`, maybe move this declaration down near the `if __name__ == "__main__"` part

tests/integration/instant_benchmark.py
```diff
@@ -219,23 +236,35 @@ def machine_translation(machine_name: str):
...
+
     bash_command = [
-        'set -euo pipefail', 'echo "Start Launching container..."',
+        'set -euo pipefail',
+        'echo "Start Launching container..."',
         f"docker pull {container}",
         f"./launch_container.sh {container} $PWD/models {machine_translation(instance)}",
-        job_template['awscurl'] + " | tee benchmark.log"
     ]
+    bash_command.extend(benchmark_command)
```
zachgk, line 267: nit, but can we just use the same variable for benchmark_command and bash_command. Then all the parts run will be in the file in order

lanking520:
https://github.com/deepjavalibrary/djl-serving/actions/runs/9104207983

zachgk (approved):

PR #1914, opened 2024-05-15
[ci] Reduce GPU mem_util for KVCache for DBRX model

maaquib:
## Description ##


maaquib:
Should be fixed by https://github.com/deepjavalibrary/djl-serving/pull/1949

PR #1920, opened 2024-05-16
[BugFix] Fix bug when invalid adapters are passed

rohithkrn:
## Description ##

There are two cases where the current implementation is buggy for adapters.

* When unregistered adapter is specified in the inference request
* When incorrect number of adapters is specified for client side batching
In both these cases, input is sent to the backend engine and error is populated for the request as well. As a result the model server will send the error response back to the client while the engine still keeps processing the request.

In this change, when an error occurs, the input request will not be sent to the engine.

### DIFF
--- a/engines/python/setup/djl_python/utils.py
+++ b/engines/python/setup/djl_python/utils.py
@@ -148,5 +148,5 @@ def _fetch_adapters_from_input(input_map: dict, inputs: Input):
 
 def _validate_adapters(adapters_per_item, adapter_registry):
     for adapter_name in adapters_per_item:
-        if adapter_name not in adapter_registry and adapter_name != "":
+        if adapter_name and adapter_name not in adapter_registry:
             raise ValueError(f"Adapter {adapter_name} is not registered")
--- a/engines/python/setup/djl_python/utils.py
+++ b/engines/python/setup/djl_python/utils.py
@@ -56,8 +56,6 @@ def parse_input_with_formatter(inputs: Input,
             if input_format_configs.is_adapters_supported:
                 adapters_per_item, found_adapter_per_item = _parse_adapters(
                     _inputs, input_map, item, adapter_registry)
-                adapters.extend(adapters_per_item)
-                found_adapters = found_adapter_per_item or found_adapters
         except Exception as e:  # pylint: disable=broad-except
             logging.warning(f"Parse input failed: {i}")
             input_size.append(0)
@@ -66,6 +64,9 @@ def parse_input_with_formatter(inputs: Input,
 
         input_data.extend(_inputs)
         input_size.append(len(_inputs))
+
+        adapters.extend(adapters_per_item)
+        found_adapters = found_adapter_per_item or found_adapters
 
         for _ in range(input_size[i]):
             parameters.append(_param)
--- a/engines/python/setup/djl_python/utils.py
+++ b/engines/python/setup/djl_python/utils.py
@@ -29,7 +29,7 @@ class InputFormatConfigs:
 
 def parse_input_with_formatter(inputs: Input,
                                input_format_configs: InputFormatConfigs,
-                               adapter_registry: dict) -> ParsedInput:
+                               adapter_registry: dict = {}) -> ParsedInput:
     """
     Preprocessing function that extracts information from Input objects.
     :param input_format_configs: format configurations for the input.
@@ -147,5 +147,5 @@ def _fetch_adapters_from_input(input_map: dict, inputs: Input):
 
 def _validate_adapters(adapters_per_item, adapter_registry):
     for adapter_name in adapters_per_item:
-        if adapter_name not in adapter_registry:
+        if adapter_name not in adapter_registry and adapter_name != "":
             raise ValueError(f"Adapter {adapter_name} is not registered")
--- a/engines/python/setup/djl_python/utils.py
+++ b/engines/python/setup/djl_python/utils.py
@@ -27,9 +27,9 @@ class InputFormatConfigs:
     tokenizer: Any = None
 
 
-def parse_input_with_formatter(
-        inputs: Input,
-        input_format_configs: InputFormatConfigs) -> ParsedInput:
+def parse_input_with_formatter(inputs: Input,
+                               input_format_configs: InputFormatConfigs,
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/utils.py
```diff
@@ -27,9 +27,9 @@ class InputFormatConfigs:
     tokenizer: Any = None
 
 
-def parse_input_with_formatter(
-        inputs: Input,
-        input_format_configs: InputFormatConfigs) -> ParsedInput:
+def parse_input_with_formatter(inputs: Input,
+                               input_format_configs: InputFormatConfigs,
+                               adapter_registry: dict) -> ParsedInput:
```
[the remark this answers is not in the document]
rohithkrn: I don't think conceptually `adapter_registry` fits well within `InputFormatConfigs` as it's not a config. 

I agree that `adapter_registry` arg should be an optional arg, good point.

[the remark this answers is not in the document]
davidthomas426: > Can adapter_registry be a part of InputFormatConfigs? The reason for input format configs, is in the future we would want to make this a public API and we dont want the number of parameters to grow here.
> 
> Also, adapter_registry optional with default value empty dict, since tensorrt llm handler does not support adapters yet.

"in the future we would want to make this a public API" -- public to whom? Can't we address that "in the future" then? And doesn't defaulting to `{}` basically solve that concern here?

If the number of parameters starts to grow and the API being public becomes a concern, can't we create some kind of "Context" container or something to hold these parameters at that time?

engines/python/setup/djl_python/utils.py
```diff
@@ -51,29 +51,25 @@ def parse_input_with_formatter(
         try:
             content_type = item.get_property("Content-Type")
             input_map = decode(item, content_type)
+            _inputs, _param, is_client_side_batch[i] = _parse_inputs_params(
```
davidthomas426, line 54: Why `_` on local variables? I'm not aware of this as a Python convention. (Note: I recognize it was like that before this PR).

engines/python/setup/djl_python/utils.py
```diff
@@ -51,29 +51,25 @@ def parse_input_with_formatter(
...
+                input_map, item, input_format_configs)
+            if input_format_configs.is_adapters_supported:
+                adapters_per_item, found_adapter_per_item = _parse_adapters(
+                    _inputs, input_map, item, adapter_registry)
+                adapters.extend(adapters_per_item)
+                found_adapters = found_adapter_per_item or found_adapters
         except Exception as e:  # pylint: disable=broad-except
```
davidthomas426, line 61: Should this really `except Exception`? Maybe it should be more targeted than that. (Note: I recognize it was like that before this PR).

engines/python/setup/djl_python/utils.py
```diff
@@ -51,29 +51,25 @@ def parse_input_with_formatter(
         try:
             content_type = item.get_property("Content-Type")
             input_map = decode(item, content_type)
+            _inputs, _param, is_client_side_batch[i] = _parse_inputs_params(
+                input_map, item, input_format_configs)
+            if input_format_configs.is_adapters_supported:
+                adapters_per_item, found_adapter_per_item = _parse_adapters(
+                    _inputs, input_map, item, adapter_registry)
+                adapters.extend(adapters_per_item)
+                found_adapters = found_adapter_per_item or found_adapters
```
davidthomas426, line 60: nit: I think `adapters.extend` and updating `found_adapters` should happen outside the try-except, to make it clearer that these state variables only get updated if we make it past the `continue` in except.

engines/python/setup/djl_python/utils.py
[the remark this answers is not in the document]
sindhuvahinis: > "in the future we would want to make this a public API" -- public to whom? Can't we address that "in the future" then? 
> 
To people who are writing custom pre/post processing in their custom handler. We have output_formatter as Callable function for output formatting, we would want the same thing for input parsing as well, as our customers are asking for it.  

> And doesn't defaulting to `{}` basically solve that concern here?
>
Yes, but we could UX could be better. We dont have to figure it today.

engines/python/setup/djl_python/utils.py
rohithkrn: there are also [places](https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/utils.py#L89-L94) in this file where that convention is followed. I don't know why. I do not want to mix those changes in this PR, will raise a separate PR to isolate the changes.

engines/python/setup/djl_python/utils.py
rohithkrn: updated.

engines/python/setup/djl_python/utils.py
rohithkrn: May be the idea was to suppress any kind of exception? Otherwise, we should target specific exceptions which is not in the scope for this PR.

zachgk (approved):

engines/python/setup/djl_python/utils.py
davidthomas426: Yeah but, suppressing any exception and trying to just continue the loop is bad.

If it's "not in the scope of this PR", then I guess I should create an Issue for stuff like this?

engines/python/setup/djl_python/utils.py
```diff
@@ -147,3 +144,9 @@ def _fetch_adapters_from_input(input_map: dict, inputs: Input):
         adapters_per_item = [adapters_per_item]
 
     return adapters_per_item
+
+
+def _validate_adapters(adapters_per_item, adapter_registry):
+    for adapter_name in adapters_per_item:
+        if adapter_name not in adapter_registry and adapter_name != "":
```
davidthomas426, line 151: Does this also need to handle `None`? I'm not sure how this works.

Maybe this is better, so it handles None and "" and any other falsy values: `if adapter_name and adapter_name not in adapter_registry:`

rohithkrn: I don't think `adapter` will be `None` but sure this is better, updated.

davidthomas426 (approved):

PR #1921, opened 2024-05-16
Add request output dataclasses

sindhuvahinis:
## Description ##

Breaking down the changes of refactoring the output formatter and supporting multiple sequences into multiple PRs. 
Added minimum getters and setters in the PR, will be adding more in the other PRs.

### DIFF
--- a/engines/python/setup/djl_python/request_io.py
+++ b/engines/python/setup/djl_python/request_io.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python
+#
+# Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+from dataclasses import field, dataclass
+from typing import List, Optional, Callable, Any, Dict, Union
+
+from djl_python.rolling_batch.rolling_batch import Token
+
+
+class Iterator:
+
+    def __init__(self):
+        self._index = 0
+
+    def get_index(self):
+        return self._index
+
+    def next_index(self):
+        next_index = self._index
+        self._index += 1
+        return next_index
+
+
+@dataclass
+class Sequence:
+    """Dataclass to store the generated tokens details of all the sequences in the request.
+    Attributes:
+        tokens: generated tokens of the sequence.
+        top_tokens: top tokens of the sequence.
+        cumulative_log_prob: cumulative log probability of the sequence.
+        finish_reason: finish reason of the sequence.
+        stop_reason: stop reason of the sequence.
+        first_token: first token of the sequence.
+        last_token: last token of the sequence.
+    """
+    tokens: List[Token] = field(default_factory=lambda: [])
+    top_tokens: Optional[List[List[Token]]] = field(default_factory=lambda: [])
+    cumulative_log_prob: float = 0.0
+    finish_reason: str = False
+    last_token: bool = False
+    first_token: bool = True
+    stop_reason: Optional[str] = None
+    _tokens_iterator: Optional[Iterator] = field(init=False, default=None)
+    _top_tokens_iterator: Optional[Iterator] = field(init=False, default=None)
+
+    def __post_init__(self):
+        self._tokens_iterator = Iterator()
+        self._top_tokens_iterator = Iterator()
+
+    def set_next_token(self, token: Token):
+        self.tokens.append(token)
+
+    def set_next_top_tokens(self, top_tokens: List[Token]):
+        self.top_tokens.append(top_tokens)
+
+    def has_next_token(self):
+        return (self._tokens_iterator.get_index() + 1) < len(self.tokens)
+
+    def has_next_top_tokens(self):
+        return (self._top_tokens_iterator.next_index() + 1) < len(
+            self.top_tokens)
+
+    def get_next_token(self):
+        if self.has_next_token():
+            index = self._tokens_iterator.next_index()
+            return self.tokens[index]
+        return None
+
+    def get_next_top_tokens(self):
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+        finish_reason: finish reason of the sequence.
+        stop_reason: stop reason of the sequence.
+        generated_text: generated text of the sequence.
+    """
+    tokens: List[Token] = field(default_factory=lambda: [])
+    top_tokens: Optional[List[List[Token]]] = field(default_factory=lambda: [])
+    cumulative_log_prob: float = 0.0
+    finish_reason: str = False
+    stop_reason: Optional[str] = None
+    generated_text: str = ""
```
lanking520, line 23: would recommend to keep this as List[str] to store the token text. If customer want to get the full text, we just need to join them

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+class Sequence:
+    """Dataclass to store the generated tokens details of all the sequences in the request.
+    Attributes:
+        tokens: generated tokens of the sequence.
+        top_tokens: top tokens of the sequence.
+        cumulative_log_prob: cumulative log probability of the sequence.
+        finish_reason: finish reason of the sequence.
+        stop_reason: stop reason of the sequence.
+        generated_text: generated text of the sequence.
+    """
```
lanking520, line 17: better to also keep an iterator here. Since there could be more set_token calls than get token calls

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def set_next_token(self, token: Token):
+        self.tokens.append(token)
+        self.generated_text += token.text
+
+    def set_next_top_tokens(self, top_tokens: List[Token]):
+        self.top_tokens.append(top_tokens)
+
+    @property
+    def next_token(self):
```
lanking520, line 33: this should be called last token. And `@property` labe may not be correct. If I understand correctly, this will make a method to be permanent into element after you fetch once. For example:

```python
self.next_token() => 5
...
# it will be forever 5
self.next_token() => 5
```

It should reflect the iterator location and also give some boolean tells wether it is the end of current token list

engines/python/setup/djl_python/request_output.py
lanking520: Maybe this is not needed, I recalled Token object contains that information

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+        return self.tokens[-1]
+
+    @property
+    def next_top_tokens(self):
```
lanking520, line 37: same applies here for iterator

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+class RequestInput:
+    """Base class for all request inputs.
+    Attributes:
+        request_id: The request ID.
+        output_formatter: Output formatter of the request
+        parameters: parameters in the request payload
+    """
+    request_id: int
+    output_formatter: Union[Callable, str] = None
+    parameters: Dict = field(default_factory=lambda: [])
```
lanking520, line 60: not only parameters, we might want to keep the whole json here

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,132 @@
...
+        finished: Whether the request is finished.
+    """
+    request_id: int
+    input: RequestInput
+    prompt_tokens_details: List[dict] = field(default_factory=lambda: {})
+    finished: bool = False
+
+
+@dataclass
+class TextGenerationOutput(RequestOutput):
```
lanking520, line 98: when will we use this class?

engines/python/setup/djl_python/request_output.py
sindhuvahinis: No, `@property` has getter and setters and it calls this getter each and everytime we get. But I understand property is more intended for read only properties. Can change this to getter method.

engines/python/setup/djl_python/request_output.py
sindhuvahinis: I agree, but the logic to get the whole json is in parse_input https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/setup/djl_python/utils.py#L94
We only parse "inputs" and "params",  only for chat completion, we take the whole json and parsed into ChatProperties. 

And this params is passed down to rolling batch inference.

lanking520: then it is better to register a new request there if possible. SF customer need some custom field from the json other than parameter itself

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,120 @@
+from dataclasses import field, dataclass
```
frankfliu, line 1: this file contains both input and outputs, better use a difference file name

engines/python/setup/djl_python/request_output.py
```diff
@@ -0,0 +1,120 @@
...
+    """Base class for all request outputs.
+    Attributes:
+        request_id: The request ID.
+        input: The request input.
+        prompt_tokens_details: prompt tokens details such as their log_probs and token_ids.
+        finished: Whether the request is finished.
+    """
+    request_id: int
+    input: RequestInput
+    prompt_tokens_details: List[dict] = field(default_factory=lambda: {})
```
frankfliu, line 87: This still tight with nlp

engines/python/setup/djl_python/request_output.py
frankfliu: Add license header

sindhuvahinis: Done

engines/python/setup/djl_python/request_output.py
sindhuvahinis: Moved this to text generation.

engines/python/setup/djl_python/request_output.py
sindhuvahinis: This class should be used to store all the text generation related output, which is the only one we have right now.

engines/python/setup/djl_python/request_output.py
sindhuvahinis: Sure, will do this in the next PRs.

lanking520 (approved):

PR #1924, opened 2024-05-17
Refactor tgi compat and removed token_id as list

sindhuvahinis:
## Description ##
In this PR:
1. option.tgi_compat in serving.properties and OPTION_TGI_COMPAT env both should work
2. Refactored rolling_batch class to accept Configs(parsed properties) instead of kwargs. 
3. [Breaks backward compatibility] token_id won't be list anymore. Since, now speculative decoding, we are adding multiple tokens. 

Testing:
Tested with scheduler rolling batch with env and serving.properties for bloom-560m model.

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/properties.py
+++ b/engines/python/setup/djl_python/properties_manager/properties.py
@@ -57,8 +57,7 @@ class Properties(BaseModel):
     output_formatter: Optional[Union[str, Callable]] = None
     waiting_steps: Optional[int] = None
     is_mpi: bool = False
-    tgi_compat: Optional[bool] = os.environ.get("OPTION_TGI_COMPAT",
-                                                "false").lower() == "true"
+    tgi_compat: Optional[bool] = False
 
     # Spec_dec
     draft_model_id: Optional[str] = None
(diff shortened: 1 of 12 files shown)


engines/python/setup/djl_python/rolling_batch/rolling_batch.py
```diff
@@ -370,6 +369,8 @@ def set_next_token(self,
         # Special handling for error case
         elif finish_reason == "error":
             details_dict["finish_reason"] = finish_reason
+        if self.output_formatter == _json_output_formatter:
+            details_dict["tgi_compat"] = self.tgi_compat
```
frankfliu, line 373: Why only set when use json?

lanking520: because only applied in json you need to add []

engines/python/setup/djl_python/properties_manager/properties.py
```diff
@@ -57,6 +57,8 @@ class Properties(BaseModel):
     output_formatter: Optional[Union[str, Callable]] = None
     waiting_steps: Optional[int] = None
     is_mpi: bool = False
+    tgi_compat: Optional[bool] = os.environ.get("OPTION_TGI_COMPAT",
```
lanking520, line 60: if you are fetching from here, you could just check whether `tgi_compat` is set right? DJLServing will be able to pick it up and trim the `OPTION_` part of it

sindhuvahinis: :) Yeah, right!! Let me remove this

sindhuvahinis: Done

frankfliu (approved):

tosterberg (approved):

PR #1927, opened 2024-05-17
[fix] inf2 pipeline fix for adapter unsupported

tosterberg:
## Description ##

Add logic to ensure that when adapters are not supported that parse input still passes

### DIFF
--- a/engines/python/setup/djl_python/utils.py
+++ b/engines/python/setup/djl_python/utils.py
@@ -65,8 +65,9 @@ def parse_input_with_formatter(inputs: Input,
         input_data.extend(_inputs)
         input_size.append(len(_inputs))
 
-        adapters.extend(adapters_per_item)
-        found_adapters = found_adapter_per_item or found_adapters
+        if input_format_configs.is_adapters_supported:
+            adapters.extend(adapters_per_item)
+            found_adapters = found_adapter_per_item or found_adapters
 
         for _ in range(input_size[i]):
             parameters.append(_param)


lanking520 (approved):

engines/python/setup/djl_python/utils.py
```diff
@@ -65,8 +65,9 @@ def parse_input_with_formatter(inputs: Input,
         input_data.extend(_inputs)
         input_size.append(len(_inputs))
 
-        adapters.extend(adapters_per_item)
-        found_adapters = found_adapter_per_item or found_adapters
+        if input_format_configs.is_adapters_supported:
+            adapters.extend(adapters_per_item)
+            found_adapters = found_adapter_per_item or found_adapters
```
davidthomas426, line 70: Is `found_adapters` supposed to be doing something here on the right-hand side? It's set to `False` at the top of this function, and then this is its first use.

Just wanted to make sure this isn't some kind of bug that crept in at some point.

tosterberg: @rohithkrn It looks like this was introduced in https://github.com/deepjavalibrary/djl-serving/pull/1920. It looks like this was moved outside of the try catch per this discussion with David.

> nit: I think adapters.extend and updating found_adapters should happen outside the try-except, to make it clearer that these state variables only get updated if we make it past the continue in except.

@davidthomas426  It looks like we are just getting a bool signaling if we have found an adapter or if it was set to true when calling this method, `found_adapters` is then passed to ParsedInput. I do think this is a little hard to read but it also isn't immediately clear what the rewrite would look like to me.

rohithkrn: yeah, I missed the check after moving outside the `try` block. `found_adapters` is used in the handler to determine whether adapter_data should be added to the Request object in the rolling batch inference. It is used to completely avoid processing adapters in the non-lora case. We can remove it as well set adapters for each request to `None` for non lora case.

PR #1931, opened 2024-05-18
Move request and output_formatter out of rolling_batch.py

sindhuvahinis:
## Description ##

Breaking down the changes of refactoring the output formatter into multiple PR.

1. Moving all output formatters function to a new python file `output_formatter.py`. This is because, for the new proposed output_formatter signature, I am adding more changes which makes the rolling_batch.py really long. 
2. Also moved Request class to `request.py`
3. Moved `Token` to `request_io.py`


engines/python/setup/djl_python/request_io.py
```diff
@@ -13,7 +13,54 @@
...
+        self.id = id
+        self.text = text
+        self.log_prob = log_prob
+        self.special_token = special_token
+        self.request_id = None
+
+    def as_dict(self):
+        output = {}
+        if self.id:
+            output["id"] = self.id
```
lanking520, line 44: need to be adjust accordingly based on my PR

sindhuvahinis: Sure, will rebase and adjust

lanking520 (approved):

PR #1935, opened 2024-05-20
[docker] Install djl-converter in cpu and aarch64 container

xyang16:
## Description ##

### DIFF
--- a/serving/docker/Dockerfile
+++ b/serving/docker/Dockerfile
@@ -11,6 +11,7 @@
 # the specific language governing permissions and limitations under the License.
 FROM ubuntu:22.04 AS base
 ARG djl_version=0.28.0~SNAPSHOT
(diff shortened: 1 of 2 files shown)


serving/docker/Dockerfile
```diff
@@ -71,9 +72,12 @@ RUN scripts/install_python.sh && \
     scripts/patch_oss_dlc.sh python && \
     rm -rf /opt/djl/logs && \
     chown -R djl:djl /opt/djl && \
-    rm -rf scripts && pip3 cache purge && \
-    apt-get clean -y && rm -rf /var/lib/apt/lists/* \
+    rm -rf scripts && \
+    pip3 install transformers onnx sentence_transformers && \
```
lanking520, line 76: let's get a verison to those dependencies. Do. we need optimum library?

lanking520:
close this PR with offline discussion to go with alternative paths

PR #1941, opened 2024-05-21
Refactor output_formatter and include all parameters

sindhuvahinis:
## Description ##

This PR refactors all the output formatter signature to support RequestOutput. The changes include

- TextGenerationOutput and TextInput is initialized in Request class. When we reset Request, all these data should be reset as well. 
- Request constructor now also receives tokenizer. 
- While parsing the input, all the extra fields in the payload are fused into `parameters`. 
- Added `tgi_compat` as one of the data variable of RequestInput. 
- prompt_token_details, now we expect the individual handlers to set it as list of tokens, now it can support token as dict with tgi compatibility. 
- Added integration tests to test for `decoder_input_details` for lmi and vllm 
- Output formatter, if the customer uses old signature, it should work with that as well.

### DIFF
--- a/engines/python/setup/djl_python/output_formatter.py
+++ b/engines/python/setup/djl_python/output_formatter.py
@@ -15,33 +15,44 @@ import logging
 import time
 from typing import Union, Callable
 
-from djl_python.request_io import Token
+from djl_python.request_io import Token, TextGenerationOutput, RequestOutput
 
 
-def _json_output_formatter(token: Token, first_token: bool, last_token: bool,
-                           details: dict, generated_text: str, id: int):
+def _json_output_formatter(request_output: RequestOutput):
     """
     json output formatter
 
     :return: formatted output
     """
+    best_sequence = request_output.sequences[
+        request_output.best_sequence_index]
+
+    parameters = request_output.input.parameters
+    generated_text = ""
+    if parameters.get("return_full_text"):
+        generated_text = request_output.input.input_text
+    next_token, first_token, last_token = best_sequence.get_next_token()
     json_encoded_str = f"{{\"generated_text\": \"{generated_text}" if first_token else ""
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/tests/test_rolling_batch.py
```diff
@@ -231,7 +231,12 @@ def test_sse_tgi_compat_fmt(self):
                 "text": " world",
                 "logprob": -0.567854
             },
-            "generated_text": "Hello world"
+            "generated_text": "Hello world",
+            "details": {
+                "finish_reason": "length",
+                "generated_tokens": 3,
+                "inputs": "This is a wonderful day"
+            }
```
sindhuvahinis, line 239: @lanking520 I believe, if tgi_compat is True, then we should always send out details even if it is not explicitly set. Previously we were not sending the details.

engines/python/setup/djl_python/output_formatter.py
```diff
@@ -191,14 +213,50 @@ def _jsonlines_chat_output_formatter(token: Token, first_token: bool,
...
+def sse_response_formatter(request_output: RequestOutput):
     """
     Decorator that used to form as SSE
     """
-    output_str = _jsonlines_output_formatter(*args, **kwargs)
+    output_str = _jsonlines_output_formatter(request_output)
     return f"data: {output_str}\n"
 
 
+def adapt_legacy_output_formatter(request_output: TextGenerationOutput):
```
lanking520, line 224: comments to add a `@deprecate` flag. Plan to remove in 0.29.0

lanking520 (approved):

frankfliu (approved):

engines/python/setup/djl_python/output_formatter.py
sindhuvahinis: Done

PR #1945, opened 2024-05-21
Creates a python Session Manager

zachgk:
The session manager can be used for stateful inference. Right now, I made three variations of the session manager:
- a local session manager for mostly testing, single worker only
- a file session manager that pickles data and stores it in files. Easy to use and good for smaller data
- an mmap session manager for larger data and better performance


engines/python/src/test/resources/sessionecho/simple/model.py
```diff
@@ -0,0 +1,66 @@
...
+    def initialize(self, properties: dict):
+        self.sessions: SessionManager = get_session_manager(properties)
+        self.initialized = True
+
+    def inference(self, inputs: Input):
+        outputs = Output()
+
+        for i, input in enumerate(inputs.get_batches()):
+            data = input.get_as_string()
+            session_id = input.get_property("sagemaker_session_id")
```
frankfliu, line 35:
```suggestion
            session_id = input.get_property("X-Amzn-SageMaker-Session-Id")
```

engines/python/setup/djl_python/__init__.py
```diff
@@ -14,3 +14,6 @@
 from .inputs import Input
 from .outputs import Output
 from .pair_list import PairList
+from .session_manager import SessionManager, Session
+
+__version__ = '0.28.0'
```
tosterberg, line 19: Run ./gradlew clean and recommit - we don't want this version in the codebase

tosterberg (approved):

PR #1946, opened 2024-05-21
[vllm 0.4.2] prompt log probs changes

sindhuvahinis:
## Description ##

vllm changed the PromptLogProbs dataclass in 0.4.2

in vllm 0.3.3
```
PromptLogprobs = List[Optional[Dict[int, float]]]
```

in vllm 0.4.2
```
@dataclass
class Logprob:
    logprob: float
    rank: Optional[int] = None
    decoded_token: Optional[str] = None

PromptLogprobs = List[Optional[Dict[int, Logprob]]]
```

Will test it with vllm 0.4.2, once it is ready.


sindhuvahinis:
Will merge it, once doing sanity testing in EC2 now.

PR #1934, opened 2024-05-20
[dep] Update flash-attn dependency build

maaquib:
## Description ##

Updates vllm dependency to `v0.4.2`. Built flash-attn from @ymwangg `v0.4.2` fork [here](https://github.com/deepjavalibrary/djl-serving/actions/runs/9161478592/job/25186488317)

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

PR #1949, opened 2024-05-21
[dep] Updgrades dependencies for lmi-dist engine

maaquib:
## Description ##

~~Depends on https://github.com/deepjavalibrary/djl-serving/pull/1934~~
Adding `max_model_len` param to ci tests [vllm-project/vllm/issues/2418]

* Nightly build: https://github.com/deepjavalibrary/djl-serving/actions/runs/9215590750
* Large model integration: https://github.com/deepjavalibrary/djl-serving/actions/runs/9214783969
* P4D integ tests: https://github.com/deepjavalibrary/djl-serving/actions/runs/9214819854

### DIFF
--- a/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
+++ b/serving/docker/dockerd-entrypoint-with-cuda-compat.sh
@@ -19,7 +19,7 @@ translateTGIToLMI() {
 if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
     CUDA_COMPAT_MAX_DRIVER_VERSION=$(readlink /usr/local/cuda/compat/libcuda.so.1 |cut -d'.' -f 3-)
     echo "CUDA compat package requires Nvidia driver ⩽${CUDA_COMPAT_MAX_DRIVER_VERSION}"
-    NVIDIA_DRIVER_VERSION=$(sed -n 's/^NVRM.*Kernel Module *\([0-9.]*\).*$/\1/p' /proc/driver/nvidia/version 2>/dev/null || true)
+    NVIDIA_DRIVER_VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader --id=0 2>/dev/null || true)
     echo "Current installed Nvidia driver version is ${NVIDIA_DRIVER_VERSION}"
     if verlte $NVIDIA_DRIVER_VERSION $CUDA_COMPAT_MAX_DRIVER_VERSION; then
         echo "Setup CUDA compatibility libs path to LD_LIBRARY_PATH"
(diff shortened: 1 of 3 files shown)


serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -19,7 +19,7 @@ translateTGIToLMI() {
 if [ -f /usr/local/cuda/compat/libcuda.so.1 ]; then
     CUDA_COMPAT_MAX_DRIVER_VERSION=$(readlink /usr/local/cuda/compat/libcuda.so.1 |cut -d'.' -f 3-)
     echo "CUDA compat package requires Nvidia driver ⩽${CUDA_COMPAT_MAX_DRIVER_VERSION}"
-    NVIDIA_DRIVER_VERSION=$(sed -n 's/^NVRM.*Kernel Module *\([0-9.]*\).*$/\1/p' /proc/driver/nvidia/version 2>/dev/null || true)
+    NVIDIA_DRIVER_VERSION=$(nvidia-smi --query-gpu=driver_version --format=csv,noheader --id=0 2>/dev/null | cut -d'.' -f1 || true)
```
lanking520, line 22: what is this for?

maaquib: I tried the sed command, but it doesn't work, returns an empty string

rohithkrn: existing command didn't work for me, probably they changed the version file?

If we want to use the version file, this worked `grep -oP '(\d+\.\d+\.\d+)' /proc/driver/nvidia/version | sed -n '1p'`.
I prefer using `nvidia-smi` as that's more standard.

serving/docker/lmi.Dockerfile
```diff
@@ -9,31 +9,33 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-ARG version=12.1.1-cudnn8-devel-ubuntu22.04
+ARG version=12.4.1-devel-ubuntu22.04
```
frankfliu, line 12: Why remove cudnn8?

frankfliu: should we manually install cudnn?

maaquib: Going with the same base as vllm: https://github.com/vllm-project/vllm/blob/v0.4.2/Dockerfile#L10

xyang16: Onnxruntime need cudnn

maaquib: Will add it back

lanking520 (approved):

rohithkrn (approved):

PR #1956, opened 2024-05-22
Update shm size for trtllm test

nskool:
## Description ##
1. Noticed during some local runs that, higher MNT value on p4d may cause shm memory OOM, very rare. 
2. This change may help in ruling out role of shm size in intermittent llama2 70b failures we see on p4d for tp8

### DIFF
--- a/tests/integration/launch_container.sh
+++ b/tests/integration/launch_container.sh
@@ -27,6 +27,9 @@ elif [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"*  || "$platform" == *
   runtime="nvidia"
   is_llm=true
   shm="12gb"
+  if [[ "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"*]]; then
+    shm="20gb"
+  fi
 elif [[ "$platform" == *"inf1"* ]]; then # if the platform is inferentia
   host_device="--device /dev/neuron0"
 elif [[ "$platform" == *"inf2"* ]]; then # inf2: pytorch-inf2-24 24 will be the total devices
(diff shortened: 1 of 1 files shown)


tests/integration/launch_container.sh
```diff
@@ -26,7 +26,11 @@ if [[ "$platform" == *"-gpu"* ]]; then # if the platform has cuda capabilities
 elif [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"*  || "$platform" == *"tensorrt-llm"* ]]; then # Runs multi-gpu
   runtime="nvidia"
   is_llm=true
-  shm="12gb"
+  if [[ "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"*]]; then
+    shm="20gb"
```
lanking520, line 30: we should rather check if it is p4d than doing 20gb on all G5 as well. Could we introduce something like p4d or something to make it larger?

nskool: Added a function to check for that

tests/integration/launch_container.sh
```diff
@@ -20,13 +20,31 @@ if [[ "$model_path" == "no_code" ]]; then
...
+}
+
 is_llm=false
 if [[ "$platform" == *"-gpu"* ]]; then # if the platform has cuda capabilities
   runtime="nvidia"
 elif [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"*  || "$platform" == *"tensorrt-llm"* ]]; then # Runs multi-gpu
   runtime="nvidia"
   is_llm=true
-  shm="12gb"
+  if [[ "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"* ]] && [[ $(is_p4d_or_p5) == "true" ]]; then
```
lanking520, line 43: This can be generic, any GPU usage could set shm

lanking520 (approved):

PR #1962, opened 2024-05-23
[docs][lmi] update input/output schema docs with error response details

siddvenk:
## Description ##

Adding error response schema details to the lmi docs. **Note: this is the state of 0.28.0**

This should cover the most common error scenarios when using LMI with the different modes (streaming/non-streaming, rolling-batch/dynamic-batch).

I've tested all the scenarios and validated that this is what the responses look like for various scenarios.

I don't think we have docs for the inference api error responses (at least I can't find them). I've left those out of scope for now as it probably deserves its own doc if we want to add those details.

### DIFF
--- a/serving/docs/lmi/user_guides/lmi_input_output_schema.md
+++ b/serving/docs/lmi/user_guides/lmi_input_output_schema.md
@@ -83,6 +83,45 @@ Example response:
 }
 ```
 
+#### Error Responses
+
+When rolling batch is enabled, errors are returned with HTTP response code 200.
+Error details are returned in the response content.
+
+If the request is poorly formatted (i.e. invalid json in the request body), the response content will be:
+
+```
+{
+    "error": "<error details>",
+    "code": 424
+}
+```
+
+If there is an issue with the request (such as bad generation parameter value), or an error occurs during generation, the response content will be:
+
+When not using streaming:
+
+```
+{
+    "generated_text": "", 
+    "details": {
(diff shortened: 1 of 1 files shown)


lanking520 (approved):

serving/docs/lmi/user_guides/lmi_input_output_schema.md
```diff
@@ -142,6 +181,24 @@ Example response:
...
+    "error": "<error details"
+}
+```
+
+When using dynamic batching with streaming, errors are returned with HTTP response code 200 and content:
+
+```
+<error details>
+```
+
```
sindhuvahinis, line 201: Dynamic batching does not allow streaming as of now, You can stream with batch_size=1, so basically batching is disabled. HF Acc is the only one which does support this use case right now.  And also we deprecated `option.enable_streaming` for rolling batch and so our rolling batch users should use `stream` option in the request body. 

We should start supported the same way for HF Accelerate also.

siddvenk: The rolling batch + stream case is covered above. The callout on dynamic batch being misleading here is good, I'll update.

Not sure what you mean by: 
> We should start supported the same way for HF Accelerate also.

siddvenk: got it - totally agree!

sindhuvahinis (approved):

PR #1966, opened 2024-05-24
fix t5 tokenizer and prompt token failures

rohithkrn:
1. In `lmi-dist`, t5 model impl is different from other models and preprocessor's tokenizer attribute contains `tokenizer` directly instead of `TokenizerGroup`
2. skip sending prompt token details for t5 as it does not populate those in request output. Current code errors out as `request_output.prompt_token_ids` is an integer of bos token and fails in the for loop enumeration.

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
@@ -47,6 +47,8 @@ class LmiDistRollingBatch(RollingBatch):
         :param properties (dict): other properties of the model, such as decoder strategy
         """
         self.lmi_dist_config = LmiDistRbProperties(**properties)
+        self.model_type = getattr(kwargs.get("model_config", None),
+                                  "model_type", None)
         super().__init__(self.lmi_dist_config)
         self.supports_speculative_decoding = supports_speculative_decoding()
         engine_kwargs = {}
@@ -83,8 +85,6 @@ class LmiDistRollingBatch(RollingBatch):
             kwargs["warmup_prefill_tokens"] = _WARMUP_PREFILL_TOKENS
         self.engine = engine_from_args(engine_args, **kwargs)
         self.request_cache = OrderedDict()
-        self.model_type = getattr(kwargs.get("model_config", None),
-                                  "model_type", None)
         self.lora_ids = defaultdict(lambda: len(self.lora_ids) + 1)
 
     def reset(self) -> None:
@@ -96,6 +96,8 @@ class LmiDistRollingBatch(RollingBatch):
         super().reset()
 
     def get_tokenizer(self):
+        if "t5" == self.model_type:
+            return self.engine.preprocessor.tokenizer
         return self.engine.preprocessor.tokenizer.tokenizer
 
     def translate_lmi_dist_params(self, parameters: dict):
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -47,6 +47,8 @@ def __init__(self, model_id_or_path: str, properties: dict, **kwargs):
         :param properties (dict): other properties of the model, such as decoder strategy
         """
         self.lmi_dist_config = LmiDistRbProperties(**properties)
+        self.model_type = getattr(kwargs.get("model_config", None),
+                                  "model_type", None)
```
sindhuvahinis, line 51: QQ: I see we are populating model_config, Where do we populate "model_type"

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -96,6 +96,8 @@ def reset(self) -> None:
         super().reset()
 
     def get_tokenizer(self):
+        if "t5" == self.model_type:
+            return self.engine.preprocessor.tokenizer
```
sindhuvahinis, line 100: Should it be self.model_type.model_type, if model config is passed?

engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
```diff
@@ -75,6 +75,9 @@ def update_request_cache_with_output(request_cache: OrderedDict,
     if "prompt_tokens_details" not in request_cache[
             request_id] and request_output.prompt_logprobs:
```
sindhuvahinis, line 76: Curious, request_output should not have prompt_logprobs as well right?

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
rohithkrn: model_config here is hugging face config.json which has model_type populated.

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
rohithkrn: why? `model_type` is set in line 50.

engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
rohithkrn: from what I can read, `request_output.prompt_logprobs` is set to `[float("nan")]` and seems like it's passing that check.

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
sindhuvahinis: Oh misread it, Model_type is read from model_config. Cool

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
maaquib: NIT: We should probably unify this on the engine side so we don't have to do if-else here. Approving for now

maaquib (approved):

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
rohithkrn: yes, agree. I initially tried doing that but it turned out to be a relatively bigger change and did not want to do that at the last minute before the release.

PR #1969, opened 2024-05-28
[ci] Switch gradle build script from groovy to kotlin

frankfliu:
## Description ##


lanking520:
Let's do this for post 0.28.0

lanking520 (approved):

PR #1972, opened 2024-05-28
[dep] Bump up transformers version to support falcon2

maaquib:
## Description ##

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -330,6 +330,13 @@ lmi_dist_model_list = {
         "option.max_rolling_batch_size": 4,
         "option.trust_remote_code": True
     },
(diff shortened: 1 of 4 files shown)


lanking520 (approved):

tests/integration/llm/prepare.py
```diff
@@ -330,6 +330,13 @@
         "option.max_rolling_batch_size": 4,
         "option.trust_remote_code": True
     },
+    "falcon-11b": {
+        "option.model_id": "tiiuae/falcon-11b",
```
rohithkrn, line 334: nit: it's there in s3 `s3://djl-llm/falcon-11B/`

maaquib: Updated

PR #2009, opened 2024-06-01
lmi cpu container with vLLM

lanking520:
## Description ##

Support CPU container build for vLLM based LLM inference. Tested with LLAMA3-7B, worked, but extremely slow

```
engine=Python
option.rolling_batch=vllm
option.model_id=NousResearch/Hermes-2-Pro-Llama-3-8B
option.tensor_parallel_degree=1
```

### DIFF
--- a/serving/docker/Dockerfile
+++ b/serving/docker/Dockerfile
@@ -72,8 +72,70 @@ RUN scripts/install_python.sh && \
     rm -rf /opt/djl/logs && \
     chown -R djl:djl /opt/djl && \
     rm -rf scripts && pip3 cache purge && \
-    apt-get clean -y && rm -rf /var/lib/apt/lists/* \
+    apt-get clean -y && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


serving/docker/Dockerfile
```diff
@@ -72,8 +72,69 @@ RUN scripts/install_python.sh && \
...
+
+RUN pip install --upgrade pip \
+    && pip install wheel packaging ninja setuptools>=49.4.0 numpy
+
+RUN git clone https://github.com/vllm-project/vllm -b ${vllm_version} && cd vllm && \
+    pip install -v -r requirements-cpu.txt --extra-index-url https://download.pytorch.org/whl/cpu && \
+    VLLM_TARGET_DEVICE=cpu python3 setup.py bdist_wheel
+
+
+FROM base AS lmi-cpu
```
ydm-amazon, line 100: I thought that there could only be one FROM in each Dockerfile, I may be wrong but just want to check

PR #2013, opened 2024-06-03
Convert cuda env tgi variables to lmi

sindhuvahinis:
## Description ##

Translating some of the cuda env variables to LMI format. 

Testing:
Bashed into a container and tested this bash script. `bash /usr/local/bin/dockerd-entrypoint.sh serve`


serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -47,6 +47,12 @@ translateTGIToLMI "SM_NUM_GPUS" "TENSOR_PARALLEL_DEGREE"
 translateTGIToLMI "MAX_CONCURRENT_REQUESTS" "SERVING_JOB_QUEUE_SIZE"
 translateTGIToLMI "MAX_BATCH_PREFILL_TOKENS" "OPTION_MAX_ROLLING_BATCH_PREFILL_TOKENS"
 translateTGIToLMI "MAX_BATCH_SIZE" "OPTION_MAX_ROLLING_BATCH_SIZE"
+translateTGIToLMI "ENABLE_CUDA_GRAPHS" "OPTION_ENFORCE_EAGER"
```
siddvenk, line 50: This one needs to be translated to the opposite value, right? If enable_cuda_graphs = true, enforce_eager = false

sindhuvahinis: Thanks Sid, https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/llm.py#L70 You are right,

sindhuvahinis: Changed it

lanking520 (approved):

siddvenk (approved):

PR #2019, opened 2024-06-04
Add ecr login credentials

vinayburugu:
## Description ##
Add ecr login credentials

- If this change is a backward incompatible change, why must this change be made? This is backward compatible


.github/workflows/instant_benchmark.yml
```diff
@@ -31,6 +31,10 @@ on:
         description: 'The container used to run benchmark (overrides the template). Should be a full docker path such as deepjavalibrary/djl-serving:0.28.0-lmi'
         required: false
         default: ''
+      ecr-login:
```
zachgk, line 34: This is only in workflow_dispatch, but would also have to be in workflow_call to be used from other workflows

lanking520 (approved):

PR #2038, opened 2024-06-07
AutoAWQ Integration Script

a-ys:
## Description ##

This PR introduces:
1. A "quantize" handler in the djl_serving.huggingface handler that uses AutoAWQ to quantize a model.
3. A code path in the serving partitioning scripts that allows DIY users to run quantization.
    - This is enabled when a user passes the `--quantization awq` option when running the partition script or `option.quantize=awq` in serving.properties
5. A Neo handler script that allows Neo to run quantization through Neo's expected interface.

## Note on serving tensor_parallel_degree
For Llama-2-7b, the tp_degree is limited to 1,2 based on vLLM AWQ implementation. The reason is the tp_degree must satisfy 
```
(intermediate_size / tp_degree) % group_size == 0
```
Where:
- llama-2-7b `intermediate_size = 11008`
- `group_size = 128` which is defined in quant_config in djl_serving.huggingface.quantize()

### Llama-2-7b (Working)
This feature has been tested with Llama-2-7b:
Quantization command:
```
docker run -it --rm \
        -v ./llama-2-7b:/opt/ml/input/data/training \
        -v ./logs:/opt/djl/logs \
        -v ./output:/opt/djl/output \
        --runtime=nvidia \
        --shm-size=12gb \
        deepjavalibrary/djl-serving:lmi-nightly partition --save-mp-checkpoint-path /opt/djl/output --skip-copy
```
Serving.properties:
```
engine=MPI
option.tensor_parallel_degree=8
option.quantize=awq
```
The outputted model was loaded and served with the LMI container.

### Llama-2-70b (Not passing)
Quantization currently failing with 
```
AssertionError
    model_service.invoke_handler("quantize", inputs)
  File "/tmp/djlserving/cache/djl_python/service_loader.py", line 29, in invoke_handler
    return getattr(self.module, function_name)(inputs)
  File "/tmp/djlserving/cache/djl_python/huggingface.py", line 607, in quantize
    _service.quantize(inputs.get_properties())
  File "/tmp/djlserving/cache/djl_python/huggingface.py", line 555, in quantize
    awq_model.quantize(self.tokenizer, quant_config=quant_config)
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/awq/models/base.py", line 186, in quantize
    self.quantizer.quantize()
  File "/usr/local/lib/python3.10/dist-packages/awq/quantize/quantizer.py", line 156, in quantize
    scales_list = [
  File "/usr/local/lib/python3.10/dist-packages/awq/quantize/quantizer.py", line 157, in <listcomp>
    self._search_best_scale(self.modules[i], **layer)
  File "/usr/local/lib/python3.10/dist-packages/torch/utils/_contextlib.py", line 115, in decorate_context
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/awq/quantize/quantizer.py", line 277, in _search_best_scale
    best_scales = self._compute_best_scale(
  File "/usr/local/lib/python3.10/dist-packages/awq/quantize/quantizer.py", line 334, in _compute_best_scale
    self.pseudo_quantize_tensor(fc.weight.data)[0] / scales_view
  File "/usr/local/lib/python3.10/dist-packages/awq/quantize/quantizer.py", line 69, in pseudo_quantize_tensor
    assert torch.isnan(w).sum() == 0
```
Serving.properties:
```
engine=MPI
option.tensor_parallel_degree=8
option.quantize=awq
```


serving/docker/lmi.Dockerfile
```diff
@@ -100,7 +100,7 @@ RUN pip3 install torch==${torch_version} torchvision==${torch_vision_version} --
     transformers==${transformers_version} hf-transfer zstandard datasets==${datasets_version} \
     mpi4py sentencepiece tiktoken blobfile einops accelerate==${accelerate_version} bitsandbytes==${bitsandbytes_version} \
     optimum==${optimum_version} auto-gptq==${auto_gptq_version} pandas pyarrow jinja2 \
-    opencv-contrib-python-headless safetensors scipy onnx sentence_transformers ${onnxruntime_wheel} && \
+    opencv-contrib-python-headless safetensors scipy onnx sentence_transformers ${onnxruntime_wheel} autoawq && \
```
lanking520, line 103: can you make this stick to a verion that match the transformers and other library?

serving/docker/dockerd-entrypoint-with-cuda-compat.sh
```diff
@@ -71,6 +80,9 @@ if [[ "$1" = "serve" ]]; then
 elif [[ "$1" = "partition" ]] || [[ "$1" = "train" ]]; then
     shift 1
     /usr/bin/python3 /opt/djl/partition/partition.py "$@"
+elif [[ "$1" = "quantize" ]]; then
```
lanking520, line 83: why did we introduce this? Could we do everything in partition?

serving/docker/partition/sm_neo_utils.py
```diff
@@ -32,7 +32,7 @@ def get_neo_env_vars():
     Get environment variables required by the SageMaker Neo interface
     """
     try:
-        compiler_options = os.environ["COMPILER_OPTIONS"]
+        compiler_options = os.environ.get("COMPILER_OPTIONS")
```
lanking520, line 35: please provide a default to getter method

engines/python/setup/djl_python/huggingface.py
```diff
@@ -540,6 +540,36 @@ def _read_model_config(self, model_config_path: str):
                 f"This is required for loading huggingface models")
             raise e
 
+    def quantize(self, properties: dict):
```
lanking520, line 543: why we add this in HF.py? We don't offer JIT for AWQ model conversion for vLLM

lanking520 (commented):
Please make sure you have some CI testing in place to make sure those functions are working

serving/docker/dockerd-entrypoint-with-cuda-compat.sh
a-ys: Sorry, this is hanging from previous implementation. Will remove.

serving/docker/partition/sm_neo_utils.py
a-ys: Intended behavior is for this to return `None` if no `COMPILER_OPTIONS` are passed since compiler_options will be optional in Neo. dict.get() already returns None by default. Should I make explicit with 
```
.get("COMPILER_OPTIONS", None)
```
?

engines/python/setup/djl_python/huggingface.py
a-ys: Should I instead create a new handler that is only used for AOT? The benefit with the HF.py handler is being able to use the model/property initialization methods offered with the HuggingFaceService.

serving/docker/partition/properties_manager.py
```diff
@@ -152,6 +158,14 @@ def set_and_validate_entry_point(self):
                     engine = self.properties.get('engine')
                     if engine is None:
                         raise ValueError("Please specify engine")
+                    elif quantize:
+                        if engine.lower() == "mpi":
+                            entry_point = "djl_python.huggingface"
```
sindhuvahinis, line 163: We could do assign entrypoint to huggingface, only when vllm or lmi-dist SERVING_FEATURES env variable.

engines/python/setup/djl_python/huggingface.py
sindhuvahinis: Had a discussion with Frank and Qing regarding quantize or partition. Your design is very similar to partition which is fine, but the user can bring in their own partition  script, but  we dont want to make quantize entryPoint public like that. So let's remove quantize from `huggingface.py `and move to `partition.py`. It does not have to call any handler at all. In this case you dont need `run_partition.py` script at all.

Just make sure, you have all the dependencies your partition script. (python engine jar extraction implementation is already there, so you should be fine). You dont have to use HuggingFaceService at all, move `_init_model` and `_init_tokenizer` `utils.py` . You can still make use of HuggingFaceProperties. 

Or if you don't want to do the refactoring, leave a TODO and write your own initialization of model and tokenizer, we can take of the refactoring later.

a-ys:
Update: these last few commits include:
- Refactoring quantization out of HuggingFace.py and into partition.py.
- envvar configuration support for partitioning, using existing functionality from trt_llm_partition.py
- bugfix for loading .bin files

serving/docker/partition/sm_neo_quantize.py
```diff
@@ -0,0 +1,115 @@
...
+        self.INPUT_MODEL_DIRECTORY: Final[str] = env[1]
+        self.OUTPUT_MODEL_DIRECTORY: Final[str] = env[2]
+        self.COMPILATION_ERROR_FILE: Final[str] = env[3]
+        self.COMPILER_CACHE_LOCATION: Final[str] = env[4]
+
+    def update_dataset_cache_location(self):
+        logging.info(
+            f"Updating HuggingFace Datasets cache directory to: {self.COMPILER_CACHE_LOCATION}"
+        )
+        os.environ['HF_DATASETS_CACHE'] = self.COMPILER_CACHE_LOCATION
```
sindhuvahinis, line 49: What if compiler_cache_location is None. Is that okay? Do we validate them somewhere?

serving/docker/partition/properties_manager.py
```diff
@@ -17,20 +17,24 @@
...
+
 
 class PropertiesManager(object):
 
     def __init__(self, args, **kwargs):
         self.entry_point_url = None
         self.properties_dir = args.properties_dir
-        self.properties = load_properties(self.properties_dir)
+        self.properties = update_kwargs_with_env_vars({})
+        self.properties |= (load_properties(self.properties_dir))
```
sindhuvahinis, line 37: Let's switch the order. Load the properties first and then read from the env variable. That's our flow in model server for inference. Lets do the same for AOT as well

serving/docker/partition/sm_neo_quantize.py
```diff
@@ -0,0 +1,115 @@
...
+        self.properties_manager = PropertiesManager(self.args)
+
+    def run_quantization(self) -> str:
+        """
+        :return: the output of the partition command captured from stdout
+        """
+        partition_service = PartitionService(self.properties_manager)
+        extract_python_jar(PYTHON_CACHE_DIR)
+        try:
+            return partition_service.run_partition()
```
sindhuvahinis, line 87: Isnt  `partition_service.run_quantization()` here?

serving/docker/partition/sm_neo_quantize.py
a-ys: We validate here: https://github.com/deepjavalibrary/djl-serving/blob/6ec92723030c7fc70021308a09549a410c8ef29b/serving/docker/partition/sm_neo_utils.py#L30

serving/docker/partition/sm_neo_quantize.py
a-ys: Yes, thank you.

sindhuvahinis (approved):

PR #2046, opened 2024-06-11
[IB] support suzuka model

lanking520:
## Description ##

Add suzuka model to message reporting

### DIFF
--- a/tests/integration/record_benchmark.py
+++ b/tests/integration/record_benchmark.py
@@ -171,10 +171,8 @@ def data_container(data: dict):
             version = container.split(":")[1]
             if not version.startswith("sha"):
                 data["tgiVersion"] = version
-        elif "suzuka" in container:
-            data["image"] = "suzuka"
         else:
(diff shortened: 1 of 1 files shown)


tests/integration/record_benchmark.py
```diff
@@ -166,11 +166,15 @@ def data_container(data: dict):
             else:  # Nightly build
                 data["djlNightly"] = "true"
                 data["image"] = container[:-len("-nightly")]
-        if "text-generation-inference" in container:
+        elif "text-generation-inference" in container:
             data["modelServer"] = "TGI"
             version = container.split(":")[1]
             if not version.startswith("sha"):
                 data["tgiVersion"] = version
+        elif "suzuka" in container:
```
zachgk, line 174: Rather than having `sazuka` and `other`, why not use `container.split("/", 1)[1]`? It will be a solid value that will work for sazuka and future use cases as well

zachgk (approved):

PR #2052, opened 2024-06-12
[test] remove trtllm flan t5 integration test

sindhuvahinis:
## Description ##

Removing this till, triton support in 0.10.0 is supported.

### DIFF
--- a/.github/workflows/llm_integration.yml
+++ b/.github/workflows/llm_integration.yml
@@ -341,17 +341,6 @@ jobs:
           python3 llm/client.py trtllm qwen-7b
           rm -rf docker_env
           docker rm -f $(docker ps -aq)
-      - name: flan-t5-xxl pre-compiled model with python backend
-        working-directory: tests/integration
-        run: |
-          rm -rf models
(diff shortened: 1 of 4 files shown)


.github/workflows/llm_integration.yml
```diff
@@ -341,17 +341,6 @@ jobs:
           python3 llm/client.py trtllm qwen-7b
           rm -rf docker_env
           docker rm -f $(docker ps -aq)
-      - name: flan-t5-xxl pre-compiled model with python backend
-        working-directory: tests/integration
-        run: |
-          rm -rf models
-          echo -en "CUDA_VISIBLE_DEVICES=0,1,2,3" > docker_env
-          python3 llm/prepare.py trtllm flan-t5-xxl
```
lanking520, line 349: could we also remove the test in the python files?

sindhuvahinis: Spared those because if we added support in the next week, it would be easier to add. Never mind, will remove those too.

sindhuvahinis: Done removed those too

siddvenk (approved):

PR #2058, opened 2024-06-13
[secure-mode] Refactor secure mode plugin

frankfliu:
1. Add jacoco report
2. improve code coverage
3. Remove unrelated features

## Description ##

### DIFF
--- a/plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeUtils.java
+++ b/plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeUtils.java
@@ -12,7 +12,8 @@
  */
 package ai.djl.serving.plugins.securemode;
 
-import ai.djl.ModelException;
+import ai.djl.serving.http.IllegalConfigurationException;
+import ai.djl.serving.wlm.ModelInfo;
 import ai.djl.util.JsonUtils;
 import ai.djl.util.Utils;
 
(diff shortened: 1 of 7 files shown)


plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeUtils.java
```diff
@@ -55,324 +51,129 @@ public final class SecureModeUtils {
...
-    /**
-     * Search for any pickle-based file in the directory, and fast-fail if found.
-     *
-     * @param directory the directory to search for pickle-based files
-     * @throws IOException if there is an error walking the directory
-     * @throws ModelException if a security check fails
-     */
-    private static void scanForPickle(Path directory) throws IOException, ModelException {
-        Pattern pattern = Pattern.compile(PICKLE_EXTENSIONS_REGEX);
+        Set<String> controls = new HashSet<>(Arrays.asList(securityControls.split("\\s+,\\s*")));
```
ethnzhng, line 95: I don't understand this regex, doesn't this require a whitespace before the comma? Ideally we just split around commas and then trim whitespace from each member

frankfliu: Good catch, fixed.

sindhuvahinis (approved):

PR #2072, opened 2024-06-17
[fix] initialize sequence dictionary for default sequence index to pr…

siddvenk:
… in an output

## Description ##

A previous change caused issues with a corner case. The issue was introduced by https://github.com/deepjavalibrary/djl-serving/commit/06eaa98518bb951ae1d7cad1e8dc6787082d382d#diff-4de063c68933067104c13e58e380b04bb922ca889f070e40bfa743ab33556426.

When we call `LLMEngine.step()` (through vllm or lmi-dist), it's not guaranteed that we receive and output for every active request. If on the first call to step for a request we do not generate a token, we never initialize the sequence dict that is part of TextGenerationOutput. Later in the flow we make a call to `request.get_next_token` https://github.com/deepjavalibrary/djl-serving/blob/1e883d5bae81a20189847cc4b2aff93ac38b8a38/engines/python/setup/djl_python/request.py#L140. For requests that did not generate a new token via step, we never call `set_new_token`, which is where the sequences property of TextGenerationOutput gets initialized.

The simples fix is to ensure that the sequences property of TextGenerationOutput always gets created for the default index so that we never try to access an array index that doesn't exist.

Fixes:
- inf2 unit tests

**Note that for some reason the generated text for one of the inputs has changed. the outputs seem fine, but i cannot figure out why this changed**

### DIFF
--- a/engines/python/setup/djl_python/request_io.py
+++ b/engines/python/setup/djl_python/request_io.py
@@ -178,6 +178,9 @@ class TextGenerationOutput(RequestOutput):
     prompt_tokens_details: List[Token] = field(default_factory=lambda: [])
     other_sequences_indices: List[int] = field(default_factory=lambda: [])
 
+    def __post_init__(self):
+        self.sequences[self.best_sequence_index] = Sequence()
+
     def set_next_token(self,
                        token: Token,
                        sequence_index=0,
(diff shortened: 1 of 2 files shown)


sindhuvahinis:
Thanks Sid for this fix. A couple of questions:
- https://github.com/deepjavalibrary/djl-serving/commit/06eaa98518bb951ae1d7cad1e8dc6787082d382d#diff-667c72a39acfc41b10ead3452cf4863e177dfb8f2fa4c3e7cf4c90656559970cR81 Why did this line not take care of this? 
- We also construct the token, if the new tokens are not generated. https://github.com/deepjavalibrary/djl-serving/commit/06eaa98518bb951ae1d7cad1e8dc6787082d382d#diff-667c72a39acfc41b10ead3452cf4863e177dfb8f2fa4c3e7cf4c90656559970cR140 
- So does it mean, LLMEngine.step() itself returns None or something?

sindhuvahinis (approved):

PR #2113, opened 2024-06-26
[serving] introduce interface to customize http response status retur…

siddvenk:
This PR introduces an interface, HttpResponseStatusMapper that exposes a single method that accepts a Throwable and returns an HttpResponseStatus.

This will allow users to customize the behavior of responses returned to the client for certain error conditions.

It also introduces an HttpResponseStatusProvider. This object allows plugins to register their implementations of HttpResponseStatusMapper with DJLServing and use them in the inference request handler

### DIFF
--- a/serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
+++ b/serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
@@ -24,6 +24,8 @@ import ai.djl.serving.cache.CacheEngine;
 import ai.djl.serving.cache.CacheManager;
 import ai.djl.serving.models.ModelManager;
 import ai.djl.serving.util.ConfigManager;
+import ai.djl.serving.util.HttpResponseStatusMapper;
+import ai.djl.serving.util.HttpResponseStatusMapperProvider;
 import ai.djl.serving.util.NettyUtils;
 import ai.djl.serving.wlm.ModelInfo;
 import ai.djl.serving.wlm.util.WlmException;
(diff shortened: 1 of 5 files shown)


zachgk:
Is there a motivating use case for this change? Do we expect different users to require different mappings, especially to the extent of creating custom ones. Or do we expect it to mostly be first-party `HttpResponseStatusMapper`s?

serving/src/main/java/ai/djl/serving/util/DefaultHttpResponseStatusMapper.java
```diff
@@ -0,0 +1,37 @@
...
+package ai.djl.serving.util;
+
+import ai.djl.serving.http.BadRequestException;
+import ai.djl.serving.wlm.util.WlmException;
+import ai.djl.translate.TranslateException;
+
+import io.netty.handler.codec.http.HttpResponseStatus;
+
+/** Default converter translating exception to HttpResponseStatus. */
+public class DefaultHttpResponseStatusMapper implements HttpResponseStatusMapper {
```
frankfliu, line 22: I feel this is a bit over engineering.

We can just add a configuration like: `bad_request=424,time_out=408,wlm_error=503,server_error=500`. We don't really need a plugin to inject a StatusMapper

serving/src/main/java/ai/djl/serving/http/InferenceRequestHandler.java
```diff
@@ -468,24 +474,21 @@ void sendOutput(Output output, ChannelHandlerContext ctx) {
     }
 
     void onException(Throwable t, ChannelHandlerContext ctx) {
-        HttpResponseStatus status;
+        HttpResponseStatus status = httpResponseStatusMapper.getHttpStatusForException(t);
         if (t instanceof TranslateException || t instanceof BadRequestException) {
             logger.debug(t.getMessage(), t);
             SERVER_METRIC.info("{}", RESPONSE_4_XX);
-            status = HttpResponseStatus.BAD_REQUEST;
         } else if (t instanceof WlmException) {
```
frankfliu, line 481: We might need to introduce a TimeoutException (extends EngineException) and implement this in PyProcess.
Currently if predict timeout, we throw EngineException. We might need to see how can we differentiate timeout vs normal TranslateException

PR #2122, opened 2024-06-28
[python] add support for 3p use-case

siddvenk:
## Description ##

1. Adds input/output formatters that conform to the 3p spec
2. Auto detect 3p request based on custom Header and value


sindhuvahinis:
What does 3p mean here?

sindhuvahinis (approved):

engines/python/setup/djl_python/three_p/three_p_utils.py
```diff
@@ -0,0 +1,22 @@
+from typing import Optional
```
sindhuvahinis, line 1: Add license header.

siddvenk: good catch

PR #2129, opened 2024-06-28
[secure-mode] Add properties allowlist validation

ethnzhng:


siddvenk:
I'd prefer a denylist instead of an allowlist With this, new configs are not allowed by default (maybe that's the desired behavior), but I feel like there are going to be new configs added that should be allowed more often than new configs added that are disallowed.

plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeUtils.java
```diff
@@ -95,10 +95,28 @@ public static void validateSecurity(ModelInfo<?, ?> modelInfo) throws IOExceptio
...
 
+    /**
+     * In Secure Mode we only allow an explicit set of DJL-Serving properties to be set.
+     *
+     * @param modelInfo ModelInfo of the model
+     * @param allowedKeys set of allowlisted properties
+     */
+    private static void validateProperties(ModelInfo<?, ?> modelInfo, Set<String> allowedKeys) {
+        Properties prop = modelInfo.getProperties();
+        Set<String> allowedKeysSet = new HashSet<>(allowedKeys);
```
siddvenk, line 111: this is redundant, allowedKeys is already a set.

plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeAllowList.java
```diff
@@ -0,0 +1,66 @@
...
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.serving.plugins.securemode;
+
+import java.util.Set;
+
+/** A class for properties that are allowlisted in Secure Mode. */
+final class SecureModeAllowList {
```
frankfliu, line 18: Make this a interface

plugins/secure-mode/src/main/java/ai/djl/serving/plugins/securemode/SecureModeAllowList.java
```diff
@@ -0,0 +1,66 @@
...
+
+    public static final Set<String> PROPERTIES_ALLOWLIST =
+            Set.of(
+                    "engine",
+                    "job_queue_size",
```
frankfliu, line 23: Add the following to whitelist:
1. max_idle_time
2. batch_size
3. max_batch_delay

sindhuvahinis: Yeah, maybe minWorkers and maxWorkers?

lanking520 (approved):

frankfliu (approved):

PR #2134, opened 2024-06-28
[CI] Add trust remote code option in TRTLLM CI

ydm-amazon:
Fixes the following CI error:

```
ValueError: The repository for ccdv/cnn_dailymail contains custom code which must be executed to correctly load the dataset. You can inspect the repository content at https://hf.co/datasets/ccdv/cnn_dailymail.
```

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -791,6 +791,7 @@ trtllm_handler_list = {
         "option.smoothquant_per_channel": "True",
         "option.rolling_batch": "trtllm",
         "option.output_formatter": "jsonlines",
+        "option.trust_remote_code": True
     },
     "internlm-7b": {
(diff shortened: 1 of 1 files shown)


sindhuvahinis:
No this is not it. This error is because of dataset version in HuggingFace. We dont need trust_remote_code

tests/integration/llm/prepare.py
```diff
@@ -791,6 +791,7 @@
         "option.smoothquant_per_channel": "True",
         "option.rolling_batch": "trtllm",
         "option.output_formatter": "jsonlines",
+        "option.trust_remote_code": True
```
lanking520, line 794: this might not pass through to actual compilation/quantization script in trtllm

ydm-amazon:
Closed PR since the root cause seems to be elsewhere

PR #2145, opened 2024-07-03
[python] refactor input parser to support Request

sindhuvahinis:
## Description ##

Refactoring input parser to support Request. Before this PR, we loop through the list of requests to list of input_text, input_size, parameters etc.. and then in rolling batch, we loop through through these parsed list and then convert to list of requests `Request` again. This PR aims to avoid this duplicate work. 

## Assumptions ## 
1. Before this PR, we duplicated the parameters for client side batching, which is not needed anymore as we no longer maintain this as list, we maintan this a list of Requests. 
2. In this PR, we assume, if adapter_registry is non empty, then adapters needs to be looked in the requests. If adapter_registry is empty, then we dont look for adapters. 
3. In this PR, we introduced server_parameters => which will have the server modified parameters and the built-in handlers should look for this in order to modify or send to backend engines like vllm. 

## After this PR, for future improvements ##
1.  For the new standard of input_formatter, we would want request_input and part of the input_formatter. This refactor makes this easier.
2. For mulitmodal parsing, this should also makes thing easier. 
3. output_formatter could be easily used by dynamic batching use cases as well. This will unify the API UX for rolling batch and dynamic batching. 

## Testing ##
- https://github.com/deepjavalibrary/djl-serving/actions/runs/9862485646/job/27233337263 
- There is a failure LMIDIst2 and vllm LoRA, will investigate in my local machine, but others are passing now. 

P.S. I could not divide this into multiple PRs, sorry about that. All these changes has to go in one PR.

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -23,17 +23,18 @@ from transformers import (pipeline, Pipeline, AutoModelForCausalLM,
                           StoppingCriteriaList)
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
 from peft import PeftConfig, PeftModel, PeftModelForCausalLM
-from typing import Tuple, List
+from typing import List, Dict
 
 from djl_python.encode_decode import encode
 from djl_python.inputs import Input
 from djl_python.outputs import Output
+from djl_python.request_io import RequestInput
 from djl_python.streaming_utils import StreamingUtils
 
 from djl_python.properties_manager.properties import StreamingEnum, is_rolling_batch_enabled, is_streaming_enabled
 from djl_python.properties_manager.hf_properties import HuggingFaceProperties
-from djl_python.utils import rolling_batch_inference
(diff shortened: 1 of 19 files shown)


serving/docker/lmi.Dockerfile
```diff
@@ -106,7 +106,7 @@ RUN pip3 install torch==${torch_version} torchvision==${torch_vision_version} --
     pip3 install ${djl_converter_wheel} --no-deps && \
     pip3 cache purge
 
-RUN pip3 install ${flash_attn_2_wheel} ${lmi_dist_wheel} ${vllm_wheel} pydantic==${pydantic_version} && \
+RUN pip3 install ${flash_attn_2_wheel} ${lmi_dist_wheel} vllm==0.5.0.post1 pydantic==${pydantic_version} && \
```
sindhuvahinis, line 109: Added this to unblock integration testing. Will remove once the testing is done.

frankfliu (approved):

siddvenk (approved):

PR #2148, opened 2024-07-07
[engine] include lmi recommended entrypoint when model.py exists

sindhuvahinis:
## Description ##

Existing flow: If model.py exists in the model directory, without explicitly set by the user either in serving.properties or environment variables, then we take that as the entryPoint. 

This PR:
- Still takes model.py as the entryPoint.
- But we also send the inferred built in handler as the recommended_entry_point to the python engine. 
- In the python side, we check whether the given entry point has the `handle` or handler function name. If it does not exist, then we use the recommended_entry_point as the entry point

Tests:
- Manually tested in my ec2 machine.

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyModel.java
@@ -77,6 +77,7 @@ public class PyModel extends BaseModel {
                     "Python engine does not support dynamic blocks");
         }
         String entryPoint = null;
+        String recommendedEntryPoint = null;
         if (options != null) {
             logger.debug("options in serving.properties for model: {}", modelName);
             for (Map.Entry<String, ?> entry : options.entrySet()) {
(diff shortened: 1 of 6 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyModel.java
```diff
@@ -152,19 +153,22 @@ public void load(Path modelPath, String prefix, Map<String, ?> options) throws I
         if (entryPoint == null) {
             entryPoint = Utils.getenv("DJL_ENTRY_POINT");
             if (entryPoint == null) {
-                Path modelFile = findModelFile(prefix);
```
frankfliu, line 155: The following seems more readable:
```
                Path modelFile = findModelFile(prefix);
                String features = Utils.getEnvOrSystemProperty("SERVING_FEATURES");
                // find default entryPoint
                if (modelFile != null) {
                    entryPoint = modelFile.toFile().getName();
                }
                // find recommendedEntryPoint
                if ("nc".equals(manager.getDevice().getDeviceType())
                    recommendedEntryPoint = "djl_python.transformers_neuronx";
                } else if ("trtllm".equals(features)) {
                    recommendedEntryPoint = "djl_python.tensorrt_llm";
                } else if (pyEnv.getInitParameters().containsKey("model_id")
                        || Files.exists(modelPath.resolve("config.json"))) {
                    recommendedEntryPoint = "djl_python.huggingface";
                }
                if (entryPoint == null && recommendedEntryPoint == null) {
                    throw new FileNotFoundException(".py file not found in: " + modelPath);
                }
 ```

frankfliu (approved):

PR #2152, opened 2024-07-09
[lmi] remove redundant auto logic from python handler

siddvenk:
## Description ##

There is quite a bit of old/outdated logic scattered throughout djl_python and our default handlers. I'll be raising PRs to slowly clear out this outdated logic.

This change removes the "auto" logic for rolling batch from out python code. This logic has been fully moved to the LmiConfigRecommender on the java side. The logic here is not up to date anymore either.

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -53,16 +53,6 @@ ARCHITECTURES_2_TASK = {
     "BloomModel": "text-generation",
 }
 
-LMI_DIST_ADV_MODEL = {
-    "RWForCausalLM",
-    "GPTNeoXForCausalLM",
-    "T5ForConditionalGeneration",
-    "LlamaForCausalLM",
-    "FalconForCausalLM",
-    "MPTForCausalLM",
-    "GPTBigCodeForCausalLM",
-}
-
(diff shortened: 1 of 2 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -85,17 +75,8 @@ def enable_flash():
...
-    if rolling_batch_type == "auto":
-        architecture = model_config.architectures[0]
-        if architecture in LMI_DIST_ADV_MODEL and is_mpi:
-            from djl_python.rolling_batch.lmi_dist_rolling_batch import LmiDistRollingBatch
-            return LmiDistRollingBatch
-        else:
-            from djl_python.rolling_batch.scheduler_rolling_batch import SchedulerRollingBatch
-            return SchedulerRollingBatch
-    elif rolling_batch_type == "scheduler":
+def get_rolling_batch_class_from_str(rolling_batch_type: str):
```
lanking520, line 78: we should have deprecation warning or a fallback for auto. There are still some customer setting that to auto today.

siddvenk: "auto" is still valid, but we just handle it here instead https://github.com/deepjavalibrary/djl-serving/blob/master/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java#L81.

I don't think the logic here is ever invoked for "auto" in the current state.

lanking520 (approved):

PR #2158, opened 2024-07-10
[python] add input formatter decorator

sindhuvahinis:
## Description ##
This PR introduces a new decorator `input_formatter`. Users are expected to annotate this in their input_formatter function in `model.py`. 

This PR also expects the user to define their input_formatter with signature 
```
def my_custom_input_formatter(input_item: Input, **kwargs) -> RequestInput:
```

Example of UX for users writing their own input formatter:

```
from djl_python import Input
from djl_python.input_parser import input_formatter
from djl_python.request_io import TextInput
from djl_python.encode_decode import decode

## **kwargs -> has most of the class attributes of TRTLLMService/HuggingFaceService
## example, for TRTLLMService, it has configs (serving.properties), tokenizer, rolling_batch

@input_formatter
def custom_input_formatter(input_item: Input, **kwargs):
    request_input = TextInput()
    content_type = input_item.get_property("Content-Type")
    input_map = decode(input_item, content_type)

    inputs = input_map.pop("inputs", input_map)
    params = input_map.pop("parameters", {})
    
    request_input.input_text = inputs
    request_input.parameters = params
    
    return request_input
```

### DIFF
--- a/engines/python/setup/djl_python/input_parser.py
+++ b/engines/python/setup/djl_python/input_parser.py
@@ -11,16 +11,28 @@
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
 import logging
+import os
 from dataclasses import dataclass, field
-from typing import List, Dict
+from typing import List, Dict, Optional, Callable
 
 from djl_python import Input
 from djl_python.chat_completions.chat_utils import is_chat_completions_request, parse_chat_completions_request
 from djl_python.encode_decode import decode
 from djl_python.properties_manager.properties import is_rolling_batch_enabled
 from djl_python.request import Request
-from djl_python.request_io import TextInput
+from djl_python.request_io import TextInput, RequestInput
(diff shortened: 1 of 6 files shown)


engines/python/setup/djl_python/input_parser.py
```diff
@@ -60,19 +98,24 @@ def parse_input_with_formatter(inputs: Input, **kwargs) -> ParsedInput:
...
-            # TODO: Decide whether it is a text input based on content-type
-            request_input = TextInput(
-                request_id=request_id,
-                tokenizer=kwargs.get("tokenizer"),
-                tgi_compat=kwargs.get("configs").tgi_compat)
-            format_input(request_input, input_item, **kwargs)
+            # input formatter can be user written as well. We look for model.py and search for the decorator.
+            input_formatter_function = get_input_formatter(
+                kwargs.get("configs"))
+            request_input = input_formatter_function(input_item, **kwargs)
```
sindhuvahinis, line 109: Will move this to Service, so we need to scan each and every time.

sindhuvahinis: Done

tosterberg (approved):

PR #2166, opened 2024-07-11
[Neo] Refactor Neo TRT-LLM partition script

ethnzhng:
## Description ##
1. Refactor structure of script to be more in line with `sm_neo_neuron_partition.py`
2. Remove code related to deprecated Neo `compiler_flags`
3. Move generation of `serving.properties` file post-compilation from `tensorrt_llm_toolkit` to here
4. Add support for HF dataset cache env var

### DIFF
--- a/serving/docker/partition/sm_neo_trt_llm_partition.py
+++ b/serving/docker/partition/sm_neo_trt_llm_partition.py
@@ -16,9 +16,10 @@ import os
 import sys
 from typing import Final
 
-from utils import update_kwargs_with_env_vars, load_properties
+from utils import (update_kwargs_with_env_vars, load_properties,
+                   update_dataset_cache_location, remove_option_from_properties)
 from sm_neo_utils import (CompilationFatalError, write_error_to_file, 
-                          get_neo_env_vars, update_dataset_cache_location)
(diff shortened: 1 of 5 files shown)


lanking520 (approved):

serving/docker/partition/sm_neo_trt_llm_partition.py
```diff
@@ -12,104 +12,76 @@
...
-        except Exception as exc:
-            raise CompilationFatalError(
-                f"Encountered an error during TRT-LLM compilation: {exc}")
-
+        neo_trtllm_partition_service = NeoTRTLLMPartitionService()
+        neo_trtllm_partition_service.neo_partition()
     except Exception as exc:
-        write_error_to_file(exc, compilation_error_file)
+        write_error_to_file(
+            exc, neo_trtllm_partition_service.COMPILATION_ERROR_FILE)
```
a-ys, line 84: Small bug here that exists in the other scripts: if this code on line 80 fails
```
neo_trtllm_partition_service = NeoTRTLLMPartitionService()
```
There will be an undefined variable exception since `neo_trtllm_partition_service.COMPILATION_ERROR_FILE` has not been initialized.

PR #2173, opened 2024-07-13
[serving][python] Support non 200 HTTP response codes for non-streami…

siddvenk:
…ng rolling-batch

1. Change non-streaming output formatters to only generate output on final token (valid and error cases)
2. Update InferenceRequestHandler to only publish HTTP Response when data is actually received
3. Update RollingBatch to properly set status code in non-streaming use-cases

Note:
- The ask for supporting non 200 HTTP responses has been asked for be a few users.

TODO:
- I need to update unit tests, but i have done extensive local testing and am comfortable with the core changes raised here. I'll change from WIP to Open once i complete the unit tests.
---

This change is motivated by two things:
- Make output formatters for non streaming use-cases easier to deal with, and more readable
- Provide non-200 HTTP status code in the non-streaming error cases.

The output formatters for non-streaming use-cases are currently pretty hard to reason about. Beyond just making it easier to read and work with, this also fixes an issue:
- If there is an error during generation after we have generated a few tokens, the current behavior results in a response that is not valid json. This change ensure that the output formatters we provide always result in valid json content in mid-generation error scenarios

The second part of this change is to provide non-200 status codes during non-streaming use-cases when an error occurs during inference. This is only possible with the first change to the output formatters. There are two small changes in InferenceRequestHandler and RollingBatch to enable this.

### DIFF
--- a/engines/python/setup/djl_python/tests/test_rolling_batch.py
+++ b/engines/python/setup/djl_python/tests/test_rolling_batch.py
@@ -28,18 +28,14 @@ class TestRollingBatch(unittest.TestCase):
         req2 = Request(req_input2)
         for req in [req1, req2]:
             req.set_next_token(Token(244, "He", -0.334532))
-            print(req.get_next_token(), end='')
-            assert req.get_next_token() == '{"generated_text": "He'
-            req.reset_next_token()
+            req.get_next_token()
             req.set_next_token(Token(576, "llo", -0.123123))
-            print(req.get_next_token(), end='')
-            assert req.get_next_token() == 'llo'
-            req.reset_next_token()
+            req.get_next_token()
             req.set_next_token(Token(4558, " world", -0.567854), True,
                                'length')
             print(req.get_next_token(), end='')
-            assert req.get_next_token() == ' world"}'
-            req.reset_next_token()
+            assert req.get_next_token() == json.dumps(
+                {"generated_text": "Hello world"})
 
     def test_json_fmt_with_appending(self):
         req_input1 = TextInput(request_id=0,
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/tests/test_rolling_batch.py
```diff
@@ -28,18 +28,14 @@ def test_json_fmt(self):
...
-            print(req.get_next_token(), end='')
-            assert req.get_next_token() == 'llo'
-            req.reset_next_token()
+            req.get_next_token()
             req.set_next_token(Token(4558, " world", -0.567854), True,
                                'length')
             print(req.get_next_token(), end='')
-            assert req.get_next_token() == ' world"}'
-            req.reset_next_token()
+            assert req.get_next_token() == json.dumps(
```
lanking520, line 37: self.assertEquals

engines/python/setup/djl_python/utils.py
```diff
@@ -98,7 +98,7 @@ def rolling_batch_inference(parsed_input, inputs: Input, outputs: Output,
             outputs.add(Output.binary_encode(err), key="data", batch_index=i)
             outputs.add_property(f"batch_{i}_Content-Type", "application/json")
         else:
-            content_type = result[idx].pop("content_type")
+            content_type = result[idx].get("content_type")
```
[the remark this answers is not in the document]
lanking520: I would suggest to use get here. Maybe in the future someone would access the content in the later time. Python GC would collect it eventually

engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
```diff
@@ -306,24 +304,6 @@ void addResponse(byte[] json, Map<String, String> properties) {
                 output.getProperties().putAll(properties);
             }
             ++count;
-            if (json[0] == '{') {
```
lanking520, line 309: This may potentially break some legacy custom handler. Maybe mark as deprecate and raise Deprecation warning for this code piece execution for now

engines/python/setup/djl_python/utils.py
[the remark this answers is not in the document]
siddvenk: `pop` does not cause issues currently, but it's incorrectly used here with the subsequent check to see if content-type is None. 

Either we use pop and get rid of that check, or use get here.

engines/python/setup/djl_python/tests/test_rolling_batch.py
siddvenk: let me take this up as a separate change. We don't have consistent usage of assert/self.assert<Condition> across our tests, so i can fix on that across our tests

engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
siddvenk: i'm not sure if this condition is still valid, but i'll add it back with a deprecation warning.

We removed the huggingface.parse_input and replaced it with the input formatter @sindhuvahinis implemented.

I think this is only valid if user does json encoding on the outputs from python side that are received on java side

PR #2187, opened 2024-07-17
[CI] Creates nightly pipeline

zachgk:
This creates a full nightly pipeline for the build, test, and release.

First, it enables a full build through temp repo approach. The build will always go to the temp repo in docker_build.yml, and then the separate docker_publish.yml can be used for the actual release step.

Following this, it also enables the integration suite to test the temp repo similarly to the integration runner.

Finally, a new action "Nightly Pipeline" is enabled to pull it all together. Right now, it will publish regardless of the result of the test. This matches the current behavior. When it changes in the future, the testing can be enabled through un-commenting the line.

@tosterberg

### DIFF
--- a/.github/workflows/docker_build.yml
+++ b/.github/workflows/docker_build.yml
@@ -1,17 +1,19 @@
-name: OLD Build and push docker nightly
-# For use in building CVE patches for branches <= 0.28.0
+name: Docker Build
 
 on:
   workflow_dispatch:
     inputs:
       mode:
-        description: 'release/nightly/temp, default is nightly'
+        description: 'release/nightly, default is nightly'
         required: true
         default: 'nightly'
+        options:
+          - nightly
+          - release
   workflow_call:
     inputs:
       mode:
-        description: 'release/nightly/temp, default is nightly'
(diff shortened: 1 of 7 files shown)


.github/workflows/docker_build.yml
```diff
@@ -1,21 +1,22 @@
-name: Build and push docker nightly
+name: Docker Build
```
lanking520, line 1: we might still want to keep the workflow, simply duplicate it. This is for the previous branch (DLC) to use the workflow

zachgk: Yeah. Another option which is more consistent but makes more changes is to cherry-pick this PR to the supported DLC branches. Then they will follow a similar process so we don't have to maintain both in docs. I'm fine with either if you have a strong preference

tosterberg:
I am good with the change, we can discuss approval and merge after 0.29.0 release has stabilized. My vote is that we set a task for the approval, merge, and cherry pick of this PR as well as the unified integration tests PR into the 0.26.0 branch through 0.29.0, and add a documentation maintenance step to update the related runbooks with the new action names and process.

PR #2188, opened 2024-07-17
[ci] fix adapters client testing

tosterberg:
## Description ##

Now that response codes are coming back with non-200 values the `send_json` functions error handling is now occurring prior to and outside of the adapters test handler. Refactored to allow `send_json` to throw errors, and handle catches appropriately.

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -13,6 +13,8 @@ import numpy as np
 from datetime import datetime
 from io import BytesIO
 
+FAILED_DEPENDENCY_CODE = 424
+
 logging.basicConfig(level=logging.INFO)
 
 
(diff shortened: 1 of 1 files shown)


tests/integration/llm/client.py
```diff
@@ -1078,16 +1078,19 @@ def test_handler_adapters(model, model_spec):
     res = requests.delete(
         f"http://127.0.0.1:8080/models/test/adapters/{del_adapter}")
     logging.info(f"del adapter {res}")
-    res = send_json(reqs[0]).content.decode("utf-8")
-    logging.info(f"call deleted adapter {res}")
-    if "error" not in res:
+
+    try:
+        # Following send_json should fail with a 424 for calling a deleted adapter
+        res = send_json(reqs[0]).content.decode("utf-8")
```
lanking520, line 1084: could we check the response status code?

tests/integration/llm/client.py
```diff
@@ -1078,16 +1078,19 @@ def test_handler_adapters(model, model_spec):
...
+        logging.info(f"call deleted adapter {res}")
+    except ValueError:
```
siddvenk, line 1086: I think we should change the logic in send_json, or call requests.post directly here so that we can directly validate the response status and code. It's possible that the request fails for another reason/status here, but still passes.

tests/integration/llm/client.py
tosterberg: Updated to an assert pulling the error code

tests/integration/llm/client.py
tosterberg: Updated to narrow the error conditions

zachgk (approved):

PR #2190, opened 2024-07-17
[serving] Adds mutliple node cluster configuration support

frankfliu:
## Description ##

### DIFF
--- a/serving/docker/config.properties
+++ b/serving/docker/config.properties
@@ -1,5 +1,6 @@
 inference_address=http://0.0.0.0:8080
 management_address=http://0.0.0.0:8080
+cluster_address=http://0.0.0.0:8888
 model_store=/opt/ml/model
 load_models=ALL
 #model_url_pattern=.*
--- a/serving/src/main/java/ai/djl/serving/util/ClusterConfig.java
+++ b/serving/src/main/java/ai/djl/serving/util/ClusterConfig.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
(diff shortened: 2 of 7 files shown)


serving/src/main/java/ai/djl/serving/util/ConfigManager.java
```diff
@@ -238,8 +239,10 @@ public Connector getConnector(Connector.ConnectorType type) {
         String binding;
         if (type == Connector.ConnectorType.MANAGEMENT) {
             binding = prop.getProperty(MANAGEMENT_ADDRESS, "http://127.0.0.1:8080");
-        } else {
+        } else if (type == Connector.ConnectorType.INFERENCE) {
             binding = prop.getProperty(INFERENCE_ADDRESS, "http://127.0.0.1:8080");
+        } else {
+            binding = prop.getProperty(CLUSTER_ADDRESS, "http://127.0.0.1:8888");
```
zachgk, line 245: This isn't correct for `ConnectorType.BOTH`

serving/src/main/java/ai/djl/serving/util/ClusterConfig.java
```diff
@@ -0,0 +1,87 @@
...
+public final class ClusterConfig {
+
+    private static final ClusterConfig INSTANCE = new ClusterConfig();
+
+    private int clusterSize;
+    private CountDownLatch latch;
+    private String error;
+
+    private ClusterConfig() {
+        clusterSize = Integer.parseInt(Utils.getenv("DJL_CLUSTER_SIZE", "1"));
```
zachgk, line 30: It might be slightly nicer to configure this using the config manager rather than just environment variables

frankfliu: 1. This is configured by LWS, and it only support env var so far
2. There are other code (PyEnv) is also using just env var

serving/src/main/java/ai/djl/serving/util/ConfigManager.java
frankfliu: I will change to switch case to make it more readable

serving/docker/config.properties
```diff
@@ -1,5 +1,6 @@
 inference_address=http://0.0.0.0:8080
 management_address=http://0.0.0.0:8080
+cluster_address=http://0.0.0.0:8888
```
nskool, line 3: Nit: any reason for choosing 8888? Should it be a closer port like 8081?

frankfliu: User may choose management port as 8081

lanking520 (approved):

PR #2191, opened 2024-07-17
add trtllm container update

lanking520:
## Description ##

Work in progress, will test a build

https://github.com/deepjavalibrary/djl-serving/actions/runs/9984782888


serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -68,7 +68,7 @@ COPY distribution[s]/ ./
 RUN mv *.deb djl-serving_all.deb || true
 
 # Install CUDNN 8
-RUN apt-get update && apt-get install -y --no-install-recommends libcudnn8 && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y --no-install-recommends libcudnn9-cuda-12 && rm -rf /var/lib/apt/lists/*
```
frankfliu, line 71: Why not just use `12.4.1-cudnn-devel-ubuntu22.04` image directly?

lanking520: it also installed libcudnn9-dev-cuda-12, which is around 1GB we don't need that

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -15,20 +15,20 @@ ARG cuda_version=cu124
...
-ARG trtllm_toolkit_version=0.10.0%2Bnightly
-ARG trtllm_version=v0.10.0
+ARG trtllm_toolkit_version=0.11.0%2Bnightly
+ARG trtllm_version=v0.11.0
 ARG cuda_python_version=12.4
 ARG peft_version=0.10.0
 ARG triton_version=r24.04
 ARG trtllm_toolkit_wheel="https://publish.djl.ai/tensorrt-llm/toolkit/tensorrt_llm_toolkit-${trtllm_toolkit_version}-py3-none-any.whl"
-ARG trtllm_wheel="https://djl-ai.s3.amazonaws.com/publish/tensorrt-llm/${trtllm_version}/tensorrt_llm-0.10.0-cp310-cp310-linux_x86_64.whl"
+ARG trtllm_wheel="https://publish.djl.ai/tensorrt-llm/${trtllm_version}/tensorrt_llm-0.11.0-cp310-cp310-linux_x86_64.whl"
```
lanking520, line 28: @ydm-amazon this will allow us to use cloudfront (an AWS services) that do global CDN on the artifacts we have and provide 10x speed up on downloading. Using raw HTTP url for S3 will be slower than that.

tosterberg (approved):

PR #2194, opened 2024-07-18
[lmi] Dependencies upgrade for 0.29.0

maaquib:
## Description ##

### DIFF
--- a/serving/docker/lmi.Dockerfile
+++ b/serving/docker/lmi.Dockerfile
@@ -15,15 +15,15 @@ ARG cuda_version=cu124
 ARG djl_version=0.29.0~SNAPSHOT
 # Base Deps
 ARG python_version=3.10
-ARG torch_version=2.3.0
-ARG torch_vision_version=0.18.0
+ARG torch_version=2.3.1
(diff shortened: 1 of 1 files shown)


serving/docker/lmi.Dockerfile
```diff
@@ -15,15 +15,15 @@ ARG cuda_version=cu124
 ARG djl_version=0.29.0~SNAPSHOT
 # Base Deps
 ARG python_version=3.10
-ARG torch_version=2.3.0
-ARG torch_vision_version=0.18.0
+ARG torch_version=2.3.1
```
siddvenk, line 18: want to confirm - i know we have different torch version for lmi-dist potentially? does that need to be reflected here, or the lmi-dist whl should be fine?

nskool (approved):

serving/docker/lmi.Dockerfile
maaquib: Yes. 2.3.1 is what we agreed upon

siddvenk (approved):

PR #2195, opened 2024-07-18
[chat][lmi] use generation prompt in tokenizer to avoid bot prompt re…

siddvenk:
…turned in response

## Description ##
Enabling this flag appends the bot instruction to the prompt.

If no bot instruction exists for the tokenizer chat template, it has no effect. See https://huggingface.co/docs/transformers/main/en/chat_templating#what-are-generation-prompts.

As a result, the bot instruction text is not returned to the user in chat use-cases. It would still be returned to the users if using LMI/TGI api and specifying "return_full_text": true.

### DIFF
--- a/engines/python/setup/djl_python/chat_completions/chat_utils.py
+++ b/engines/python/setup/djl_python/chat_completions/chat_utils.py
@@ -41,7 +41,7 @@ def parse_chat_completions_request(input_map: Dict,
         tokenizer_inputs.append(
             message.get_tokenizer_inputs(image_token=image_token))
         images.extend(message.get_images())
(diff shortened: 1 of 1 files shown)


tosterberg:
Not sure if this is still the case but there were tokenizers that did not support `add_generation_prompt` and would fail when called with `add_generation_prompt=True`. This may no longer be a concern with newer versions but I wanted to call this out.

siddvenk:
> Not sure if this is still the case but there were tokenizers that did not support `add_generation_prompt` and would fail when called with `add_generation_prompt=True`. This may no longer be a concern with newer versions but I wanted to call this out.

that's a good callout. I'll add some logic to account for that

lanking520 (approved):

PR #2202, opened 2024-07-19
[post-7/22]Add chunked prefill support in vllm and lmi-dist

rohithkrn:
## Description ##

* Adds `enable_chunked_prefill` flag for `lmi-dist` and `vllm` backends.
* Tests pass locally.


lanking520:
let's add this after 7/22 release.

lanking520 (approved):

PR #2220, opened 2024-07-22
[vLLM][0.5.3] add new configs to engine

lanking520:
## Description ##

Add new quantization kernel to vLLM/LMI-Dist

Also adding CPU offloading option to GPUs


engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
```diff
@@ -53,6 +55,7 @@ class VllmRbProperties(Properties):
     preloaded_model: Optional[Any] = None
     max_logprobs: Optional[int] = 20
     enable_chunked_prefill: Optional[bool] = False
+    cpu_offload_gb_per_gpu: Optional[int] = None
```
maaquib, line 58: Should this be `cpu_offload_gb`([ref](https://github.com/vllm-project/vllm/blob/e0c15758b85827dcac78379e60ea975ebc0ec795/vllm/engine/arg_utils.py#L57))?

lanking520: cpu_offload_gb is confusing... the actual meaning of it is cpu_offload_gb_per_gpu. If you set to 40 and do TP4, it will allocate 160GB CPU memory

maaquib (approved):

PR #2221, opened 2024-07-23
[ci] Update correctness testing

xyang16:
## Description ##

### DIFF
--- a/.github/workflows/correctness.yml
+++ b/.github/workflows/correctness.yml
@@ -48,7 +48,7 @@ jobs:
 
   test:
     runs-on: [ "${{ matrix.test.instance }}" ]
-    timeout-minutes: 60
+    timeout-minutes: 90
     needs: create-runners
     strategy:
       fail-fast: false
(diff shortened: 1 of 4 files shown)


lanking520 (approved):

.github/workflows/correctness.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         env:
           TEST_DJL_VERSION: ${{ inputs.djl-version }}
         run: |
-          python -m pytest --capture=tee-sys -vv -k ${{ matrix.test.test }} tests.py
+          python -m pytest -k ${{ matrix.test.test }} tests.py
```
zachgk, line 98: Is there a reason for putting the correctness in a separate workflow than the main integration?

zachgk: I made integration with the idea that we can keep all of the tests and models in it. It is easy enough to add more instances to it if we need to, but otherwise there isn't really a limit to how much can be put there

xyang16: Is performance test also in the integration pipeline?

zachgk: In general, my plan was that a test should be part of the main integration if it both runs nightly and isn't flaky. By that criteria, the performance tests don't run nightly so I didn't merge them in. There are also a few that I didn't merge in just because I prioritized getting the rest of the related CI pieces done. On a more conceptual level, I would say that any test we care enough about and is cheap enough to run nightly and would want to validate before a release should be part of it

PR #2157, opened 2024-07-10
[ci] add tests for llava (vllm)

sindhuvahinis:
## Description ##

Brief description of what this PR is about

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

PR #2234, opened 2024-07-25
[multimodal][ci] adding multimodal tests

siddvenk:
## Description ##

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -574,6 +574,16 @@ lmi_dist_model_list = {
         "option.model_id": "s3://djl-llm/llama-2-tiny/",
         "option.quantize": "awq",
         "option.tensor_parallel_degree": 4
+    },
+    "llava_v1.6-mistral": {
+        "option.model_id": "s3://djl-llm/llava-v1.6-mistral-7b-hf/",
+    },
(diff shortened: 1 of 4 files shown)


lanking520:
should we just close or migrate this one? https://github.com/deepjavalibrary/djl-serving/pull/2157

tests/integration/llm/prepare.py
```diff
@@ -574,6 +574,16 @@
         "option.model_id": "s3://djl-llm/llama-2-tiny/",
         "option.quantize": "awq",
         "option.tensor_parallel_degree": 4
+    },
+    "llava_v1.6-mistral": {
+        "option.model_id": "s3://djl-llm/llava-v1.6-mistral-7b-hf",
+    },
+    "paligemma-3b-mix-448": {
+        "option.model_id": "s3://djl-llm/phi-3-vision-128k-instruct"
```
lanking520, line 582: url here is not right

siddvenk: good catch

siddvenk:
> should we just close or migrate this one? #2157

I think we can close that one - we don't need a lot of the configs that are there, and while the openai client is useful - I'd rather take that up separately and have a full Test suite for chat completions use-case/integ tests

lanking520 (approved):

PR #2242, opened 2024-07-30
[ci] add handler performance test

tosterberg:
## Description ##

This PR adds nightly benchmarking for our main GPU handlers (lmi-dist, vllm, trtllm), using tiny-llama it does performance testing for batch size 1 and 128 for both the standard and chat template inference paths. The performance metrics are persisted to cloudwatch for monitoring.

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -791,6 +791,14 @@ reranking_model_spec = {
     }
 }
 
+handler_performance_model_spec = {
+    "tiny-llama-model": {
+        "batch_size": [1, 512],
+        "seq_length": [256],
+        "tokenizer": "TinyLlama/TinyLlama-1.1B-Chat-v1.0"
+    }
+}
+
 
 def add_file_handler_to_logger(file_path: str):
     handler = logging.FileHandler(file_path, mode='w')
@@ -902,7 +910,9 @@ def awscurl_run(data,
                 concurrency,
                 num_run=5,
                 dataset=False,
-                output=False):
+                output=False,
+                json_results=False,
+                random_delay=False):
     find_awscurl()
     headers = "Content-type: application/json"
     endpoint = f"http://127.0.0.1:8080/invocations"
@@ -916,9 +926,18 @@ def awscurl_run(data,
     else:
         json_data = json.dumps(data)
         command_data = f"-d '{json_data}'"
+
+    json_output = ""
+    if json_results:
+        json_output = "--json-path benchmark.json"
+
+    delay = ""
+    if random_delay:
+        delay = '--delay "rand(0,1000)"'
+
     command = (f"./awscurl -c {concurrency} "
                f"-N {num_run} -X POST {endpoint} --connect-timeout 300 "
-               f"-H {headers} {command_data} -P -t")
+               f"-H {headers} {command_data} {delay} {json_output} -P -t")
     if tokenizer:
         command = f"TOKENIZER={tokenizer} {command}"
     if output:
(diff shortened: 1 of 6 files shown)


tests/integration/llm/client.py
```diff
@@ -772,6 +772,14 @@ def get_model_name():
     }
 }
 
+handler_performance_model_spec = {
+    "tiny-llama-model": {
+        "batch_size": [1, 128],
```
siddvenk, line 777: In our performance tests previously, we have observed a pretty noticeable change in behavior around batch size 64. Could we include some more batch-sizes here? Maybe [1, 8, 32, 64, 128, 256]?

tests/integration/llm/client.py
```diff
@@ -1356,6 +1375,86 @@ def test_handler(model, model_spec):
...
+            "MetricName": f"{metric_name}_tokenPerRequest",
+            "Unit": "Count",
+            "Value": raw_metrics["tokenPerRequest"]
+        })
+        LOGGER.info(f"{metric_name}")
+        LOGGER.info(f"raw metrics: {raw_metrics}")
+        command = f'aws cloudwatch put-metric-data --namespace "serving_handler" ' \
+                  f'--region "us-east-1" --metric-data \'{json.dumps(metrics)}\''
+        LOGGER.info(command)
+        sp.call(command, shell=True)
```
siddvenk, line 1423: not necessarily for this pr, but I feel like we should be using boto3 to publish to cloudwatch rather than a subprocess call. I think we have this in a few other places, so might be worth refactoring later.

tests/integration/llm/client.py
```diff
@@ -1356,6 +1375,86 @@ def test_handler(model, model_spec):
...
+
+
+def run_rb_handler_performance(benchmark_name, model_spec, req):
+    for batch_size in model_spec["batch_size"]:
+        high_concurrency = batch_size > 16
+        metric_name = f"{benchmark_name}-high-concurrency" if high_concurrency else benchmark_name
```
siddvenk, line 1429: I'd prefer we keep the metric name consistent for all batch_sizes/concurrency, and instead use a dimension for batch size. I'm worried that if we change the batch_size test cases above, we'll lose insight into the behavior differences across more granular batch size tests

tests/integration/llm/client.py
tosterberg: We can if that is preferable, the original goal was to get a no batch and a high batch baseline set. And I see what you are looking to accomplish with this, making this more diagnostic and flexible. Current runtime for the test is about 34 minutes, so I think I will make some changes to support the expansion but keep it to these two values until we discuss in ops next week.

tests/integration/llm/client.py
tosterberg: I agree - I did just follow the subprocess call from previous uses. If we bring in Boto3 it will mean that we need to manage the dependency for some CI tests, which isn't big but is worth noting before we make the refactor.

tests/integration/llm/client.py
tosterberg: I will make an update to just have concurrency in the name, this along with the note above make sense. I think my longer range goal is to be able to also handle multi-modal handler performance like this, so there might be some discussion around metric name there as well.

tests/integration/llm/client.py
lanking520, line 777: batch size is a bit small, we can go even larger like 512 with tiny llama. This could help to zoom the perf diff

siddvenk: yea, we can easily change it in the future so i'm cool with that.

siddvenk (approved):

PR #2245, opened 2024-07-31
[CI] Update integration test scripts for Neo

ethnzhng:
This PR adds helpers & model configs to the integration test scripts to enable simulating the SageMaker Neo environment for TRT-LLM, vLLM and Neuron compilation/quantization jobs.


tests/integration/llm/client.py
```diff
@@ -192,6 +192,34 @@ def get_model_name():
...
+transformers_neuronx_neo_model_spec = {
+    "llama-2-13b-rb": {
+        "seq_length": [1024],
+        "batch_size": [1, 4],
+        "tokenizer": "TheBloke/Llama-2-13B-fp16"
+    },
+    "mixtral-8x22b": {
+        "workers": 1,
+        "seq_length": [512],
+        "batch_size": [2]
```
lanking520, line 204: do we have tokenizer defined here as well?

ethnzhng: @a-ys

lanking520 (approved):

frankfliu (approved):

tests/integration/llm/client.py
a-ys: My understanding is: only included tokenizer for rolling batch handler tests since those use awscurl to test. `‎test_transformers_neuronx_handler` doesn't use tokenizer.

PR #2251, opened 2024-07-31
[ci] Update text embedding integration to Rust engine

xyang16:
## Description ##


tests/integration/llm/prepare.py
```diff
@@ -1145,6 +1157,24 @@ def build_correctness_model(model):
...
 
+def build_text_embedding_model(model):
+    if model not in text_embedding_model_list:
+        raise ValueError(
+            f"{model} is not one of the supporting handler {list(onnx_list.keys())}"
+        )
+    options = text_embedding_model_list[model]
+    options["engine"] = "Rust"
+    options["option.task"] = "text_embedding"
+    options["option.mapLocation"] = "true"
```
frankfliu, line 1168: We don't need mapLocation for Rust engine

tests/integration/llm/prepare.py
```diff
@@ -1145,6 +1157,24 @@ def build_correctness_model(model):
...
+    options["padding"] = "true"
+    options["pooling"] = "cls"
+    options["includeTokenTypes"] = "false"
```
frankfliu, line 1171: This is not always false

frankfliu: Can we rely on `djl-convert` to generate those values?

tests/integration/llm/prepare.py
xyang16: Removed.

tests/integration/llm/prepare.py
xyang16: Changed.

tests/integration/llm/prepare.py
```diff
@@ -1145,6 +1157,17 @@ def build_correctness_model(model):
     write_model_artifacts(options)
 
 
+def build_text_embedding_model(model):
+    if model not in text_embedding_model_list:
+        raise ValueError(
+            f"{model} is not one of the supporting handler {list(onnx_list.keys())}"
+        )
+    options = text_embedding_model_list[model]
+    options["engine"] = "Rust"
```
lanking520, line 1166: how will sagemaker custoemr specify Rust engine?

xyang16: They can use OPTION_ENGINE=Rust environment variable.

frankfliu (approved):

PR #2250, opened 2024-07-31
[docker] Update DJL to 0.30.0

xyang16:
## Description ##

Brief description of what this PR is about

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

PR #2276, opened 2024-08-05
[cherrypick] 29 new commits from master

ydm-amazon:
Cherry pick 29 new commits from master


lanking520:
#2254  #2252 #2250 @frankfliu please check if we need some changes in those PRs

ydm-amazon:
Sounds good - I will merge then

(only option for merge is squash and merge so I will do that)

PR #2282, opened 2024-08-06
Add integration tests for Mistral-7B-Instruct-v0.3 with and without fp8 quantization

Mancera1:
## Description ##

This PR adds integration tests for the Mistral-7B-Instruct-v0.3 model, both with and without fp8 quantization. The changes include:

Adding two new test cases:

1. test_mistral_7b: Runs the correctness test for the non-quantized Mistral-7B-Instruct-v0.3 model.
2. test_mistral_7b_fp8: Runs the correctness test for the fp8 quantized version of the Mistral-7B-Instruct-v0.3 model.
3. Adding configuration for the trtllm-mistral-7b-instruct-v0.3 and trtllm-mistral-7b-instruct-v0.3-fp8 models in client.py and prepare.py.

Note: The score parameter for both models has been set to 0.59 instead of the standard 0.6. During various correctness tests, the score consistently hovered around 0.59 regardless of the num_run value. The best-performing configuration was with num_run set to 100, which scored 0.001 higher than when set to 66. However, as num_run increased towards 200, performance seemed to degrade, although no tests beyond 200 were conducted.

Based on these observations, it seems logical to lower the score to 0.59. However, feedback from the team is requested to ensure this decision aligns with the project's requirements and expectations.

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -974,14 +974,14 @@ correctness_model_list = {
     "trtllm-mistral-7b-instruct-v0.3": {
         "engine": "Python",
         "option.task": "text-generation",
-        "option.model_id": "mistralai/Mistral-7B-Instruct-v0.3",
+        "option.model_id": "s3://djl-llm/mistral-7b-instruct-v03/",
         "option.rolling_batch": "trtllm",
         "option.tensor_parallel_degree": 4
     },
     "trtllm-mistral-7b-instruct-v0.3-fp8": {
         "engine": "Python",
         "option.task": "text-generation",
-        "option.model_id": "mistralai/Mistral-7B-Instruct-v0.3",
+        "option.model_id": "s3://djl-llm/mistral-7b-instruct-v03/",
         "option.rolling_batch": "trtllm",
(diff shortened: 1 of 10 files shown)


tests/integration/llm/prepare.py
```diff
@@ -962,6 +962,29 @@
...
+        "option.task": "text-generation",
+        "option.model_id": "s3://djl-llm/llama-3-8b-hf/",
+        "option.rolling_batch": "trtllm",
+        "option.tensor_parallel_degree": 4,
+        "option.quantize": "fp8"
+    },
+    "trtllm-mistral-7b-instruct-v0.3": {
+        "engine": "Python",
+        "option.task": "text-generation",
+        "option.model_id": "mistralai/Mistral-7B-Instruct-v0.3",
```
siddvenk, line 977: We'll want to change these to the s3 locations. Let's use `s3://djl-llm/mistral-7b-instruct-v03/` here and for the other mistral config below.

siddvenk:
This change also includes the commits for the llama3 change - please fetch the latest changes from master branch and rebase

siddvenk (approved):

PR #2284, opened 2024-08-06
[trt] use cu123 base image and cudacompat as trtllm 0.9.0 is dependen…

siddvenk:
…t on cuda12.3.2

## Description ##

This bumps the cuda version, and cuda compat version, for the trt-0.28.0 container.

TRT-LLM 0.9.0 states that it is dependent on cuda 12.3.2 in the release notes. 

I have validated these changes work with a sample model, but this will require full integration tests validation.

Note: Since we're updating the cuda version here, this will require a "new" dlc release process

### DIFF
--- a/serving/docker/tensorrt-llm.Dockerfile
+++ b/serving/docker/tensorrt-llm.Dockerfile
@@ -9,9 +9,9 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-ARG version=12.2.2-cudnn8-devel-ubuntu22.04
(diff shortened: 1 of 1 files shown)


serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -9,9 +9,9 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-ARG version=12.2.2-cudnn8-devel-ubuntu22.04
+ARG version=12.3.2-devel-ubuntu22.04
```
frankfliu, line 12: Can we use cu12.4.1?

siddvenk: Yea i think that should be fine as well - will update

lanking520 (approved):

PR #2286, opened 2024-08-06
[TRTLLM] take out cudnn

lanking520:
## Description ##

Tried to take out CUDNN


lanking520:
https://github.com/deepjavalibrary/djl-serving/actions/runs/10272449471

ydm-amazon:
At that time it was necessary, has anything changed now?

lanking520:
https://nvidia.github.io/TensorRT-LLM/installation/linux.html @ydm-amazon looks like we only need cudnn at build time, but not at runtime

ydm-amazon (approved):

PR #2289, opened 2024-08-06
[python] fix the NoneType error when decoded_token is empty

sindhuvahinis:
## Description ##

Yet to test this use cases


lanking520 (approved):

siddvenk:
What type of test do we need to add to ensure no regression here?

PR #2322, opened 2024-08-16
fix: allow chat template for non batch

sindhuvahinis:
…or rb

## Description ##

### DIFF
--- a/engines/python/setup/djl_python/chat_completions/chat_utils.py
+++ b/engines/python/setup/djl_python/chat_completions/chat_utils.py
@@ -13,17 +13,22 @@
 from typing import Dict, Optional
 
 from djl_python.chat_completions.chat_properties import ChatProperties
+from djl_python.properties_manager.properties import Properties
 
 
 def is_chat_completions_request(inputs: Dict) -> bool:
     return "messages" in inputs
 
 
-def parse_chat_completions_request(input_map: Dict,
-                                   is_rolling_batch: bool,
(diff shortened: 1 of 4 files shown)


engines/python/setup/djl_python/chat_completions/chat_utils.py
```diff
@@ -13,17 +13,22 @@
...
+    is_rolling_batch: bool,
+    tokenizer,
+    image_token: Optional[str] = None,
+    configs: Properties = None,
+):
+    # Chat completions can either be a rolling batch or no-batching .
+    if not (is_rolling_batch or configs.batch_size == 1):
         raise ValueError(
             "chat completions support is not currently available for dynamic batching. "
             "You must enable rolling batch to use the chat completions format."
```
siddvenk, line 34: we should update the error here to reflect that either rolling batch or dynamic batch with size == 1 is needed. 

Is there a reason we can't use bs > 1 with dynamic batch here? (I know that i put in this rolling batch restriction a while ago lol, just forgot why)

sindhuvahinis: lol, Frank was trying to use chat template in a object detection model and found this issue. 

Technically we can, but with dynamic batching, parameters cannot be different between each requests. And we dont use chat output formatter with non-rolling batch cases. We should fix that as well.

frankfliu (approved):

PR #2331, opened 2024-08-21
[ci] Fix awscurl run headers

xyang16:
## Description ##

Fix awscurl run headers: Remove space because it's string.

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -975,7 +975,7 @@ def awscurl_run(data,
                 json_results=False,
                 random_delay=False):
     find_awscurl()
-    headers = "Content-type: application/json"
+    headers = "Content-type:application/json"
     endpoint = f"http://127.0.0.1:8080/invocations"
     if dataset:
         dataset_dir = os.path.join(os.path.curdir, "dataset")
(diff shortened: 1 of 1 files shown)


tests/integration/llm/client.py
```diff
@@ -975,7 +975,7 @@ def awscurl_run(data,
                 json_results=False,
                 random_delay=False):
     find_awscurl()
-    headers = "Content-type: application/json"
+    headers = "Content-type:application/json"
```
frankfliu, line 978: Why this causing problem?

xyang16: See this example:

`./awscurl -c 1 -N 5 -X POST http://127.0.0.1:8080/invocations --connect-timeout 300 -H Content-type: application/json -d '[{"key": "what is panda?", "value": "hi"}]'   -P -t`

Having space in this the Content-type not passed in correctly. The other fix is add quote for Content-type.

frankfliu: I see, we should add quote instead of remove space

xyang16: Changed.

frankfliu (approved):

PR #2341, opened 2024-08-27
[awscurl] Allows set max length by env var

frankfliu:
## Description ##


tosterberg (approved):

CoolFish88:
Hello, 
I was curious how is the current setup impacting token metric computation. Can we assume the returned values are accurate and reliable, or is the PR going to correct for some underlying biases?

Thanks

frankfliu:
`awscurl` is a tool trying measure the model performance from client side. It uses tokenizer to count tokens rather than word count. But this assumes decode and encode are symmetric which is not always true. If model generate long context which exceed max model length, currently `awscurl` implementation won't be able to get correct token count (can be resolved by counting overflow tokens).

PR #2342, opened 2024-08-28
[fix] prevent requests being sent to python model until model is full…

siddvenk:
…y loaded and warmup request has completed

## Description ##

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
@@ -47,6 +47,7 @@ class PyProcess {
     private List<Connection> connections;
     private CountDownLatch latch;
     private volatile boolean started; // NOPMD
+    private volatile boolean modelLoaded; // NOPMD
     private AtomicInteger restartCount;
     private CompletableFuture<Void> restartFuture;
     private boolean trtLlmMode;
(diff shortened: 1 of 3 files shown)


engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
```diff
@@ -188,6 +190,7 @@ synchronized void startPythonProcess() {
     }
 
     synchronized void stopPythonProcess(boolean error) {
+        modelLoaded = false;
```
frankfliu, line 193: I feel move this to startPythonProcess is more readable

engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
```diff
@@ -237,6 +240,10 @@ boolean isStopped() {
         return !started;
     }
 
+    boolean isModelLoaded() {
```
frankfliu, line 243: We can combine isStopped() and isModelLoaded() into `isRead()`

engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
siddvenk: done

engines/python/src/main/java/ai/djl/python/engine/PyProcess.java
siddvenk: done

frankfliu (approved):

PR #2356, opened 2024-09-03
[fix][sf] fix bug with PyPredictor to remove worker, add specific fla…

siddvenk:
…g for sm health check behavior

## Description ##

### DIFF
--- a/serving/src/main/java/ai/djl/serving/models/ModelManager.java
+++ b/serving/src/main/java/ai/djl/serving/models/ModelManager.java
@@ -29,6 +29,7 @@ import ai.djl.serving.wlm.WorkLoadManager;
 import ai.djl.serving.wlm.WorkerPool;
 import ai.djl.serving.wlm.WorkerPoolConfig;
 import ai.djl.serving.workflow.Workflow;
+import ai.djl.util.Utils;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


serving/src/main/java/ai/djl/serving/models/ModelManager.java
```diff
@@ -416,6 +417,18 @@ public CompletableFuture<Map<String, Object>> workerStatus() {
                                         if (wlm.getWorkerPool(wpc).isFullyScaled()) {
                                             data.put(modelName, new StatusResponse("Healthy"));
                                         } else {
+                                            boolean sageMakerHealthCheckOverride =
```
frankfliu, line 420:
```suggestion
if (Boolean.getBoolean("SAGEMAKER_HEALTH_CHECK_OVERRIDE")) {
```

serving/src/main/java/ai/djl/serving/models/ModelManager.java
```diff
@@ -416,6 +417,18 @@ public CompletableFuture<Map<String, Object>> workerStatus() {
...
+                                                    Boolean.parseBoolean(
+                                                            Utils.getenv(
+                                                                    "SAGEMAKER_HEALTH_CHECK_OVERRIDE",
+                                                                    "false"));
+                                            if (sageMakerHealthCheckOverride) {
+                                                logger.warn(
```
frankfliu, line 426: This should be info level

serving/src/main/java/ai/djl/serving/models/ModelManager.java
siddvenk: updated

serving/src/main/java/ai/djl/serving/models/ModelManager.java
siddvenk: updated, and also changed the other log we had for SERVING_HEALTH_CHECK_OVERRIDE from error -> info

frankfliu (approved):

sindhuvahinis (approved):

PR #2363, opened 2024-09-05
[serving] add request id logging on invocations/predictions path

siddvenk:
## Description ##

1. Parse requestId from HTTP header that is present on SageMaker (and SageMaker like) platforms
2. Log requestId details in relevant places on the invocations/predictions route

This is part 1 of the requestId logging changes. It only touches the java side
- This is the best we can to on the Java side, and should be sufficient. 

Part 2 will focus on the python side

### DIFF
--- a/serving/src/main/java/ai/djl/serving/http/Session.java
+++ b/serving/src/main/java/ai/djl/serving/http/Session.java
@@ -13,6 +13,7 @@
 package ai.djl.serving.http;
 
 import ai.djl.modality.Input;
+import ai.djl.serving.util.ConfigManager;
 
 import io.netty.handler.codec.http.HttpRequest;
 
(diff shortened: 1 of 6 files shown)


serving/src/main/java/ai/djl/serving/http/Session.java
```diff
@@ -46,7 +46,11 @@ public Session(String remoteIp, HttpRequest request) {
             method = "GET";
             protocol = "HTTP/1.1";
         }
-        requestId = UUID.randomUUID().toString();
+        // SageMaker populates requestId in this header
+        requestId = request.headers().get("X-Amzn-SageMaker-Inference-Id");
```
frankfliu, line 50: Better make `X-Amzn-SageMaker-Inference-Id` configurable, e.g. `ConfigManager.getRequestIdKey()`

siddvenk: good call

frankfliu (approved):

PR #2364, opened 2024-09-05
[CI] Add llama-3.1 lmi-dist test, with secure mode enabled

ethnzhng:
Adding lmi-dist test case using llama-3.1-8b, while having the SM Secure Mode enabled


tests/integration/tests.py
```diff
@@ -520,6 +520,16 @@ def test_llama2_7b_chat(self):
             r.launch()
             client.run("lmi_dist_chat llama2-7b-chat".split())
 
+    def test_speculative_llama_secure(self):
+        with Runner('lmi', 'speculative-llama-13b') as r:
```
sindhuvahinis, line 524: Does it have to be bigger model? Just Asking, because this might increase the running time.

ethnzhng: It doesn't need to be a bigger model, but since currently secure mode is enabled chiefly when speculative decoding is being used, I figured it would be a more realistic case

sindhuvahinis (approved):

PR #2382, opened 2024-09-12
[Neo][vLLM] Accept quant options for awq, fp8

a-ys:
## Description ##

This PR adds additional pass-through options to configure awq & fp8 quantization.

For **AWQ**, we add the following options. These map to the options [defined here](https://github.com/casper-hansen/AutoAWQ/blob/eab1a4accf56bd01b3050b091fb74cfb42bf3c6e/examples/cli.py#L16).
- `option.awq_zero_point`: toggles zero point quantization
- `option.awq_block_size`: (existing field) group/block size for awq quantization
- `option.awq_weight_bit_width`: bit width for quantization. currently only 4 is supported.
- `option.awq_mm_version`: awq matmul implementation
- `option.awq_ignore_layers`: layers to ignore during quantization

For **FP8**, we add the following options. These are [defined here](https://github.com/neuralmagic/AutoFP8/blob/2cd265ffbcf73cfcca4c11b42673d93c5d508476/auto_fp8/config.py#L4).
- `option.fp8_activation_scheme`: static or dynamic activation scaling factors
- `option.fp8_kv_cache_quant_targets`: modules to target for kv cache quantization (currently unused)
- `option.fp8_ignore_patterns`: layers to ignore
- `option.calib_size`: (existing field) number of samples for activation scales calibration

For fields that are read by the underlying library as a list of strings, we accept them like so:
```
option.fp8_kv_cache_quant_targets=k_proj, v_proj
```

Also removes previously implemented configuration options for FP8 as added in the original PR (#2272)

### DIFF
--- a/serving/docker/partition/partition.py
+++ b/serving/docker/partition/partition.py
@@ -18,7 +18,6 @@ import logging
 import argparse
 import subprocess
 
-from typing import Optional
 from pathlib import Path
 
 import utils
@@ -243,7 +242,7 @@ class PartitionService(object):
             if entry_point_file:
                 os.remove(os.path.join(saved_checkpoints_dir, 'model.py'))
 
-    def run_quantization(self, autofp8_config: Optional[dict] = None):
+    def run_quantization(self):
         quant_method = self.properties['option.quantize']
         if quant_method == 'awq':
             logging.info("Running AutoAWQ quantization")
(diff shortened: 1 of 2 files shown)


siddvenk:
Changes lgtm, but i don't have much context. Have these been tested? do we have tests in lmi-distro?

a-ys:
Have done some tests locally to test that the parsing works properly and
the default workflow (no options) is still functional. Working on expanded
CI tests right now for lmi distro

siddvenk (approved):

siddvenk:
Can you make a cherry-pick PR for this on the 0.29.0-dlc branch as well? At the very least let's ensure no regression with our automated tests here before the patch release

a-ys:
Sounds good, will do.

PR #2410, opened 2024-09-26
Update trtllm deps

ydm-amazon:
- Increases timeout minutes: Due to changes in v0.12.0, compilation takes longer. I tested this locally and can verify that it can take over 2 hours, so changed the amount of minutes before timeout to add a nice buffer
- Change the default to something convenient

### DIFF
--- a/.github/workflows/trtllm-deps-build.yml
+++ b/.github/workflows/trtllm-deps-build.yml
@@ -6,11 +6,11 @@ on:
       trtllm_branch:
         description: 'The interal branch of trtllm'
         required: true
-        default: 'lmi_v0.11.0'
+        default: 'v0.12.0'
       release_tag:
         description: 'The released tag version of trtllm'
         required: true
-        default: 'v0.11.0'
+        default: 'v0.12.0'
       python_version:
         description: 'The python version of release'
         required: false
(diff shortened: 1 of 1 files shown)


siddvenk:
Do we know why there is a huge increase in the compilation time? Any idea whether this is expected or not?

ydm-amazon:
Good question - I am not sure. I spent some time yesterday to manually look into whether it was hanging, but could not reach the end of that thread. I hope to just run it via automation today; if it takes more than 200 minutes, then something is clearly wrong.

siddvenk:
This might be something worth bringing up with Nvidia - for now i think it's ok to do this in CI so we can work on getting things integrated.

But, I'm hesitant to do a full release with this version without some signoff/guidance from Nvidia.

ydm-amazon:
Yep, I think that's needed in the long run, but I still have a couple steps after this - including testing breaking changes after I get the TensorRT-LLM wheel built here, so this is a good greedy type of choice to do here

siddvenk (approved):

ydm-amazon:
Update: looks to be normal. Build finished in 1h 46 min which is only 20% slower than the last version

PR #2413, opened 2024-10-02
[vllm] Support speculative decoding in vllm rolling batch

xyang16:
## Description ##


sindhuvahinis (approved):

tests/integration/llm/prepare.py
```diff
@@ -625,6 +625,24 @@
         "option.tensor_parallel_degree": 4,
         "option.enable_chunked_prefill": "true",
     },
+    "llama-68m-speculative-medusa": {
+        "option.model_id": "JackFram/llama-68m",
+        "option.task": "text-generation",
+        "option.speculative_model": "abhigoyal/vllm-medusa-llama-68m-random",
```
tosterberg, line 631: Non blocking: we should move these to s3 to avoid long term availability flakiness like we have seen with adapter models.

tests/integration/llm/prepare.py
```diff
@@ -625,6 +625,24 @@
...
+        "option.num_speculative_tokens": 4,
+        "option.use_v2_block_manager": True,
+        "option.tensor_parallel_degree": 1,
+        "option.max_rolling_batch_size": 4,
+    },
+    "llama-68m-speculative-eagle": {
+        "option.model_id": "JackFram/llama-68m",
+        "option.task": "text-generation",
+        "option.speculative_model": "abhigoyal/vllm-eagle-llama-68m-random",
```
tosterberg, line 640: Non blocking: we should move these to s3 to avoid long term availability flakiness like we have seen with adapter models.

tests/integration/llm/prepare.py
xyang16: Thanks Tyler. Uploaded llama-68m.

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
```diff
@@ -59,6 +59,22 @@ class VllmRbProperties(Properties):
     enable_prefix_caching: Optional[bool] = False
     disable_sliding_window: Optional[bool] = False
     limit_mm_per_prompt: Optional[Mapping[str, int]] = None
+    use_v2_block_manager: bool = False
```
davidthomas426, line 62: FYI: https://github.com/vllm-project/vllm/pull/8678 (not merged yet, but changes this default to True in vLLM).

PR #2415, opened 2024-10-02
[vllm] Validate speculative model config

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
@@ -13,7 +13,7 @@
 from enum import Enum
 from typing import Optional, Any, Mapping
 
-from pydantic import field_validator
+from pydantic import field_validator, model_validator
 
 from djl_python.properties_manager.properties import Properties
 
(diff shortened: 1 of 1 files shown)


sindhuvahinis (approved):

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
```diff
@@ -103,3 +103,11 @@ def validate_limit_mm_per_prompt(cls, val) -> Mapping[str, int]:
                     f"Conflicting values specified for key: {key}")
             out_dict[key] = parsed_value
         return out_dict
+
+    @model_validator(mode='after')
+    def validate_speculative_model(self):
+        if self.speculative_model is not None and not self.use_v2_block_manager:
+            raise ValueError(
+                "Speculative decoding requires usage of the V2 block manager. Enable it with option.use_v2_block_manager=true."
```
davidthomas426, line 111: That's actually not true.

davidthomas426: Ok, just double-checked. It actually is true. I was confused by the inconsistency between these two pieces of code in vllm v0.6.2:
- https://github.com/vllm-project/vllm/blob/f58d4fccc9b270838be438f5f0db71bea156a56d/vllm/engine/arg_utils.py#L964-L968 - sets the variable with a warning log if not set and it's required
- https://github.com/vllm-project/vllm/blob/7193774b1ff8603ad5bf4598e5efba0d9a39b436/vllm/config.py#L1203-L1206 - throws error if not set and it's required
-

PR #2418, opened 2024-10-04
[lmi] add stricter payload validation on default schema for rolling-b…

siddvenk:
…atch use-cases

## Description ##

Adding some light validation on the default schema for requests in the rolling batch case. For rolling batch, we have a published schema, but no real validation in the code.

I noticed that while we have relatively good support for input processing and formatting across the different batching mechanisms and engines, the output_formatter only takes effect on rolling_batch use-cases. We don't have good output_formatting support for dynamic batch use-cases (at least from the default codepath i've tried, we can issue a chat completions request to a model using dynamic batching, but it's not output formatted correctly)

As a follow up here, I'm going to clean up some code and unify these behaviors.


sindhuvahinis (approved):

sindhuvahinis:
Nice!

PR #2420, opened 2024-10-07
[lmi-dist][breaking] remove default 4096 token number for prefill - a…

siddvenk:
…lign with vllm default behavior

## Description ##

### DIFF
--- a/serving/docs/lmi/announcements/breaking_changes.md
+++ b/serving/docs/lmi/announcements/breaking_changes.md
@@ -1,0 +1,43 @@
+# LMI Breaking Changes
+
+This document details breaking changes for LMI container version releases.
+
+## LMI Container v12 (0.30.0)
+
+### Changes to model loading/warmup behavior.
+
+#### Summary
+Starting in LMI v12 (0.30.0), the default behavior for `option.max_rolling_batch_prefill_tokens` when using the lmi-dist 
+rolling-batch backend has changed.
+This configuration maps to vLLM's `max_num_batched_tokens` configuration.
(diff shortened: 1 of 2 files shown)


sindhuvahinis (approved):

serving/docs/lmi/announcements/breaking_changes.md
```diff
@@ -0,0 +1,43 @@
...
+with v11 and earlier may start failing in v12 due to higher memory requirements during model loading and warmup 
+for models with greater than a supported sequence length greater than 4096 tokens.
+
+#### Guidance
+As a result of this change, we recommend you take the following actions:
+
+* Enable chunked-prefill. You can do this by setting `option.enable_chunked_prefill=true`.
+  * You can learn more about chunked prefill here https://docs.vllm.ai/en/latest/models/performance.html#chunked-prefill.
+  * We expect that chunked prefill will be beneficial for most users and use-cases. But, you should confirm this for your specific use-case.
+  * With chunked-prefill, the default value for max_rolling_batch_prefill_tokens is 512, which is set to optimize inter token latency (ITL).
```
davidthomas426, line 38: `max_rolling_batch_prefill_tokens` is not the best name for this parameter, especially with chunked prefill. Perhaps we should also add an alias for `max_num_batched_tokens` to match vllm? Of course, we could do that next version instead, or in a separate PR, or whatever.

siddvenk: alias sounds reasonable to me, will add

siddvenk: will do in follow up pr

PR #2426, opened 2024-10-10
[ci][utils] allow specifying output dir, make allow_patterns easier t…

siddvenk:
…o work with in local mode

## Description ##

### DIFF
--- a/tests/integration/llm/download_llm.py
+++ b/tests/integration/llm/download_llm.py
@@ -3,15 +3,9 @@ from huggingface_hub import snapshot_download
 from pathlib import Path
 
 
-def llm_download(model_id, token, allow_patterns):
-    local_model_path = Path("./model")
+def llm_download(model_id, token, allow_patterns, download_dir):
+    local_model_path = Path(download_dir)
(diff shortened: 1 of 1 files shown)


tests/integration/llm/download_llm.py
```diff
@@ -27,10 +21,17 @@ def llm_download(model_id, token, allow_patterns):
                         required=False,
                         type=str,
                         help='The HuggingFace token have access to the model')
-    parser.add_argument('--allow-patterns',
+    parser.add_argument(
+        '--allow-patterns',
+        required=False,
+        type=str,
+        default='*.json,*.pt,*.safetensors,*.txt,*.model,*.tiktoken',
```
sindhuvahinis, line 28: Should it be a list? i.e, Should we need nargs='+' or passing it as string is fine?

siddvenk: it expects a string now, see a few lines below where it does the replace/split to turn the string into a list

sindhuvahinis (approved):

PR #2445, opened 2024-10-15
[lmi][lmi-dist] Add pipeline-parallel support to lmi-dist, and use_passive_workers by default

davidthomas426:
## Description ##

This adds pipeline-parallel support to lmi-dist rolling batch, whether multi-node or not. It also updates logic on the java side for how to handle pipeline parallel degree and tensor parallel degree. Finally, it switches to a passive worker mode when using MPI, where the frontend only sends inference messages to rank 0 after model load--trtllm already worked this way, but it's extra important when running with pipeline parallelism to not synchronize the worker processes after each iteration.


siddvenk:
lgtm - you can format the code with `./gradlew formatJava`

I want to clean up the tp logic, it's being handled in too many places. Not for this PR, but as a follow up (likely after this release)

PR #2447, opened 2024-10-15
[lmi] Change speculative_length default to 4

davidthomas426:
## Description ##

Given the fact that the bonus token was thrown away in previous versions of lmi_dist, the true number of speculative tokens was effectively off by 1 from how it was configured. Now that we correctly handle bonus token rather than throwing it away, the new version will have the same behavior with speculative_length 4 that the previous versions had with speculative_length 5 (but with better performance!). Thus, we change the default to 4.

We need a followup PR to update the documentation accordingly.


siddvenk (approved):

siddvenk:
I'll work with @tosterberg to get the docs updated for this

PR #2448, opened 2024-10-15
[fix] optimum token selector do sample defaults

tosterberg:
## Description ##

Fixing error with default behavior when do_sample is true but neither top_p / top_k are set. Fix added for defaulting the prefill/decode bool to None inside the NiaveRollingBatcher to support the updates to Optimum modeling. Additionally, removing stale model tests.

### DIFF
--- a/engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_neuron_scheduler.py
+++ b/engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_neuron_scheduler.py
@@ -177,8 +177,9 @@ class NeuronGenerator(ABC):
             slot_request.slot.clear()
         return generation, finish_reason
 
-    def _generate_token(self, inputs: GenerationInputs,
-                        prefill: Optional[bool]) -> List[Generation]:
+    def _generate_token(self,
+                        inputs: GenerationInputs,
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/transformers_neuronx_scheduler/optimum_neuron_scheduler.py
```diff
@@ -178,7 +178,7 @@ def make_generations(
         return generation, finish_reason
 
     def _generate_token(self, inputs: GenerationInputs,
-                        prefill: Optional[bool]) -> List[Generation]:
+                        prefill: Optional[bool] = None) -> List[Generation]:
```
sindhuvahinis, line 181: Is this the error for do_sample?

tosterberg: It is the error for the unit test with the tiny models, I can move it out into a separate PR if you would like. I forgot that I had added that fix here.

sindhuvahinis: No that's okay

sindhuvahinis (approved):

PR #2449, opened 2024-10-15
Decouple djl and serving in dockerfiles and gradle and add serving publish

brndysgit:
This PR aims to move towards decoupling the DJL library from DJL Serving by:

1. Separating version dependencies between DJL and DJL Serving in the Dockerfiles and Gradle build files.
2. Introducing the serving_publish.yml workflow in the DJL-Serving repository to manage releases independently.
We also remove the Maven publishing logic in the serving_publish workflow as it is no longer needed. 
3. Decoupling DJL version and DJL-Serving versioned publish artifacts for docker container builds in djl serving installation script 

      - aarch64 in nightly publish needs to grab env variables again because it's on a separate runner

### DIFF
--- a/.github/workflows/docker-nightly-publish.yml
+++ b/.github/workflows/docker-nightly-publish.yml
@@ -58,6 +58,13 @@ jobs:
         with:
           path: ~/.gradle/caches
           key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*') }}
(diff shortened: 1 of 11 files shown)


brndysgit:
After this PR is merged, I'll make a PR to remove serving publish from the DJL repo.

tosterberg (approved):

frankfliu:
For DJL user, how can they get snapshot release on djl-serving?

PR #2451, opened 2024-10-16
[lmi][docker] use index url instead of extra index url, cu124 for torch

siddvenk:
## Description ##


sindhuvahinis (approved):

frankfliu:
Why we use `--index-url`, does `--extra-index-url` causing issue?

PR #2453, opened 2024-10-16
[ci] update snapshot logic for serving

tosterberg:
## Description ##

Continuation of the decoupling work:
- specify snapshot build args for serving, where it had previously been djl snapshots
- update version logic for serving publish to be based off of serving version number rather than djl version number
- update docker images to default to snapshot version of serving for nightly build

### DIFF
--- a/.github/workflows/docker-nightly-publish.yml
+++ b/.github/workflows/docker-nightly-publish.yml
@@ -76,7 +76,7 @@ jobs:
           export NIGHTLY="-nightly"
           docker compose build --no-cache \
           --build-arg djl_version=${{ env.DJL_VERSION }} \
-          --build-arg djl_serving_version=${{ env.SERVING_VERSION }} \
+          --build-arg djl_serving_version=${{ env.SERVING_VERSION }}-SNAPSHOT \
           ${{ matrix.arch }}
           docker compose push ${{ matrix.arch }}
       - name: Build and push temp image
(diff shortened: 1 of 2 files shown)


tosterberg:
@frankfliu - if you have time to take a look to see if I am going in the right direction for this.

serving/docker/Dockerfile
```diff
@@ -11,7 +11,7 @@
 # the specific language governing permissions and limitations under the License.
 FROM ubuntu:22.04 AS base
 ARG djl_version
```
siddvenk, line 13: for all - why don't we have a default for djl_version?

brndysgit: djl_version bumping is centralized in the TOML file now

brndysgit:
would it be a better idea for the nightly build CI to dynamically handle the logic for specifying using the snapshot versions to reduce redunancy (we can just manage versions with the TOML file and avoid hardcoding in dockerfiles?)

tosterberg:
It is an option - I am still trying to recall the original reason we did it this way in the past for DJL. My initial recollection was that we use -SNAPSHOT for our publish and ~SNAPSHOT for pulling the artifacts which meant that they ARGs weren't one to one, but I can't remember why.

serving/docker/Dockerfile
tosterberg: yeah, we would make the change in the toml file's version for djl https://github.com/deepjavalibrary/djl-serving/blob/master/gradle/libs.versions.toml#L5. It will be different then it had been, but @brndysgit  is going to work on the documentation for the change.

frankfliu:
@tosterberg I don't think this will work. Will discuss with you offline

PR #2455, opened 2024-10-17
[lmi][docker] update to vllm0.6.2 with cherry-pick for mllama, additi…

siddvenk:
…onal minor dep updates

## Description ##

I still need to rebuild the vllm wheel (for some reason my cherry-pick wasn't being built into the wheel), but that cherry-pick is only needed for MLlama.

I also moved the pytorch pip install to its own command since --index-url is applied globally, and that causes issues with packages not found in the pytorch index.


xyang16:
vllm 0.6.3 is available.

siddvenk:
> vllm 0.6.3 is available.

We're not compatible with vllm 0.6.3 - quite a few breaking changes that we'll need to address in a separate release

sindhuvahinis (approved):

PR #2456, opened 2024-10-17
[lora] Support pin adapter API

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -602,7 +602,7 @@ def unregister_adapter(inputs: Input):
     try:
         removed = _service.rolling_batch.remove_lora(adapter_name)
         if not removed:
-            logging.debug(
+            logging.info(
                 f"Remove LoRA adapter {adapter_name} returned false, the adapter may have already been evicted."
             )
     except Exception as e:
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -23,7 +23,7 @@ from transformers import (pipeline, Pipeline, AutoModelForCausalLM,
                           StoppingCriteriaList)
 from transformers.tokenization_utils_base import PreTrainedTokenizerBase
 from peft import PeftConfig, PeftModel, PeftModelForCausalLM
-from typing import List, Dict
+from typing import List, Dict, Optional
 
 from djl_python.encode_decode import encode
 from djl_python.inputs import Input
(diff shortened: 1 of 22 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -242,6 +242,41 @@ def inference(self, inputs: Input) -> Output:
...
+    def add_lora(self,
+                 lora_name: str,
+                 lora_path: str,
+                 long_lora_max_len: Optional[int] = None,
+                 base_model_name: Optional[str] = None):
+        if is_rolling_batch_enabled(self.hf_configs.rolling_batch):
+            return self.rolling_batch.add_lora(lora_name, lora_path,
+                                               long_lora_max_len,
+                                               base_model_name)
+        else:
```
siddvenk, line 254: do we need to support all of these apis for huggingface accelerate (non-rolling batch?)

engines/python/setup/djl_python/huggingface.py
```diff
@@ -498,39 +533,81 @@ def register_adapter(inputs: Input):
...
+                "error",
+                message=f"Failed to register LoRA adapter {adapter_name}")
+
+        if adapter_pin:
+            result = _service.pin_lora(adapter_name)
+            if not result:
+                return Output().error(
+                    "error",
+                    message=f"Failed to pin LoRA adapter {adapter_name}")
+    except (ValueError, RuntimeError) as e:
```
siddvenk, line 556: For the Memory errors, is there any chance we have access to specific error types in the backend rather than relying on parsing the strings/cause from the exception to determine whether it is a Memory error or not?

engines/python/setup/djl_python/huggingface.py
```diff
@@ -498,39 +533,81 @@ def register_adapter(inputs: Input):
...
     del _service.adapter_registry[adapter_name]
-    if not is_rolling_batch_enabled(_service.hf_configs.rolling_batch):
-        _service.model.base_model.delete_adapter(adapter_name)
-    return Output()
+    result = _service.remove_lora(adapter_name)
+    if not result:
+        return Output().error(
+            "error", message=f"Failed to remove LoRA adapter {adapter_name}")
+    logging.info(f"Unregistered adapter {adapter_name} successfully")
+    return Output(message=f"Adapter {adapter_name} unregistered")
```
siddvenk, line 610: I think we need think about this api a bit. Some questions I have:

- We should probably validate that the adapter_name exists in the adapter registry before attempting to delete it. 
- What happens if service.remove_lora fails? Why would this call fail? In scenarios where this fails, what does that mean for the adapter state within the engine? Is there a chance that this adapter continues to occupy memory?

engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
```diff
@@ -217,16 +217,25 @@ def supports_speculative_decoding() -> bool:
...
+                        long_lora_max_len: Optional[int],
+                        base_model_name: Optional[str]) -> LoRARequest:
+    params = {
+        "lora_name": lora_name,
+        "lora_int_id": lora_id,
+        "lora_path": lora_path
+    }
+    if long_lora_max_len is not None:
+        params["long_lora_max_len"] = long_lora_max_len
+    if base_model_name is not None:
```
siddvenk, line 230: Do we need to support this parameter? What happens if base_model_name is different than the current model we currently have hosted in the engine?

serving/docs/adapters_api.md
```diff
@@ -29,6 +31,22 @@ curl -X POST "http://localhost:8080/models/adaptecho/adapters?name=a1&src=..."
 }
 ```
 
+### Update an adapter
+
+`POST /models/{model_name}/adapters/{adapter_name}/update`
+
+* src - The adapter src. It currently requires a file, but eventually an id or URL can be supported depending on the model handler.
+* pin (optional): Whether to pin the adapter.
```
siddvenk, line 39: can we add a short description of what pinning means? if it's already in the docs elsewhere that's good too.

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
```diff
@@ -37,9 +46,14 @@
 /** A class handling inbound HTTP requests to the management API for adapters. */
 public class AdapterManagementRequestHandler extends HttpRequestHandler {
 
+    private static final Logger logger =
+            LoggerFactory.getLogger(AdapterManagementRequestHandler.class);
+
     static final Pattern ADAPTERS_PATTERN =
             Pattern.compile("^(/models/[^/^?]+)?/adapters([/?].*)?");
 
+    private final ConfigManager config = ConfigManager.getInstance();
```
siddvenk, line 55: we shouldn't make this an object property, let's just fetch the instance directly in `OnException` since that's the only place we use it

siddvenk:
I've only made it halfway through the PR so far, but wanted to raise some questions/thoughts on the python side first. I'll take a look at the java side and add some more comments.

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
xyang16: Changed.

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,96 @@ def register_adapter(inputs: Input):
     """
     Registers lora adapter with the model.
     """
+    if not is_rolling_batch_enabled(_service.hf_configs.rolling_batch):
```
siddvenk, line 499: I think we should further restrict this to `lmi-dist` and `vllm`. I don't know if we have the capabilities in scheduler to do multi-lora (unless you added them)

maybe we add a method `is_lora_enabled` and validate that lmi-dist or vllm are being used. I know you also added that enable_lora param, so we can use that in validation as well.

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,96 @@ def register_adapter(inputs: Input):
...
+    result = _service.rolling_batch.add_lora(adapter_name, adapter_path)
+    if adapter_pin:
+        try:
+            if not result:
+                return Output().error(
+                    "error",
+                    message=f"Failed to register LoRA adapter {adapter_name}")
+            result = _service.rolling_batch.pin_lora(adapter_name)
+            if not result:
+                return Output().error(
```
siddvenk, line 522: In this case, we have added the lora successfully, but the pinning was unsuccessful. I think we need to remove the lora adapter in this case before returning error

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,96 @@ def register_adapter(inputs: Input):
...
         raise ValueError(
             f"Only local LoRA models are supported. {adapter_path} is not a valid path"
         )
     logging.info(f"Registering adapter {adapter_name} from {adapter_path}")
     _service.adapter_registry[adapter_name] = inputs
+
+    result = _service.rolling_batch.add_lora(adapter_name, adapter_path)
+    if adapter_pin:
+        try:
+            if not result:
```
siddvenk, line 516: this should be moved before the if adapter_pin check

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,96 @@ def register_adapter(inputs: Input):
...
+            "LoRA adapter API is only supported for rolling batch.")
 
-        if isinstance(_service.hf_pipeline, Pipeline):
-            _service.hf_pipeline.model = _service.model
+    adapter_name = inputs.get_property("name")
+    adapter_path = inputs.get_property("src")
+    adapter_pin = inputs.get_property("pin") == "true"
+    if adapter_name not in _service.adapter_registry:
+        raise ValueError(f"Adapter {adapter_name} not registered.")
+    _service.adapter_registry[adapter_name] = inputs
```
siddvenk, line 552: If the src changes, is it sufficient to just update the adapter registry? Or do we need to unregister the previous adapter and re-register this one?

engines/python/src/test/resources/adaptecho/serving.properties
```diff
@@ -2,3 +2,4 @@ option.pythonExecutable=python3
 option.entryPoint=model.py
 option.handler=handle
 retry_threshold=0
+option.enable_lora=true
```
siddvenk, line 5: nit: new line

engines/python/setup/djl_python/huggingface.py
siddvenk: let's add a validation that src is the same from what we have already, and what is now provided in the inputs

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,96 @@ def register_adapter(inputs: Input):
...
+                    "error",
+                    message=f"Failed to pin LoRA adapter {adapter_name}")
+        except (ValueError, RuntimeError) as e:
+            if any(msg in str(e)
+                   for msg in ("No free lora slots",
+                               "greater than the number of GPU LoRA slots")):
+                raise MemoryError(str(e))
+            return Output().error(f"Failed to pin LoRA adapter {adapter_name}",
+                                  message=str(e))
+
```
siddvenk, line 568: let's add an unsupported operation case for when an update is requesting a currently pinned adapter to be unpinned since we don't support that at the engine level

engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
siddvenk: spoke offline - we'll just remove base_model_name for now since it isn't used by the engine either

engines/python/setup/djl_python/huggingface.py
```diff
@@ -498,39 +533,81 @@ def register_adapter(inputs: Input):
     """
     adapter_name = inputs.get_property("name")
     adapter_path = inputs.get_property("src")
+    adapter_pin = inputs.get_property("pin") == "true"
```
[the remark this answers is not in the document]
xyang16: Changed.

engines/python/setup/djl_python/huggingface.py
xyang16: Removed for non rolling batch.

engines/python/setup/djl_python/huggingface.py
xyang16: I hope there's better way. Right now the error is thrown as:

https://github.com/vllm-project/vllm/blob/main/vllm/lora/models.py#L375-L376

https://github.com/vllm-project/vllm/blob/main/vllm/lora/worker_manager.py#L190-L193

I agree this way of catching is error prone if vllm change the error message. But I can't find other way.

engines/python/setup/djl_python/huggingface.py
xyang16: Added remove_lora in case of error.

engines/python/setup/djl_python/huggingface.py
xyang16: Done

engines/python/setup/djl_python/huggingface.py
xyang16: raise NotImplementedError

engines/python/src/test/resources/adaptecho/serving.properties
xyang16: Added new line

engines/python/setup/djl_python/huggingface.py
```diff
@@ -496,41 +496,120 @@ def register_adapter(inputs: Input):
...
+        raise ValueError(f"Adapter {adapter_name} not registered.")
     del _service.adapter_registry[adapter_name]
-    if not is_rolling_batch_enabled(_service.hf_configs.rolling_batch):
-        _service.model.base_model.delete_adapter(adapter_name)
-    return Output()
+
+    try:
+        removed = _service.rolling_batch.remove_lora(adapter_name)
+        if not removed:
+            logging.debug(
```
siddvenk, line 605: let's make this an info level for now - most endpoints are going to be deploying at that level and would be good to know initially how frequently this occurs

xyang16: Done.

siddvenk:
changes lgtm - can we add a unit test for updateAdapter in ModelServerTest using the adaptecho model? After that i'm good to approve and merge

xyang16:
> changes lgtm - can we add a unit test for updateAdapter in ModelServerTest using the adaptecho model? After that i'm good to approve and merge

Added unit test.

siddvenk (approved):

PR #2460, opened 2024-10-17
Add test case for first token last token bug

brndysgit:
Funny bug where if a first token is also the last token (index 0) then we will fail to get the next token as it checks for "None". This adds a test case to ensure we took care of that edge case.

### DIFF
--- a/engines/python/setup/djl_python/tests/test_rolling_batch.py
+++ b/engines/python/setup/djl_python/tests/test_rolling_batch.py
@@ -37,6 +37,48 @@ class TestRollingBatch(unittest.TestCase):
             self.assertEqual(json.dumps({"generated_text": "Hello world"}),
                              req.get_next_token())
 
+    def test_last_token_is_first_token(self):
+        req_input1 = TextInput(
+            request_id=0,
+            input_text="This is a wonderful day",
+            parameters={
+                "max_new_tokens": 256,
+                "details": True
+            },
+            output_formatter=_json_output_formatter,
+        )
+        req1 = Request(req_input1)
+        req_input2 = TextInput(
+            request_id=1,
+            input_text="This is a wonderful day",
+            parameters={
+                "max_new_tokens": 256,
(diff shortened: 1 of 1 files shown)


sindhuvahinis (approved):

engines/python/setup/djl_python/tests/test_rolling_batch.py
```diff
@@ -37,6 +37,48 @@ def test_json_fmt(self):
...
+                "max_new_tokens": 256,
+                "stream": False,
+                "details": True
+            },
+        )
+        req2 = Request(req_input2)
+        for req in [req1, req2]:
+            req.request_output.set_next_token(Token(244, "He", -0.334532),
+                                              is_last_token=True,
+                                              finish_reason="length")
```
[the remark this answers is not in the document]
brndysgit: yes, fixed!

engines/python/setup/djl_python/tests/test_rolling_batch.py
```diff
@@ -37,6 +37,53 @@ def test_json_fmt(self):
...
+        request_id=0,
+        input_text="This is a wonderful day",
+        parameters={
+            "max_new_tokens": 256,
+            "details": True
+        },
+        output_formatter=_json_output_formatter,
+    )
+    req1 = Request(req_input1)
+    req_input2 = TextInput(
```
sindhuvahinis, line 52: What's the diff between req_input1 and req_input2? Both are using the same output_formatter.

sindhuvahinis: There looks like an indentation problem.

brndysgit: ah sorry, i think i was playing around with Stream and forgot to remove the test. there's no reason to have req 2 if stream isn't making any functional difference.

sindhuvahinis (approved):

PR #2467, opened 2024-10-19
[ci][tests] adding new multimodal and pipeline parallel tests

siddvenk:
## Description ##

There are a few changes in this PR, mostly related to adding new tests

1. Add multimodal tests for vllm backend (previously we only ran them for lmi-dist)
  - remove paligemma test - this model does not contain a chat template, and we only support multi-modal requests via chat api. It used to work until transformers removed default chat template behavior
3. Add pixtral tests for both lmi-dist and vllm
  - Note that succes of these tests depends on this PR https://github.com/deepjavalibrary/djl-serving/pull/2468/files
4. Add Mllama-11b test for vllm
  - This doesn't work with lmi-dist currently
5. Add a variety of pipeline parallel tests for lmi-dist
  - tp1, pp1, 4 workers
  - tp1, pp2, 2 workers
  - tp2, pp1 2 workers
  - pp4
  - tp2 pp2
  - tp2 pp2 + spec dec
6. Mount aws env vars in launch_container script to make it easy to run these tests with local credentials

### DIFF
--- a/tests/integration/launch_container.sh
+++ b/tests/integration/launch_container.sh
@@ -213,6 +213,9 @@ elif $is_sm_neo_context; then
     -e SM_NEO_CACHE_DIR=/opt/ml/compilation/cache \
     -e SM_NEO_HF_CACHE_DIR=/opt/ml/compilation/cache \
     -e COMPILER_OPTIONS={} \
+    -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+    -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
+    -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN \
     ${env_file} \
     ${runtime:+--runtime="${runtime}"} \
     ${shm:+--shm-size="${shm}"} \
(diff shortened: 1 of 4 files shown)


tosterberg (approved):

tests/integration/launch_container.sh
```diff
@@ -213,6 +213,9 @@ elif $is_sm_neo_context; then
     -e SM_NEO_CACHE_DIR=/opt/ml/compilation/cache \
     -e SM_NEO_HF_CACHE_DIR=/opt/ml/compilation/cache \
     -e COMPILER_OPTIONS={} \
+    -e AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID \
+    -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
+    -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN \
```
sindhuvahinis, line 218: @siddvenk  Why do we need explicit setting? -v ~/.aws:/root/.aws \ this is supposed to work, I believe. Do we need session token now

siddvenk: This is just to make it easier for local testing

sindhuvahinis: Oh got it. Thanks!

PR #2470, opened 2024-10-20
[ci][trt] add llama3-1 test for trtllm

siddvenk:
## Description ##

Llama-3-1 test for trtllm

### DIFF
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -686,7 +686,12 @@ trtllm_model_spec = {
         "batch_size": [1, 8],
         "seq_length": [256],
         "tokenizer": "google/flan-t5-xl",
-    }
+    },
+    "llama-3-1-8b": {
+        "batch_size": [1, 8],
+        "seq_length": [256],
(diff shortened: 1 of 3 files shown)


tests/integration/llm/client.py
```diff
@@ -686,7 +686,12 @@ def get_model_name():
         "batch_size": [1, 8],
         "seq_length": [256],
         "tokenizer": "google/flan-t5-xl",
-    }
+    },
+    "llama-3-1-8b": {
+        "batch_size": [1, 8],
+        "seq_length": [256],
+        "tokenizer": "NousResearch/Meta-Llama-3-8B",
```
sindhuvahinis, line 693: Should we use NousResearch/Meta-Llama-3.1-8B? Or both are same?

siddvenk: yea that's a good catch - will update

sindhuvahinis (approved):

tests/integration/llm/client.py
sindhuvahinis: Okay, approved it for now. So you could update it and merge

PR #2472, opened 2024-10-21
[lmi] add minimal requestId logging for new requests, and exceptions/errors

siddvenk:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/input_parser.py
+++ b/engines/python/setup/djl_python/input_parser.py
@@ -78,6 +78,7 @@ def parse_input_with_formatter(inputs: Input, **kwargs) -> ParsedInput:
     input_formatter_function = configs.input_formatter if configs.input_formatter else format_input
     for i in range(start_batch_id, len(batch)):
         input_item = batch[i]
+        client_request_id = input_item.get_property("requestId")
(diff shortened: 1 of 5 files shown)


engines/python/setup/djl_python/input_parser.py
```diff
@@ -78,22 +78,27 @@ def parse_input_with_formatter(inputs: Input, **kwargs) -> ParsedInput:
     input_formatter_function = configs.input_formatter if configs.input_formatter else format_input
     for i in range(start_batch_id, len(batch)):
         input_item = batch[i]
+        request_id = input_item.get_property("requestId")
```
sindhuvahinis, line 81: So, will the customers give us the request_id?

siddvenk: Yes, there are details in https://github.com/deepjavalibrary/djl-serving/pull/2363. We expose a property `request_id_header_key` that allows users to specify which header to use as the requestId. If not provided, we default to `X-Amzn-SageMaker-Inference-Id` on SageMaker, or `x-request-id` otherwise

sindhuvahinis (approved):

PR #2473, opened 2024-10-21
[lmi] avoid creating huggingface model config in rolling batch cases …

siddvenk:
…to avoid issues with mistral type models

## Description ##

Some mistral models are not vended in the HuggingFace pre-trained format (i.e. there is no config.json). They also contain different tokenizers that are not compatible with PreTrainedTokenizer from huggingface.

This change removes the instantiation of the huggingface config from the rolling batch path. The only place the model config is used in our rolling batch path is in lmi-dist to figure out whether the model is t5 or not. This is needed to address the failing pixtral MultiModal tests.

The reason these tests were passing locally is because I was using model artifacts mounted in the container at path `/opt/ml/model/pixtral`. There is logic in AutoConfig.from_pretrained to see if the model_id/path contains a known model type, and if so use that class to create it. In the integ test case we are using an s3 url that gets downloaded and stored in a directory without the pixtral name, so it fails.

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -870,7 +870,8 @@ vllm_model_list = {
         "option.max_rolling_batch_size": 16,
         "option.tokenizer_mode": "mistral",
         "option.limit_mm_per_prompt": "image=4",
-        "option.entryPoint": "djl_python.huggingface"
+        "option.entryPoint": "djl_python.huggingface",
+        "option.tensor_parallel_degree": "max"
     },
     "llama32-11b-multimodal": {
         "option.model_id": "s3://djl-llm/llama-3-2-11b-vision-instruct/",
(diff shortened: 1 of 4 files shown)


tests/integration/llm/prepare.py
```diff
@@ -870,7 +870,8 @@
         "option.max_rolling_batch_size": 16,
         "option.tokenizer_mode": "mistral",
         "option.limit_mm_per_prompt": "image=4",
-        "option.entryPoint": "djl_python.huggingface"
+        "option.entryPoint": "djl_python.huggingface",
+        "option.tensor_parallel_degree": "max"
```
sindhuvahinis, line 874: QQ: Is tp_degree needed here? Since the default should be max?

sindhuvahinis (approved):

tests/integration/llm/prepare.py
siddvenk: i should maybe add a comment here. For this specific case it is needed. This test is using vllm, so not MPI mode meaning it doesn't get the default tp max logic. It's also not a typical huggingface model so it doesn't go through the auto configuration path which would set tp degree.

PR #2482, opened 2024-10-23
[lmi] add t5 integration test for lmi-dist

siddvenk:
## Description ##

t5 was broken in lmi-dist. David fixed it, and i validated it locally with the new lmi wheel

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -80,7 +80,7 @@ jobs:
 
   test:
     runs-on: [ "${{ matrix.test.gh-runner && matrix.test.instance || 'self-hosted' }}", "${{ matrix.test.instance }}" ]
-    timeout-minutes: 60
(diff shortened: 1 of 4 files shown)


davidthomas426:
By the way, just so this is documented somewhere for the future with this PR, these were the two djl-serving PRs that led to the issue (which needed lmi_dist updates to fix, to align with the new djl-serving code):
- https://github.com/deepjavalibrary/djl-serving/pull/2259 - T5 in particular didn't set the `Logprob` `decoded_token` field
- https://github.com/deepjavalibrary/djl-serving/pull/2445 - T5 in particular would hang with use_passive_workers because it always created the non-use_passive_workers Engine class

Also, we have a new Known Issue that this PR doesn't fix: `"details": true` parameter does not work with T5--you get the same output with or without it, which is 1 logprob.

PR #2492, opened 2024-10-25
[ci] add integration test for spec-dec vllm

sindhuvahinis:
## Description ##

### DIFF
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -215,6 +215,19 @@ transformers_neuronx_handler_list = {
         "option.n_positions": 1024,
         "option.model_loading_timeout": 2400,
     },
+    "tiny-llama-vllm-speculative-rb": {
+        "option.model_id": "s3://djl-llm/llama-2-tiny/",
+        "option.speculative_draft_model": "s3://djl-llm/llama-2-tiny/",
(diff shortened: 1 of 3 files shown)


tests/integration/llm/prepare.py
```diff
@@ -215,6 +215,19 @@
         "option.n_positions": 1024,
         "option.model_loading_timeout": 2400,
     },
+    "tiny-llama-vllm-speculative-rb": {
+        "option.model_id": "s3://djl-llm/llama-2-tiny/",
+        "option.speculative_draft_model": "s3://djl-llm/llama-2-tiny/",
```
tosterberg, line 220: I think this test will require a bigger model, or minimally a change to context length estimates. Tiny llama has a max embedding size of 2048, so going to 16k on the context estimates should be a no op.

sindhuvahinis: For bigger model, we need already pre-compiled model. TinyLlama by itself takes 5 mins for compilation. Maybe I could change it 8B with tiny llama, with pre-compiled models.

tosterberg (approved):

sindhuvahinis:
Waiting for new artifacts updates.

sindhuvahinis:
This got deferred.

PR #2496, opened 2024-10-28
[lmi][v12][docs] update lmi documentation for v12

siddvenk:
## Description ##

Changes:
- Move all usages of `engine=MPI` to `engine=Python, option.mpi_mode=true`
- updates to the LMI-Dist breaking changes doc
- Removes the hardcoded list of supported models in lmi-dist/vllm in favor of a link to vllm's docs.

This is the first PR related to updating LMI docs for v12. There will be additional PRs that address:
- TensorRT-LLM update
- Neuron update

Additionally, after these release we need to do the following:
- Update the deployment guide with guidance for v12+
- Move configuration usages to either all Environment variables, or all serving.properties. We should still describe that both configuration mechanisms are supported, but there's no need to duplicate these configurations everywhere.

### DIFF
--- a/serving/docs/lmi/user_guides/vision_language_models.md
+++ b/serving/docs/lmi/user_guides/vision_language_models.md
@@ -1,10 +1,7 @@
 # Vision Language Models in LMI
 
-> [!WARNING]
-> Vision Language Model support is currently experimental in v0.29.0.
-> LMI currently only supports using VLMs with the lmi-dist and vllm backends.
 
-Starting with v0.29.0, LMI supports deploying the following types of Vision Language Models:
+LMI supports deploying the following types of Vision Language Models:
 
 * llava (e.g. llava-hf/llava-1.5-7b-hf)
 * llava_next (e.g. llava-hf/llava-v1.6-mistral-7b-hf)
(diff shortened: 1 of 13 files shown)


serving/docs/lmi/user_guides/vision_language_models.md
```diff
@@ -1,17 +1,16 @@
 # Vision Language Models in LMI
 
-> [!WARNING]
-> Vision Language Model support is currently experimental in v0.29.0.
-> LMI currently only supports using VLMs with the lmi-dist and vllm backends.
 
-Starting with v0.29.0, LMI supports deploying the following types of Vision Language Models:
+LMI supports deploying the following types of Vision Language Models:
```
sindhuvahinis, line 4: Should we specify about llama 3.2 with custom scripts support here?

sindhuvahinis (approved):

serving/docs/lmi/user_guides/vision_language_models.md
siddvenk: I think I'll add something for that here. I'm planning on making some demo notebooks, and then I'll link those here.

PR #2500, opened 2024-10-30
[docker] Update dependencies in dockerfile

xyang16:
## Description ##

### DIFF
--- a/serving/docker/pytorch-gpu.Dockerfile
+++ b/serving/docker/pytorch-gpu.Dockerfile
@@ -21,6 +21,8 @@ ARG torch_vision_version=0.19.0
 ARG onnx_version=1.19.0
 ARG python_version=3.10
 ARG numpy_version=1.26.4
+ARG pydantic_version=2.8.2
+ARG djl_converter_wheel="https://publish.djl.ai/djl_converter/djl_converter-0.31.0-py3-none-any.whl"
 
 RUN mkdir -p /opt/djl/conf && \
     mkdir -p /opt/ml/model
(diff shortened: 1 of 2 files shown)


frankfliu (approved):

serving/docker/pytorch-gpu.Dockerfile
```diff
@@ -44,7 +46,7 @@ ENV PYTORCH_FLAVOR=cu124-precxx11
 ENV TORCH_CUDNN_V8_API_DISABLED=1
 ENV JAVA_OPTS="-Xmx1g -Xms1g -XX:+ExitOnOutOfMemoryError -Dai.djl.default_engine=PyTorch"
 ENV HF_HOME=/tmp/.cache/huggingface
-ENV TRANSFORMERS_CACHE=/tmp/.cache/huggingface/transformers
```
sindhuvahinis, line 47: @xyang16 Don't we need TRANSFORMERS_CACHE anymore?

serving/docker/pytorch-gpu.Dockerfile
```diff
@@ -21,6 +21,8 @@ ARG torch_vision_version=0.19.0
 ARG onnx_version=1.19.0
 ARG python_version=3.10
 ARG numpy_version=1.26.4
+ARG pydantic_version=2.8.2
+ARG djl_converter_wheel="https://publish.djl.ai/djl_converter/djl_converter-0.31.0-py3-none-any.whl"
```
sindhuvahinis, line 25: Curious, What is this for?

serving/docker/pytorch-gpu.Dockerfile
frankfliu: This is deprecated, only need HF_HOME

serving/docker/pytorch-gpu.Dockerfile
frankfliu: 1. `djl-converter` is used to convert huggingface model automatically to OnnxRuntime/Rust on the fly to optimize the performance
2. we depends on peft and pydantic when using `huggingface.py`, we should consider move them to where it used. right now they are hard dependency if run any python model even not related to peft in the container.

sindhuvahinis (approved):

PR #2508, opened 2024-10-31
[lora] Support SageMaker adapter header

xyang16:
## Description ##

### DIFF
--- a/engines/python/setup/djl_python/input_parser.py
+++ b/engines/python/setup/djl_python/input_parser.py
@@ -210,7 +210,7 @@ def parse_adapters(request_input: TextInput, input_item: Input,
                                kwargs.get("adapter_registry"))
         else:
             # inference with just base model.
-            adapters_per_item = [""] * input_len
+            adapters_per_item = [("", "")] * input_len
 
         if input_len != len(adapters_per_item):
             raise ValueError(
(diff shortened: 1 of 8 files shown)


wlm/src/main/java/ai/djl/serving/wlm/ModelInfo.java
```diff
@@ -298,6 +298,7 @@ public void load(Device device) throws ModelException, IOException {
                                                 Adapter.newInstance(
                                                         this,
                                                         adapterName,
+                                                        null,
```
siddvenk, line 301: I think by default we should set alias to the name, not null. We basically have that now with the `getAlias` call, but i'd prefer we set the alias explicitly to the name rather than the condition check there.

That way we can also avoid passing alias as an argument to unregisterAdapter in ModelInfo

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
```diff
@@ -89,7 +89,7 @@ protected void handleRequest(
                     handleListAdapters(ctx, decoder, modelName);
                     return;
                 } else if (HttpMethod.POST.equals(method)) {
-                    handleRegisterAdapter(ctx, decoder, modelName);
+                    handleRegisterAdapter(ctx, req, decoder, modelName);
```
siddvenk, line 92: I think we should look into creating a requestParser, or leveraging the existing requestParser like we do in the InferenceRequestHandler. We can consolidate a lot of the code in that parser and leverage it for all the handle*Adapter methods

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
```diff
@@ -228,9 +233,11 @@ private void handleUpdateAdapter(
         }
 
         Adapter<Input, Output> adapter = modelInfo.getAdapter(adapterName);
+        String alias = req.headers().get("X-Amzn-SageMaker-Adapter-Alias");
 
         if (adapter == null) {
-            throw new BadRequestException(404, "The adapter " + adapterName + " was not found");
+            throw new BadRequestException(
+                    404, "The adapter " + (alias == null ? adapterName : alias) + " was not found");
```
siddvenk, line 240: alias should always be provided, right? If so, we should validate. I know that behavior may change in the future, but for now I believe it is required. Correct me if i'm wrong

serving/src/main/java/ai/djl/serving/http/AdapterManagementRequestHandler.java
xyang16: Done

engines/python/setup/djl_python/input_parser.py
```diff
@@ -229,15 +229,17 @@ def parse_adapters(request_input: TextInput, input_item: Input,
...
 
     # check content, possible in workflow approach
     if input_item.contains_key("adapter"):
-        adapters_per_item = input_item.get_as_string("adapter")
+        adapters_per_item = (input_item.get_as_string("adapter"), None)
 
     # check properties, possible from header
-    if "adapter" in input_item.get_properties():
-        adapters_per_item = input_item.get_properties()["adapter"]
+    if "X-Amzn-SageMaker-Adapter-Identifier" in input_item.get_properties():
```
siddvenk, line 239: do we need to parse these header here? I thought we already parse and convert on the frontend?

siddvenk (approved):

PR #2524, opened 2024-11-06
Introducing SageMaker Fast Model Loader

Lokiiiiii:
## Description ##

Introducing SageMaker Fast Model Loader and the associated scripts for Model Sharding

### DIFF
--- a/serving/docker/partition/sm_neo_dispatcher.py
+++ b/serving/docker/partition/sm_neo_dispatcher.py
@@ -53,7 +53,7 @@ class NeoDispatcher:
         return properties
 
     def is_valid_sharding_config(self):
-        # TODO: once pp is supported, add add requirement and remove default value
+        # TODO: once pp is supported, add requirement and remove default value
         return (self.properties.get("option.load_format") in VALID_LOAD_FORMATS
                 and self.properties.get("option.tensor_parallel_degree")
                 is not None)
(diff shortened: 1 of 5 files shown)


serving/docker/partition/sm_neo_dispatcher.py
```diff
@@ -0,0 +1,129 @@
...
+        properties = update_kwargs_with_env_vars({})
+        properties.update(
+            load_properties(neo_environ["SM_NEO_INPUT_MODEL_DIR"]))
+        return properties
+
+    def is_valid_sharding_config(self):
+        # TODO: once pp is supported, add add requirement and remove default value
+        return (self.properties.get("option.load_format") in VALID_LOAD_FORMATS
+                and self.properties.get("option.tensor_parallel_degree")
+                is not None)
```
siddvenk, line 59: how would this work today if someone did specify pp=2 for example? do we need to validate and prevent that specifically?

serving/docker/partition/sm_neo_shard.py
```diff
@@ -0,0 +1,168 @@
...
+                pp_degree=pp_degree,
+                tp_degree=tp_degree,
+                input_dir=input_dir,
+                output_dir=output_dir,
+                configs=configs,
+            )
+
+        del engine
+        torch.cuda.empty_cache()
+        # MPI.COMM_WORLD.Barrier()
```
siddvenk, line 131: can remove, or useful to keep around for debugging? same with below on 162

serving/docker/partition/sm_neo_dispatcher.py
Lokiiiiii: yes Fast Model Loading is not supported with PP for this release. It is WIP. 

I will add a validation to fail fast.

siddvenk (approved):

PR #2534, opened 2024-11-08
[docker] Fix the pytorch JNI version

xyang16:
## Description ##

Previously pytorch python library and JNI library have the same version. So we have `PYTORCH_VERSION=${torch_version}`. Now since pytorch python is 2.5, need to set PYTORCH_VERSION to 2.4


serving/docker/lmi.Dockerfile
```diff
@@ -58,7 +58,7 @@ ENV DJL_CACHE_DIR=/tmp/.djl.ai
 ENV LD_LIBRARY_PATH=/usr/local/lib/python${python_version}/dist-packages/nvidia/cudnn/lib/
 ENV PYTORCH_LIBRARY_PATH=/usr/local/lib/python${python_version}/dist-packages/torch/lib
 ENV PYTORCH_PRECXX11=true
-ENV PYTORCH_VERSION=${torch_version}
+ENV PYTORCH_VERSION=2.4.0
```
sindhuvahinis, line 61: Maybe use ${djl_torch_version}. Will this affect lmi's pytorch version? I assume lmi wheel installs the pytorch version so this env should not affect lmi-dist, right? @xyang16 @siddvenk

xyang16: This PYTORCH_VERSION is only used by JNI library.

xyang16: This is what I mentioned about pytorch JNI version and wheel version will diverge in the decouple meeting.

sindhuvahinis, line 61:
```suggestion
ENV PYTORCH_VERSION=${djl_torch_version}
```

siddvenk: oops yea i missed that

siddvenk (approved):

sindhuvahinis (approved):

PR #2530, opened 2024-11-07
Fixing Neo integ tests & adding more input validations

Lokiiiiii:
## Description ##

Brief description of what this PR is about

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

PR #2535, opened 2024-11-08
++ #2530 Fixing Neo integ tests & adding more input validations 

HappyAmazonian:
## Description ##
did a format based on https://github.com/deepjavalibrary/djl-serving/pull/2530

### DIFF
--- a/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
@@ -88,7 +88,7 @@ class LmiDistRbProperties(Properties):
 
     @model_validator(mode='after')
     def validate_speculative_and_fml(self):
-        if self.load_format == LmiDistLoadFormats.sagemaker_fast_model_loader and self.speculative_draft_model:
+        if self.load_format == LmiDistLoadFormats.sagemaker_fast_model_loader.value and self.speculative_draft_model:
             raise AssertionError(
                 f"Cannot enable sagemaker_fast_model_loader and speculative decoding at the same time"
             )
--- a/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
+++ b/engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
@@ -31,6 +31,10 @@ class LmiDistQuantizeMethods(str, Enum):
     squeezellm = 'squeezellm'
 
 
+class LmiDistLoadFormats(str, Enum):
+    sagemaker_fast_model_loader = 'sagemaker_fast_model_loader'
+
+
 class LmiDistRbProperties(Properties):
     engine: Optional[str] = None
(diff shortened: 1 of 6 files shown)


serving/docker/partition/sm_neo_dispatcher.py
```diff
@@ -53,10 +53,26 @@ def _get_serving_properties(self) -> dict:
...
-                and self.properties.get("option.tensor_parallel_degree")
-                is not None)
+        if self.properties.get("option.load_format") in VALID_LOAD_FORMATS:
+            if self.properties.get("option.quantize"):
+                raise ValueError(
+                    f"Sharding & Quantization are mutually exclusive. "
+                    f"Please Quantize first & then Shard. "
+                    f"Received load_format={self.properties.get('option.load_format')} "
+                    f"& quantize={self.properties.get('option.quantize')}.")
+            if not int(self.properties.get("option.tensor_parallel_degree")):
```
sindhuvahinis, line 63: int(None) -> would throw error.

engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
```diff
@@ -82,6 +86,14 @@ def validate_speculative_and_lora(self):
             )
         return self
 
+    @model_validator(mode='after')
+    def validate_speculative_and_fml(self):
+        if self.load_format == LmiDistLoadFormats.sagemaker_fast_model_loader and self.speculative_draft_model:
```
sindhuvahinis, line 91: self.load_format seems to be str. So you might need to change to `self.load_format == LmiDistLoadFormats.sagemaker_fast_model_loader.value`.  Or change the load_format type from str to LmiDistLoadFormats enum. I prefer the latter.

HappyAmazonian: it seems self.load_format  is used in other places. Let me do the .value change first.

engines/python/setup/djl_python/properties_manager/lmi_dist_rb_properties.py
```diff
@@ -82,6 +86,14 @@ def validate_speculative_and_lora(self):
             )
         return self
 
+    @model_validator(mode='after')
+    def validate_speculative_and_fml(self):
+        if self.load_format == LmiDistLoadFormats.sagemaker_fast_model_loader.value and self.speculative_draft_model:
+            raise AssertionError(
+                f"Cannot enable sagemaker_fast_model_loader and speculative decoding at the same time"
```
siddvenk, line 93: What is the customer facing name for fast_model_loader? We should probably use the recognizable name here so customers know exactly what the incompatibility is. 

Once we have that, I would suggest changing this to `speculative decoding and <fast model loading name> are currently not compatible. Disable speculative decoding, or use standard checkpoints for the target model`

siddvenk (approved):

PR #2536, opened 2024-11-08
[lora] Add load option to LoRA adapter API

xyang16:
## Description ##

Add an additional `load` option to the register adapter API and update adapter API.

The reason is for this change is to keep consistent with other model servers like vllm and lorax. vllm load_lora_adapter API don't load adapter weights, the adapter weights is only loaded when running inference using one particular adapter.

Discussed with Hosting team, making default to `true`.

Example:

```
curl -X POST "http://localhost:8080/models/model/adapters?name=eng_alpaca&load=false&src=/opt/ml/model/adapters/eng_alpaca"
```

### DIFF
--- a/engines/python/setup/djl_python/huggingface.py
+++ b/engines/python/setup/djl_python/huggingface.py
@@ -302,6 +302,38 @@ class HuggingFaceService(object):
                                  self.hf_configs.device, **parameters))
         return outputs
 
+    def add_lora(self, lora_name: str, lora_alias: str, lora_path: str):
+        if not is_rolling_batch_enabled(self.hf_configs.rolling_batch):
+            raise NotImplementedError(
+                "LoRA adapter API is only supported for rolling batch.")
+
+        loaded = self.rolling_batch.add_lora(lora_name, lora_path)
+        if not loaded:
+            raise RuntimeError(f"Failed to load LoRA adapter {lora_alias}")
+        return loaded
+
+    def remove_lora(self, lora_name: str, lora_alias: str):
+        if not is_rolling_batch_enabled(self.hf_configs.rolling_batch):
+            raise NotImplementedError(
+                "LoRA adapter API is only supported for rolling batch.")
+
+        removed = self.rolling_batch.remove_lora(lora_name)
+        if not removed:
+            logging.info(
+                f"Remove LoRA adapter {lora_alias} returned false, the adapter may have already been evicted."
+            )
+        return removed
+
+    def pin_lora(self, lora_name: str, lora_alias: str):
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/huggingface.py
```diff
@@ -546,39 +575,45 @@ def update_adapter(inputs: Input):
...
-            if not pinned:
-                raise RuntimeError(
-                    f"Failed to pin LoRA adapter {adapter_alias}")
+        old_adapter_load = old_adapter.get_as_string("load").lower(
+        ) == "true" if old_adapter.contains_key("load") else True
+        if adapter_load != old_adapter_load:
+            if adapter_load:
+                _service.add_lora(adapter_name, adapter_alias, adapter_path)
+            else:
+                _service.remove_lora(adapter_name, adapter_alias)
```
siddvenk, line 605: if adapter_load is false, why are we removing the adapter? Should this just be a noop?

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -283,5 +283,5 @@ def pin_lora(self, lora_name):
         # 2) An adapter is not evicted, call add_lora() is not necessary.
         # But since whether an adapter is evicted is not exposed outside of engine,
         # and add_lora() in this case will take negligible time, we will still call add_lora().
-        self.engine.add_lora(lora_request)
-        return self.engine.pin_lora(lora_request.lora_int_id)
+        return self.engine.add_lora(lora_request) and self.engine.pin_lora(
+            lora_request.lora_int_id)
```
siddvenk, line 287: do we need to check the result of add_lora before pinning?

Also, I would prefer if we kept these calls separate. It's more readable.

Sames questions for vlm_rolling_batch.py

engines/python/setup/djl_python/huggingface.py
xyang16: This is to have an unload option.

engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
xyang16: 1) This is to make sure it's successfully loaded before pinning.
2) Made these calls separate.

engines/python/setup/djl_python/huggingface.py
siddvenk: I see, so we support unloading only for unpinned adapters

xyang16: Yes

siddvenk (approved):

PR #2539, opened 2024-11-08
[test] fix transformers neuronx integration test failure

sindhuvahinis:
## Description ##

EDIT: Even tiny models have higher context length. So, we trained model and tokenizer to have smaller context length 4k on runtime. I trained the model offline and uploaded it to s3. So the previous fix is not needed. 

---

This should fix the tnx unit tests in integration tests failures. This is due to transformers getting upgraded to 4.45.2. This does not affect our handlers. This only affects, `train_new_from_iterator` in transformers. 

Raised PR in transformers to fix it. https://github.com/huggingface/transformers/pull/34661

### DIFF
--- a/engines/python/setup/djl_python/tests/neuron_test_scripts/test_neuron_tnx_rolling_batch.py
+++ b/engines/python/setup/djl_python/tests/neuron_test_scripts/test_neuron_tnx_rolling_batch.py
@@ -14,6 +14,8 @@
 import unittest
 import json
 import os
+import subprocess
+import glob
 
 try:
     import transformers_neuronx
@@ -34,6 +36,21 @@ expected_text_30 = {
         "The future of AI is bright, and it's already here. With the help of AI, we can create more personalized experiences, automate repetitive tasks, and even predict the future.",
     }
 }
+
+
+def download_model_from_s3(s3_url, model_name, download_dir=None):
+    if download_dir is None:
+        download_dir = f"/tmp/download/{model_name}"
+
+    commands = ["aws", "s3", "sync", s3_url, download_dir]
+
+    subprocess.run(commands)
+
+    # check if any file was downloaded.
+    if not glob.glob(os.path.join(download_dir, '*')):
+        raise Exception('Model download from s3url failed')
+
+    return download_dir
 
 
 @unittest.skipIf(SKIP_TEST, "Neuron dependencies are not available")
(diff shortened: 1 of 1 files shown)


siddvenk:
Do we need 4.45.2 for neuron at the moment? (I think yes for multimodal?)

engines/python/setup/djl_python/tests/neuron_test_scripts/tiny_models.py
```diff
@@ -118,6 +120,22 @@ def opt():
...
 
 
+def set_pre_tokenizer(tokenizer):
+    """
+    Adding this due to a bug in transformers in train_new_from_iterator since 4.45.2.
+    TODO: Remove this once the issue is fixed.
+    """
+    underlying_tokenizer = tokenizer.backend_tokenizer
+    if underlying_tokenizer.pre_tokenizer is None:
+        split_pattern = r"(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\r\n\p{L}\p{N}]?\p{L}+|\p{N}{1,3}| ?[^\s\p{L}\p{N}]+[\r\n]*|\s*[\r\n]+|\s+(?!\S)|\s+"
```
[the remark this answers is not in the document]
sindhuvahinis: This comes from Llama 3.2 1B tokenizer.json. https://huggingface.co/meta-llama/Llama-3.2-1B/blob/main/tokenizer.json. 
Here is the raw json version. Here i just converted into Object. 
```
  "pre_tokenizer": {
    "type": "Sequence",
    "pretokenizers": [
      {
        "type": "Split",
        "pattern": {
          "Regex": "(?i:'s|'t|'re|'ve|'m|'ll|'d)|[^\\r\\n\\p{L}\\p{N}]?\\p{L}+|\\p{N}{1,3}| ?[^\\s\\p{L}\\p{N}]+[\\r\\n]*|\\s*[\\r\\n]+|\\s+(?!\\S)|\\s+"
        },
        "behavior": "Isolated",
        "invert": false
      },
      {
        "type": "ByteLevel",
        "add_prefix_space": false,
        "trim_offsets": true,
        "use_regex": false
      }
    ]
  }
  ```

**Comment by sindhuvahinis** [comment]
> Do we need 4.45.2 for neuron at the moment? (I think yes for multimodal?)

Yes. For vllm 0.6.2 version, it is needed. https://github.com/deepjavalibrary/djl-serving/pull/2518 MultiModal support is also based on vllm 0.6.2

**Comment by siddvenk** [comment]
spoke offline - we're going to look into whether we need to do the tokenizer training stuff we currently do. If we can just use the tiny llama tokenizer directly, then that simplifies things here.

File: engines/python/setup/djl_python/tests/neuron_test_scripts/test_neuron_tnx_rolling_batch.py [hunk 2]
```diff
@@ -36,6 +38,21 @@
 }
 
 
+def download_model_from_s3(s3_url, model_name=None, download_dir=None):
```
siddvenk, line 41: seems like model_name is required, let's remove the `=None`?

sindhuvahinis: model_name is not required if they gave download_dir.

siddvenk (approved):

PR #2546, opened 2024-11-12
[neo] Allow setting GPU memory-related LMI options in sharding jobs

ethnzhng:
This PR allows customers to pass in values for the LMI options
- `option.gpu_memory_utilization` 
- `option.enforce_eager` 
- `option.max_rolling_batch_size `
- `option.max_model_len`

during Neo AOT sharding jobs. Changing the values of these options can potentially allow a customer to shard a model on an instance which would result in OOM when using default values. Note that the customized options need to be set to the same values when loading back the artifacts during runtime; the generated `serving.properties` in the output artifacts will ensure this parity.

### DIFF
--- a/serving/docker/partition/sm_neo_shard.py
+++ b/serving/docker/partition/sm_neo_shard.py
@@ -98,15 +98,19 @@ class NeoShardingService():
         # Use default LMI value of 0.9 for gpu_memory_utilization, unless set otherwise
         gpu_memory_utilization = float(
             self.properties.get("option.gpu_memory_utilization", 0.9))
+        # Use default LMI value of False for enforce_eager, unless set otherwise
+        enforce_eager: bool = str(
+            self.properties.get("option.enforce_eager",
+                                False)).lower() == "true"
 
         engine_args = VllmEngineArgs(
             model=input_dir,
             pipeline_parallel_size=pp_degree,
             tensor_parallel_size=tp_degree,
-            enforce_eager=True,
             disable_custom_all_reduce=True,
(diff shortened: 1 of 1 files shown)


serving/docker/partition/sm_neo_shard.py
```diff
@@ -103,6 +103,7 @@ def shard_lmi_dist_model(self, input_dir: str, output_dir: str,
             enforce_eager=True,
             disable_custom_all_reduce=True,
             distributed_executor_backend="mp",
+            gpu_memory_utilization=0.99,
```
siddvenk, line 106: While this may allow for the sharding of 405b on a p4de, will the resulting converted artifacts even be usable by the customer on the same instance? Do we also provide the necessary configs in the output to ensure the model can be loaded on the instance for inference?

ethnzhng: That is correct, it would have to be set to the same value during runtime. I updated the PR to simply allow customers to specify `option.gpu_memory_utilization` during sharding jobs, and with #2545 that value will be propagated to the `serving.properties`. Same with `option.enforce_eager`, since it also affects GPU mem utilization.


Thus customers wanting to use a model/instance combination which requires a change to `gpu_memory_utilization` or `enforce_eager` will have to be aware of this at the time of converting the artifacts.

siddvenk: Awesome, this sounds like the right approach!

siddvenk (approved):

PR #2547, opened 2024-11-12
Create a single shareable ClusterRequestHandler for all channels to prevent ssh-keygen race conditions. Add unit tests for the ClusterRequestHandler

si2d:
## Description ##

For multinode communication with > 2 nodes, it was possible to get into a race condition where two different channels try to run ssh keygen at the same time because both would fail the file-exists check. In this case, one of the keygen would be stuck asking if it should overwrite the previous file

```
~/workspace/djl/djl-serving   master  ssh-keygen -q -t rsa -N "" -f rsa_test
 ~/workspace/djl/djl-serving   master a ssh-keygen -q -t rsa -N "" -f rsa_test
rsa_test already exists.
Overwrite (y/n)?
```

This would cause the worker node to not be able to join the cluster and the leader node to be stuck waiting for all the workers.

With this change, we move to a single shareable ClusterRequestHandler which is used by all channels, and a `synchronize` before the file check to ensure we do not get into this condition. Moreover, we also add a timeout while we wait  for the process to end so that we don't deadlock next time. 

Also added a bunch of unit tests for new and existing functionalities of the ClusterRequestHandler to prevent future breakage
Tested by creating a container and running with 4 nodes in the cluster

### DIFF
--- a/serving/build.gradle.kts
+++ b/serving/build.gradle.kts
@@ -38,6 +38,7 @@ dependencies {
     testImplementation(libs.testng) {
         exclude(group = "junit", module = "junit")
     }
+    testImplementation(libs.mockito.core)
 }
 
 tasks {
--- a/serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
+++ b/serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
@@ -1,0 +1,267 @@
+/*
+ * Copyright 2024 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+ * with the License. A copy of the License is located at
+ *
+ * http://aws.amazon.com/apache2.0/
+ *
+ * or in the "license" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions
+ * and limitations under the License.
+ */
+package ai.djl.serving.http;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyLong;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static org.mockito.MockitoAnnotations.openMocks;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertTrue;
+
+import ai.djl.serving.Arguments;
+import ai.djl.serving.util.ConfigManager;
+import ai.djl.util.Utils;
+
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.netty.handler.codec.http.DefaultFullHttpRequest;
+import io.netty.handler.codec.http.DefaultFullHttpResponse;
+import io.netty.handler.codec.http.HttpMethod;
+import io.netty.handler.codec.http.HttpVersion;
+
+import org.apache.commons.cli.CommandLine;
+import org.mockito.ArgumentCaptor;
+import org.mockito.Mock;
+import org.mockito.invocation.InvocationOnMock;
+import org.testng.annotations.AfterClass;
+import org.testng.annotations.AfterMethod;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.io.BufferedWriter;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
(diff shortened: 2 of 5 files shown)


serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
```diff
@@ -0,0 +1,263 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
siddvenk, line 2: 2020 -> 2024

serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
```diff
@@ -0,0 +1,263 @@
...
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public class ClusterRequestHandlerTest {
+    private static final String TEST_DIRECTORY = "/tmp/test-clusterrequesthandler";
```
siddvenk, line 61: I was looking through our code to see how we handle tmp files like this in unit tests. I don't think we have a strong preference, but i think 'build/tmp' could be a good alternative here to keep in line with the repo like https://github.com/deepjavalibrary/djl-serving/blob/master/serving/src/test/java/ai/djl/serving/ConfigManagerTest.java#L59.

serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    @Mock private Process sshProcess;
+    private String expectedKey;
+    private String fileToDelete;
+    private String originalHome;
+
+    private ClusterRequestHandler handler;
+
+    @BeforeClass
+    public void setup() throws Exception {
+        originalHome = System.getProperty("user.home");
```
siddvenk, line 73: I see that we use this in the ClusterRequestHandler as the base path for sshdir. I think that makes sense as a default, but thoughts on making it configurable? Maybe we can discuss offline, curious what host dirs we mount to the container for multinode

serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    public void setupTest() throws Exception {
+        openMocks(this).close();
+        handler = new ClusterRequestHandler(cmd -> mockProcessBuilder.command(cmd));
+        setupMockProcessBuilder();
+    }
+
+    @AfterClass(alwaysRun = true)
+    public void tearDownClass() throws Exception {
+        System.setProperty("user.home", originalHome);
+        FileUtils.deleteDirectory(new File(TEST_DIRECTORY));
```
siddvenk, line 90: We can use `ai.djl.utils.Utils.deleteQuietly(<path>)`. We typically use that in our tests to recursively delete a directory and all it's contents (it does deletion without throwing any exceptions)

serving/build.gradle.kts
```diff
@@ -38,6 +38,7 @@ dependencies {
     testImplementation(libs.testng) {
         exclude(group = "junit", module = "junit")
     }
+    testImplementation("org.mockito:mockito-core:5.11.0")
```
siddvenk, line 41: -> `testImplementation(libs.mockito.core)`

we already declare the mockito version in gradle/libs.versions.toml (this needs to be updated in the wlm build.gradle.kts too, i can fix that separately)

serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
```diff
@@ -0,0 +1,263 @@
...
+        when(mockProcessBuilder.command(any(String[].class)))
+                .thenAnswer(this::processCommandInvocation);
+    }
+
+    private ProcessBuilder processCommandInvocation(InvocationOnMock invocation)
+            throws IOException {
+        String publicFile = invocation.getArgument(invocation.getArguments().length - 1) + ".pub";
+        try (FileWriter fileWriter = new FileWriter(publicFile)) {
+            fileWriter.write(expectedKey);
+        }
```
siddvenk, line 259: we typically use

```
try (BufferedWriter writer = Files.newBufferedWriter(publicFile)) {
    writer.write(expectedKey);
}
```

serving/src/test/java/ai/djl/serving/http/ClusterRequestHandlerTest.java
si2d: Let me make it configurable with this CR. I did not like updating System properties in the test

si2d: One thing is that if the path is changed after the handler is initialized, it would create new ssh key on the next invocation. Do we want to allow that? I can also limit this change to be a protected method, to be configurable with test only

siddvenk:
I think the unit tests are failing on windows because of `Charset.defaultCharset()`. I guess the default charset on windows is different -> let's update to `StandardCharsets.UTF-8`?

siddvenk (approved):

PR #2552, opened 2024-11-13
add Lora config to arg list in Neo sharding script& its integ test change

HappyAmazonian:
## Description ##

- pass Lora config to arg list  in Neo sharding script
- also add tinyllama fast model loading for related integ code

### DIFF
--- a/serving/docker/partition/sm_neo_shard.py
+++ b/serving/docker/partition/sm_neo_shard.py
@@ -108,12 +108,31 @@ class NeoShardingService():
             self.properties.get("option.gpu_memory_utilization", 0.9))
         enforce_eager: bool = str(
             self.properties.get("option.enforce_eager",
-                                False)).lower() == "true"
+                                "False")).lower() == "true"
         max_rolling_batch_size = int(
             self.properties.get("option.max_rolling_batch_size", 256))
         max_model_len = self.properties.get("option.max_model_len", None)
         if max_model_len is not None:
             max_model_len = int(max_model_len)
+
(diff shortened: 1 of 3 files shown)


serving/docker/partition/sm_neo_shard.py
```diff
@@ -108,13 +108,32 @@ def shard_lmi_dist_model(self, input_dir: str, output_dir: str,
...
-                                False)).lower() == "true"
+                                "False")).lower() == "true"
         max_rolling_batch_size = int(
             self.properties.get("option.max_rolling_batch_size", 256))
         max_model_len = self.properties.get("option.max_model_len", None)
         if max_model_len is not None:
             max_model_len = int(max_model_len)
 
+        # LoraConfigs
+        enable_lora: bool = str(
```
siddvenk, line 119: I think we should only be passing lora configs to the engine if the user specifies them. I'd prefer we rewrite this section as
```
lora_kwargs = {}
if self.properties.get("option.enable_lora"):
    lora_kwargs["enable_lora"] = self.properties.get("option.enable_lora").lower() == "true"
if self.properties.get("option.fully_sharded_loras"):
    lora_kwargs["fully_shareded_loras"] = option.get("option.fully_sharded_loras").lower() == "true"

engine_args = VllmEngineArgs(
    <prior configs>
    **lora_kwargs
)
...
```
and so on. This way we're only providing the lora configs to the engine if they have been specified

siddvenk (approved):

PR #2554, opened 2024-11-14
[lora] Add lora configs and validation

xyang16:
## Description ##


tests/integration/llm/prepare.py
```diff
@@ -538,6 +542,10 @@
         "true",
         "option.max_lora_rank":
         64,
+        "option.fully_sharded_loras":
+        "true",
+        "option.long_lora_scaling_factors":
+        "(4.0, )",
```
sindhuvahinis, line 548: Why does it have to be a strict tuple? Can we get comma separated strings and convert into tuple internally? This is what we do for context_length_estimate for tnx.

xyang16: Changed. Now we can get comma separated strings and convert into tuple.

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
```diff
@@ -51,7 +52,7 @@ class VllmRbProperties(Properties):
     max_lora_rank: Optional[int] = 16
     fully_sharded_loras: bool = False
     lora_extra_vocab_size: int = 256
-    long_lora_scaling_factors: Optional[Tuple[float]] = None
+    long_lora_scaling_factors: Optional[Tuple[float, ...]] = None
```
sindhuvahinis, line 55: QQ, `...` Is this to allow empty tuple to this?

sindhuvahinis (approved):

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
xyang16: ... allow for tuple with variable length. for example, (3.0,) (3.0,4.0)

Tuple[float] only allow for one element in the tuple.

siddvenk (approved):

PR #2566, opened 2024-11-16
[docker] Update torch to 2.5.1 in docker

xyang16:
## Description ##


frankfliu (approved):

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -13,7 +13,7 @@ ARG version=12.5.1-devel-ubuntu22.04
 FROM nvidia/cuda:$version
 ARG cuda_version=cu125
 ARG python_version=3.10
-ARG TORCH_VERSION=2.4.0
```
siddvenk, line 16: We can't update torch version in trtllm yet (unless we separate djl torch version and python torch version). trtllm still uses 2.4.0

siddvenk (approved):

PR #2569, opened 2024-11-17
[serving] Improve logging for exec shell command

xyang16:
## Description ##

Improve logging for exec shell command

### DIFF
--- a/wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
+++ b/wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
@@ -60,7 +60,7 @@ public final class LmiUtils {
         }
         int exitCode = exec.waitFor();
         if (0 != exitCode || logOutput.startsWith("ERROR ")) {
-            logger.info("{}", logOutput);
+            logger.error("{}", logOutput);
             throw new EngineException("Failed to execute: [" + String.join(" ", cmd) + "]");
         } else {
             logger.debug("{}", logOutput);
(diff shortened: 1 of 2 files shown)


xyang16:
Raising PR on behalf of Frank.

frankfliu (approved):

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -54,6 +52,21 @@ public final class LmiUtils {
 
     private LmiUtils() {}
 
+    static void exec(List<String> cmd) throws IOException, InterruptedException {
```
siddvenk, line 55: We can do this as a follow up, but might be nice to put this into a generic ProcessUtils class or something similar. We have a few more places where we do this (like cluster request handler)

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -54,6 +52,21 @@ public final class LmiUtils {
...
+        Process exec = new ProcessBuilder(cmd).redirectErrorStream(true).start();
+        String logOutput;
+        try (InputStream is = exec.getInputStream()) {
+            logOutput = Utils.toString(is);
+        }
+        int exitCode = exec.waitFor();
+        if (0 != exitCode || logOutput.startsWith("ERROR ")) {
+            logger.info("{}", logOutput);
```
siddvenk, line 63: should this be logged at error level?

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
frankfliu: We can make this in `ProcessUtils` class, but need to make error checking more generic, this class currently has logic specific for s3cmd:

```
logOutput.startsWith("ERROR ")
```

We can pass a regex if caller need extra error checking:
```
(regex != null && logOutput.matches(regex))
```

siddvenk (approved):

PR #2615, opened 2024-12-02
[serving] Updates onnxruntime to 1.20.0 and add integration tests

xyang16:
## Description ##

### DIFF
--- a/serving/docker/lmi.Dockerfile
+++ b/serving/docker/lmi.Dockerfile
@@ -16,7 +16,7 @@ ARG djl_version
 ARG djl_serving_version
 ARG python_version=3.11
 ARG djl_torch_version=2.5.1
-ARG djl_onnx_version=1.19.0
+ARG djl_onnx_version=1.20.0
 
 EXPOSE 8080
 
@@ -81,6 +81,7 @@ RUN apt-get update && apt-get install -yq libaio-dev libopenmpi-dev g++ unzip cu
 COPY requirements-lmi.txt ./requirements.txt
 RUN pip3 install torch==2.5.1 torchvision==0.20.1 --index-url https://download.pytorch.org/whl/cu124 && pip3 cache purge
 RUN pip3 install -r requirements.txt \
+    && pip3 install onnxruntime-gpu==$djl_onnx_version \
     && git clone https://github.com/neuralmagic/AutoFP8.git \
     && cd AutoFP8 \
     && git reset --hard 4b2092c \
--- a/serving/docker/lmi.Dockerfile
+++ b/serving/docker/lmi.Dockerfile
@@ -18,6 +18,9 @@ ARG python_version=3.11
 ARG djl_torch_version=2.5.1
 ARG djl_onnx_version=1.20.0
 
+# djl converter wheel for text-embedding use case
+ARG djl_converter_wheel="https://publish.djl.ai/djl_converter/djl_converter-0.31.0-py3-none-any.whl"
+
 EXPOSE 8080
 
 COPY dockerd-entrypoint-with-cuda-compat.sh /usr/local/bin/dockerd-entrypoint.sh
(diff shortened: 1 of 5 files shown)


sindhuvahinis (approved):

serving/docker/lmi.Dockerfile
```diff
@@ -81,6 +81,7 @@ RUN apt-get update && apt-get install -yq libaio-dev libopenmpi-dev g++ unzip cu
 COPY requirements-lmi.txt ./requirements.txt
 RUN pip3 install torch==2.5.1 torchvision==0.20.1 --index-url https://download.pytorch.org/whl/cu124 && pip3 cache purge
 RUN pip3 install -r requirements.txt \
+    && pip3 install onnxruntime-gpu==$djl_onnx_version \
```
siddvenk, line 84: Can we add this to the requirements-lmi.txt file?

serving/docker/lmi.Dockerfile
```diff
@@ -16,7 +16,7 @@ ARG djl_version
 ARG djl_serving_version
 ARG python_version=3.11
 ARG djl_torch_version=2.5.1
-ARG djl_onnx_version=1.19.0
+ARG djl_onnx_version=1.20.0
```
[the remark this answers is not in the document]
xyang16: I upgraded onnx to 1.20.0 in DJL.

serving/docker/lmi.Dockerfile
xyang16: To make onnxruntime-gpu to work, it needs to be installed after onnxruntime is installed. I found that if I put it in requirements-lmi.txt, the order is not guaranteed.

serving/docker/lmi.Dockerfile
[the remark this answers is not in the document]
xyang16: Onnxruntime java only has 1.20.0: https://mvnrepository.com/artifact/com.microsoft.onnxruntime/onnxruntime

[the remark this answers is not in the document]
xyang16: Just realized I have 1.20.1 in title, fixed it.

serving/docker/lmi.Dockerfile
siddvenk: Do we need to install both onnxruntime and onnxruntime-gpu? Could we replace onnxruntime in the requirements file with onnxruntime-gpu==1.20.0?

xyang16: We need both onnxruntime and onnxruntime-gpu

xyang16: 1. djl_converter will install onnxruntime as a dependency. I have moved djl_converter out of requirements-lmi.txt with --no-deps.
2. Moved onnxruntime-gpu into requirements-lmi.txt

serving/docker/lmi.Dockerfile
```diff
@@ -16,7 +16,10 @@ ARG djl_version
...
+
+# djl converter wheel for text-embedding use case
+ARG djl_converter_wheel="https://publish.djl.ai/djl_converter/djl_converter-0.31.0-py3-none-any.whl"
```
siddvenk, line 22: once we update this wheel to 0.32.0, will we be able to include it in the requirements file?

siddvenk (approved):

serving/docker/lmi.Dockerfile
xyang16: I can make some changes to include it in the requirements file

siddvenk: Ideally we should be putting all python dependencies in the requirements file so that we can see any dependency conflicts at build time. 

The torch dependencies are outside for two reasons:
- torch by itself is like 5-6gb, so it helps parallelize the image download
- we're using an index url to fetch the cuda 124 version.

Other than that, we should put deps into the requirements file (torch is also included in the requirements file to ensure no version conflicts with other deps)

xyang16: Because requirements file does not support --no-deps, I have to move djl-convert wheel outside of requirements. I can remove the onnxruntime dependency from djl-convert, so we don't need --no-deps for djl-convert.

PR #2616, opened 2024-12-02
[python][fix] check whether config has enable_lora attribute

sindhuvahinis:
## Description ##

Neuronx unit tests cases are failing because of this. `enable_lora` is only available in lmi_dist and vllm as of now. Lora is not supported in neuron yet.

### DIFF
--- a/engines/python/setup/djl_python/input_parser.py
+++ b/engines/python/setup/djl_python/input_parser.py
@@ -198,7 +198,8 @@ def add_server_maintained_params(request_input: RequestInput,
 
 def parse_adapters(request_input: TextInput, input_item: Input,
                    input_map: Dict, **kwargs):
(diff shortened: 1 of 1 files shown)


engines/python/setup/djl_python/input_parser.py
```diff
@@ -198,7 +198,8 @@ def add_server_maintained_params(request_input: RequestInput,
 
 def parse_adapters(request_input: TextInput, input_item: Input,
                    input_map: Dict, **kwargs):
-    if kwargs.get("configs").enable_lora:
+    configs = kwargs.get("configs")
+    if hasattr(configs, "enable_lora") and configs.enable_lora:
```
xyang16, line 202:
```suggestion
    if getattr(configs, "enable_lora", False):
```

PR #2628, opened 2024-12-09
[nightly] Create Nightly Pipeline, make docker-nightly-publish.yml & integration.yml more modular

HappyAmazonian:
## Description ##

This PR includes the following changes
* `docker-nightly-publish.yml` now pushes image to ECR only
* `docker-nightly-publish.yml` add arch arg

* `integration.yml` add workflow_call  for reusing in future
* `integration.yml` add tag-suffix & envs to allow the tests to fetch the image from ECR

* `tests/integration/tests.py` now it uses ECR image to perform the tests
* `docker_publish.yml` syncs images in the temp ECR to the staging ECR repo and dockerhub

* `nightly.yml` introduce a pipeline of build,integtest,publish nightly images

---
* docker-nightly-publish now only pushes to ECR.
nightly.yml will handles build, integ test, and push to the staging ECR/dockerhub.

* you can use `docker-nightly-publish` to build the image and the built image will be in the temp ECR repo.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

I have created one run here 
integ test: https://github.com/deepjavalibrary/djl-serving/actions/runs/12360671952/job/34496566447
nightly build: https://github.com/deepjavalibrary/djl-serving/actions/runs/12401201241/job/34620194153
docker publish: https://github.com/deepjavalibrary/djl-serving/actions/runs/12400140451

nightly pipeline run: https://github.com/deepjavalibrary/djl-serving/actions/runs/12399112958/job/34622823684


.github/workflows/integration.yml
```diff
@@ -224,11 +236,25 @@ jobs:
...
       - name: Download models and dockers
         run: |
-          docker pull deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG
+          if [ ${{ github.event.inputs.djl-version }} == "temp"  ]; then
+            DOCKER_IMAGE_URI="185921645874.dkr.ecr.us-east-1.amazonaws.com/djl-ci-temp:pytorch-inf2-${GITHUB_SHA}"
+          else
+            DOCKER_IMAGE_URI="deepjavalibrary/djl-serving:$DJLSERVING_DOCKER_TAG"
+          fi
+          echo "DOCKER_IMAGE_URI=$DOCKER_IMAGE_URI" >>$GITHUB_ENV
+          aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${{env.AWS_ECR_REPO}}
```
Lokiiiiii, line 255: Auto infer region from `DOCKER_IMAGE_URI`

.github/workflows/docker-nightly-publish.yml
```diff
@@ -70,29 +79,25 @@ jobs:
...
-          export NIGHTLY="-nightly"
-          docker compose build --no-cache \
-          --build-arg djl_version=${{ env.DJL_VERSION }}-SNAPSHOT \
-          --build-arg djl_serving_version=${{ env.SERVING_VERSION }}-SNAPSHOT \
-          ${{ matrix.arch }}
-          repo="185921645874.dkr.ecr.us-east-1.amazonaws.com/djl-ci-temp"
-          aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin $repo
-          tempTag="$repo:${{ matrix.arch }}-${GITHUB_SHA}"
-          docker tag deepjavalibrary/djl-serving:${{ matrix.arch }}-nightly $tempTag
+          aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin ${{env.AWS_ECR_REPO}}
```
Lokiiiiii, line 93: Auto infer the region from the Docker image URI

Lokiiiiii:
Pushing and Pulling the built container in itself takes >10 mins. I would like to avoid this if possible. 

I added a persistent FSX cache volume to all the self hosted runners. Can you try to use the cache to persist data between runners instead of using ECR ? Maybe this can save us some time. I assume we would run into the same issue with PR sanity tests.

Lokiiiiii (changes requested):

.github/workflows/integration.yml
Lokiiiiii: Consider https://github.com/deepjavalibrary/lmi-distro/blob/main/.github/workflows/neo-fml-integration.yml#L109-L112

.github/workflows/docker-nightly-publish.yml
Lokiiiiii: Consider - https://github.com/deepjavalibrary/lmi-distro/blob/main/.github/workflows/neo-fml-integration.yml#L109-L112

.github/workflows/docker-nightly-publish.yml
```diff
@@ -70,29 +79,25 @@ jobs:
...
+          tempTag="${{ env.AWS_ECR_REPO }}:${{ matrix.arch }}-${GITHUB_SHA}"
```
siddvenk, line 94: This will result in new temp images in ECR nightly. Have we configured the retention policy in this ECR repo? We should discuss what that retention period looks like

.github/workflows/docker-nightly-publish.yml
```diff
@@ -111,6 +116,34 @@ jobs:
...
+    with:
+      djl-version: temp
+
+  push-to-dockerhub:
+    runs-on: ubuntu-latest
+    needs: [run-integration-tests]
+    strategy:
+      matrix:
+        arch: [ cpu, cpu-full, pytorch-inf2, pytorch-gpu, tensorrt-llm, lmi ]
+    steps:
```
siddvenk, line 133: It looks like we're a binary success/fail situation here for all containers. Integ tests for all containers must pass for any new nightly publish. We should think through what it takes to make this more granular, which may involve splitting up the current integration.yml file into a few separate worfklows, or adding specific steps here that run tests on a container by container basis

.github/workflows/docker-nightly-publish.yml
HappyAmazonian: Someone created the ECR repo already with 7-day retention policy.

HappyAmazonian:
> Pushing and Pulling the built container to transfer it between hosts in itself takes >10 mins. I would like to avoid this if possible.
> 
> I added a persistent FSX cache volume to all the self hosted runners. Can you try to use the cache to persist data between runners instead of using ECR ? Maybe this can save us some time. I assume we would run into the same issue with PR sanity tests.

I have done some experiment today. And I found it is not possible to use Lustre right now for docker volume storage. By checking https://docs.docker.com/engine/storage/drivers/select-storage-driver/#supported-backing-filesystems, Lustre as a file system is not supported. The only way that we can start the dockerd service with the Lustre folder is to use `vfs`.

according to doc,
```The vfs storage driver is intended for testing purposes, and for situations where no copy-on-write filesystem can be used. Performance of this storage driver is poor, and is not generally recommended for production use.``` 

 And it throws error during the `docker pull` when we use `vfs`.

File: .github/workflows/docker-nightly-publish.yml (hunk 3 above)
Comment (HappyAmazonian) [reply .github/workflows/docker-nightly-publish.yml:94]: The price for ECR storage is $0.09 per GB per month. ~$90 per TB per month.
If we want to make it more frugal. I believe a retention policy of 2days is enough since tests only last a few hours

Comment (Lokiiiiii) [reply .github/workflows/docker-nightly-publish.yml:94]: I am not really worried about the cost here. I am worried about the time wasted in saving and retrieving images from ECR. ECR is a fairly slow service.

**Comment by Lokiiiiii** [comment]
> > Pushing and Pulling the built container to transfer it between hosts in itself takes >10 mins. I would like to avoid this if possible.
> > I added a persistent FSX cache volume to all the self hosted runners. Can you try to use the cache to persist data between runners instead of using ECR ? Maybe this can save us some time. I assume we would run into the same issue with PR sanity tests.
> 
> I have done some experiment today. And I found it is not possible to use Lustre right now for docker volume storage. By checking https://docs.docker.com/engine/storage/drivers/select-storage-driver/#supported-backing-filesystems, Lustre as a file system is not supported. The only way that we can start the dockerd service with the Lustre folder is to use `vfs`.
> 
> according to doc, `The vfs storage driver is intended for testing purposes, and for situations where no copy-on-write filesystem can be used. Performance of this storage driver is poor, and is not generally recommended for production use.`
> 
> And it throws error during the `docker pull` when we use `vfs`.

We don't have to start docker on the FSX volume. We just have to save & load from the FSX volume. Can you try - `docker save` followed by `docker load`. 

Ref:
https://docs.docker.com/reference/cli/docker/image/save/
https://docs.docker.com/reference/cli/docker/image/load/

You don't even need to compress images. Instead of saving & loading to ECR, you will be using a FSX volume.

File: tests/integration/tests.py [hunk 5]
```diff
@@ -10,6 +10,9 @@
 import test_client
 
 djl_version = os.environ.get('TEST_DJL_VERSION', '').strip()
+override_image_tag_suffix = os.environ.get('OVERRIDE_IMAGE_TAG_SUFFIX',
+                                           '').strip()
+image_repo = os.environ.get('IMAGE_REPO', '').strip()
```
siddvenk, line 15: What's the reasoning for two separate env vars here? Would it be simpler to just have a single `OVERRIDE_CONTAINER_URI` env var?

HappyAmazonian: since we have multiple arches, the tags will be arch-{suffix}. Having  `OVERRIDE_CONTAINER_URI` here will require us to generate image uri for each arch.

.github/workflows/docker-nightly-publish.yml
```diff
@@ -7,26 +7,40 @@ on:
...
+  AWS_ECR_REPO: "185921645874.dkr.ecr.us-east-1.amazonaws.com/djl-ci-temp"
+  DOCKER_HUB_REPO: "deepjavalibrary/djl-serving"
+
 jobs:
   nightly-build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        arch: [ cpu, cpu-full, pytorch-inf2, pytorch-gpu, tensorrt-llm, lmi ]
+        arch: ${{ fromJson(inputs.arch) }}
```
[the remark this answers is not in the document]
HappyAmazonian: I added condition statements to make aarch64 as exception
line44
exclude:
          - arch: aarch64
and if: contains(fromJSON(inputs.arch), 'aarch64')

siddvenk (approved):

PR #2639, opened 2024-12-13
enable full passthrough of vllm engine args, with backwards compatibi…

siddvenk:
…lity for existing differences with engine agnostic configs

## Description ##

This PR addresses a common pain point with using vllm. We currently restrict the set of engine args that can be passed based on what we expose through VllmRbProperties. This change does the following:
- enable pass through args via Pydantic extra configs
- remove configs from VllmRbProperties that are exactly the same as the vllm engine args in both name and default value
- validate specific configs that we have different engine agnostic names for

I plan on making a similar change for LMI-Dist. I'm starting with VLLM first to keep the change contained, and gather feedback suggestions which I'll take forward for LMI-Dist.

## Type of change

- [x] This change requires a documentation update

## Checklist:

- [x] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L72); One example would be `pytest tests.py -k "TestCorrectnessLmiDist"  -m "lmi_dist"`

Ran `python -m pytest -k "Vllm" tests.py` locally and they are passing


siddvenk:
putting this back to draft for the time being. I need to rebase and then validate tests again. Will mark ready for review when complete

PR #2660, opened 2025-01-14
[workflow] Use self-host instance for trtllm image sync

HappyAmazonian:
## Description ##
Currently, we use ubuntu-latest to upload images. It gets out of storage since the trtllm image is too big
This PR change is to use self-host instance for trtllm image sync.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

I commented out the uploading code and verified that trtllm sync is using aarch instance
https://github.com/deepjavalibrary/djl-serving/actions/runs/12759641850/job/35563767282


.github/workflows/docker_publish.yml
```diff
@@ -48,9 +48,26 @@ env:
   ECR_REPO_REGION: "us-east-1"
 
 jobs:
+  create-aarch64-runner:
```
siddvenk, line 51: can we use a cpu runner instead?

.github/workflows/docker_publish.yml
```diff
@@ -101,3 +118,13 @@ jobs:
...
+  stop-aarch64-runner:
+    if: always()
+    runs-on: [ self-hosted, scheduler ]
+    needs: [docker-sync, create-aarch64-runner]
+    steps:
+      - name: Stop all instances
+        run: |
+          cd /home/ubuntu/djl_benchmark_script/scripts
+          instance_id=${{ needs.create-aarch64-runner.outputs.aarch64_instance_id }}
+          ./stop_instance.sh $instance_id
```
siddvenk, line 130: newline at the end

.github/workflows/docker_publish.yml
HappyAmazonian: Sure. small question here: is there any specific reason we prefer the cpu one?

The graviton instance `c7g.4xlarge` is less than half of the cost of the CPU instance used here `c6i.8xlarge`.

siddvenk: oh interesting. let's keep graviton then. graviton instance AMI doesn't come installed with docker and some other packages, so there's just some additional overhead with those.

siddvenk (approved):

PR #2667, opened 2025-01-17
Support for NVIDIA Nsight Systems Profiling of DJL Serving

Lokiiiiii:
## Description
This PR introduces support for performance profiling and debugging in DJL Serving using NVIDIA Nsight Systems. The changes enable detailed profiling and performance analysis to optimize the model server's operation.

#### Debug Mode Activation
- Introduced a `DEBUG_MODE` environment variable:
  - When set, the model server will run with performance profiling enabled.
  - Ensures smoother debugging workflows.

#### Nsight Systems Integration
- Automatic installation of NVIDIA Nsight Systems when `DEBUG_MODE` is enabled.
- Nsight Systems version can be specified using the `NSIGHT_VERSION` environment variable.

#### Profiling Configuration
- Added environment variables to control profiling behavior:
  - **`NSYS_PROFILE_DELAY`**: Time delay (in seconds) before starting profiling, to capture relevant operations.
  - **`NSYS_PROFILE_DURATION`**: Duration (in seconds) of profiling. It is recommended to keep this under 10 minutes to avoid large report sizes.
  - **`NSYS_PROFILE_TRACE`**: Specify APIs and libraries to trace during profiling (e.g., `cuda,nvtx,osrt,cudnn`).

#### Report Handling
- Profiling reports are automatically saved and uploaded to an S3 bucket if the `S3_DEBUG_PATH` environment variable is provided.

## Testing
- Verified the installation and usage of NVIDIA Nsight Systems in debug mode.
- Validated environment variable handling and error scenarios.
- **!!Yet to test report generation in multi-node setting!!**

## Documentation
- Created the troubleshooting guide to include details on using the new debugging and profiling features.

### DIFF
--- a/TROUBLESHOOTING.md
+++ b/TROUBLESHOOTING.md
@@ -1,0 +1,61 @@
+# **Troubleshooting Guide**
+
+This guide provides steps and information to troubleshoot issues related to the model server and debugging tools. It is a work in progress and will eventually be moved to `serving/docs` upon finalization.
+
+---
+
+## **Profiling**
+
+> Note that profiling is still being worked upon and the interfaces are bound to change until finalized. In it's current state this is only recommended for personal debugging.
+
+The container can be started in **DEBUG mode** by setting the environment variable `DEBUG_MODE=1`. When enabled, this mode facilitates advanced profiling and debugging capabilities with the following effects:
+
+### **1. Installation of Debugging Tools**
+
+In DEBUG mode, the following tool will be installed automatically:
+
+- **[NVIDIA Nsight Systems](https://docs.nvidia.com/nsight-systems/)**
+  - Nsight Systems enables system-wide performance analysis.
+  - The version of Nsight can be controlled using the environment variable:
+    - `NSIGHT_VERSION`: Specifies the version of Nsight Systems to install (e.g., `2024.6.1`).
+
+### **2. Profiling with Nsight Systems**
+
+The model server will automatically start under the `nsys` profiler when `DEBUG_MODE` is enabled. The following environment variables can be configured to customize the profiling behavior:
+
+- **`NSYS_PROFILE_DELAY`**:  
+  - Specifies the delay in seconds before profiling begins.  
+  - Use this to exclude startup activities and capture only relevant information.  
+  - **Default**: `30` seconds.
+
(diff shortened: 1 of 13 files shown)


siddvenk:
Few questions:
- Is it possible to leverage this profiler on systems without NVIDIA GPUS? It looks like there are different profiles we can capture, so i'm curious if this has utility for Neuron/CPU images. If no, before install maybe we do a `nvidia smi` check or something similar to exit quickly on non gpu systems
- Not necessarily for this PR, but since this does require network access to download the tool at runtime, it won't work in network isolation use-cases. We may want to consider extending the current containers with a new debug build stage that installs this tool and can be used directly for JS. Thoughts?
- I would prefer a different flag name instead of `DEBUG_MODE`. `LMI_DEBUG_MODE`, `LMI_ENABLE_PROFILER` or something similar is clearer to me. I'm not sure if DEBUG_MODE would conflict with a flag in any of the deps we pull in

Lokiiiiii:
>     * Is it possible to leverage this profiler on systems without NVIDIA GPUS? It looks like there are different profiles we can capture, so i'm curious if this has utility for Neuron/CPU images. If no, before install maybe we do a `nvidia smi` check or something similar to exit quickly on non gpu systems

I am no expert but the tool can be used for CPU profiling from what I have seen so far. So I don't think the test is necessary.

>     * Not necessarily for this PR, but since this does require network access to download the tool at runtime, it won't work in network isolation use-cases. We may want to consider extending the current containers with a new debug build stage that installs this tool and can be used directly for JS. Thoughts?

Good idea but requires a much deeper discussion - 
1. How do we keep bloat minimal and containers cache-able ?
2. How do we get Hosting to copy debug artifacts ?

I am currently introducing this for developers' sake. If it proves to be truly useful, we can think about making this a self-serviceable customer feature. 

>     * I would prefer a different flag name instead of `DEBUG_MODE`. `LMI_DEBUG_MODE`, `LMI_ENABLE_PROFILER` or something similar is clearer to me. I'm not sure if DEBUG_MODE would conflict with a flag in any of the deps we pull in

Ack. I will fix this in the next itr.

siddvenk (approved):

PR #2676, opened 2025-01-23
[lmi][vllm] do not require do_sample to enable sampling

siddvenk:
## Description ##

This change removes the need for `do_sample` to be set explicitly for sampling to occur. Previously, sampling parameters would be ignored unless do_sample was set to true. While that behavior did follow the transformers behavior, it is no longer what users expect.

We will still respect do_sample: false as a greedy sampling indicator.

I also removed the .keys() from this method as it's unnecessary. 

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
+++ b/engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
@@ -154,28 +154,27 @@ class LmiDistRollingBatch(RollingBatch):
         :return: The same parameters dict, but with lmi-dist style parameter names.
         """
         parameters["max_tokens"] = parameters.pop("max_new_tokens", 30)
-        # If `do_sample` is not provided, force temperature=0.0, i.e. greedy
-        # else set to user-provided value or default to 1.0
-        if not parameters.pop('do_sample', False):
-            parameters['temperature'] = 0.0
-        else:
-            parameters['temperature'] = parameters.get('temperature', 1.0)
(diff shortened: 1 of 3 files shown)


engines/python/setup/djl_python/rolling_batch/lmi_dist_rolling_batch.py
```diff
@@ -154,28 +154,25 @@ def translate_lmi_dist_params(self, parameters: dict):
...
-        # If `do_sample` is not provided, force temperature=0.0, i.e. greedy
-        # else set to user-provided value or default to 1.0
-        if not parameters.pop('do_sample', False):
-            parameters['temperature'] = 0.0
-        else:
-            parameters['temperature'] = parameters.get('temperature', 1.0)
-        if "seed" in parameters.keys():
+        # if do_sample is provided and temperature is not, default to 1.0
+        if parameters.pop('do_sample', False):
+            parameters['temperature'] = parameters.get("temperature", 1.0)
```
sindhuvahinis, line 159: Shouldn't we default to 0.0 for greedy?

siddvenk: oh yes, good catch. If do_sample == false, then temp = 0. will update

siddvenk: updated

sindhuvahinis (approved):

PR #2690, opened 2025-01-28
[docker] separate vllm and lmi-dist modes into separate virtual envir…

siddvenk:
…onments

## Description ##

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B


wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java
```diff
@@ -206,4 +207,20 @@ private static boolean isTextGenerationModel(LmiUtils.HuggingFaceModelConfig mod
                 OPTIMIZED_TASK_ARCHITECTURES);
         return false;
     }
+
+    private static void setPythonExecutable(Properties lmiProperties) {
+        if (lmiProperties.containsKey("option.pythonExecutable")) {
+            return;
+        }
+        String rollingBatch = lmiProperties.getProperty("option.rolling_batch");
+        if ("vllm".equals(rollingBatch)) {
```
xyang16, line 216: How about option.rolling_batch=auto?

siddvenk: we translate rolling_batch=auto earlier in this process in setRollingBatch - we shouldn't have auto as a value here

serving/docker/scripts/create_virtual_env.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+# used in the dockerfiles to create virtualenvs per engine
+# currently only intended for use in lmi.Dockerfile, need to refactor this to work for trtllm/neuron if needed
+base_directory=$1
+requirements_file=$2
+
+virtual_env_path="${base_directory}/venv"
```
davidthomas426, line 7: Any particular reason for the "/venv" suffix? This makes the name of all virtual envs created with this script "venv". I noticed when activating it in a shell, since it shows the name in parens in the prompt and it was "(venv) ...".

siddvenk: yea i don't think this is actually needed - can remove venv.

I also don't need to source/deactivate - can just call the pip from <venv>/bin/pip

xyang16 (approved):

PR #2692, opened 2025-01-28
[python] remove deepspeed related AOT code

sindhuvahinis:
## Description ##

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/serving/docker/partition/utils.py
+++ b/serving/docker/partition/utils.py
@@ -42,16 +42,6 @@ def get_partition_cmd(is_mpi_mode, properties):
         ]
 
 
-def get_engine_configs(properties):
-    engine = properties.get('engine')
-    configs = {}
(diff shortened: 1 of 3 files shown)


serving/docker/partition/utils.py
```diff
@@ -72,7 +62,7 @@ def get_djl_version_from_lib():
 
 
 def is_engine_mpi_mode(engine):
```
siddvenk, line 64: do we use this function anywhere currently?

siddvenk (approved):

serving/docker/partition/utils.py
sindhuvahinis: No, not right now. But if we unify TRTLLM which is in a separate script to this partition script, then we need this.

PR #2701, opened 2025-02-01
AutoFP8 to llmcompressor migration for FP8 quantization

a-ys:
## Description ##

Migrates existing FP8 quantization functionality to llmcompressor from AutoFP8. The existing quantization recipe is fp8 weights and activation quantization, with static activation scales. Uses cnn-dailymail for calibration, defaulting to 512 samples at 2048 seq len.

Installed the previous version of llm-compressor (0.3.1) due to an issue with compatibility between llm-compressor (0.4.0) and transformers 2.5.2, which is the current version in djl-serving.

## TODOs

- Update CI tests in lmi-distro to verify these changes.
- Remove AutoFP8 dependency from container.
- Future feature support (prioritization tbd)
    - MoE model quantization. (May run with exiting code, but will not ignore the correct layers)
    - KV cache quantization with static scales.
    - Calibration dataset selection
    - AWQ migration (not yet supported in llmcompressor)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Checklist:
- [ ] Please add the link of [**Integration Tests Executor** run](https://github.com/deepjavalibrary/djl-serving/actions/workflows/integration_execute.yml) with related tests.
- [x] Have you [manually built the docker image](https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docker/README.md#build-docker-image) and verify the change?
- [ ] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L72); One example would be `pytest tests.py -k "TestCorrectnessLmiDist"  -m "lmi_dist"`
- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [x] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

## Feature/Issue validation/testing

- Tested quantization of tinyllama with llmcompressor and serving with v14 v2 preview container through Neo workflow.

PR #2703, opened 2025-02-03
fix lmi/vllm virtual envs, update to vllm 0.7.1

siddvenk:
## Description ##

This changes updates to vllm 0.7.1, which involves shuffling around some dependencies and being less strict with dependency versions. 

Additionally, it updates the chat processing for vllm to be functional. There is still a good amount we need to implement for chat processing, that i'll take up in a follow up PR:
- use the sampling params provided directly from the vllm chat object (to_sampling_params method). this will ensure we use the correct sampling params for chat
- validate function calling and tool usage with this update
- allow user to specify override chat template, and chat format

I have tested this with (single test for each):
- HF non rolling batch
- HF scheduler rolling batch
- vLLM rolling batch
- lmi-dist rolling batch

I also added a chat test for mistral with vllm.

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/chat_completions/vllm_chat_utils.py
+++ b/engines/python/setup/djl_python/chat_completions/vllm_chat_utils.py
@@ -21,16 +21,11 @@ from vllm.entrypoints.chat_utils import (apply_hf_chat_template,
                                          resolve_chat_template_content_format)
 
 
-def is_chat_completions_request(inputs: Dict) -> bool:
-    return "messages" in inputs
-
-
 def parse_chat_completions_request_vllm(
     input_map: Dict,
     is_rolling_batch: bool,
     rolling_batch,
     tokenizer,
-    chat_template: Optional[str] = None,
     configs: Properties = None,
     is_mistral_tokenizer: bool = False,
 ):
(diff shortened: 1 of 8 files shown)


engines/python/setup/djl_python/chat_completions/vllm_chat_utils.py
```diff
@@ -21,16 +21,11 @@
                                          resolve_chat_template_content_format)
 
 
-def is_chat_completions_request(inputs: Dict) -> bool:
```
siddvenk, line 24: deleted because it's not used

engines/python/setup/djl_python/chat_completions/vllm_chat_utils.py
```diff
@@ -41,12 +36,6 @@ def parse_chat_completions_request_vllm(
             "You must enable rolling batch to use the chat completions format."
         )
 
-    if not is_mistral_tokenizer and not hasattr(tokenizer,
```
siddvenk, line 44: deleted because the vllm utils do this validation for us already

serving/docker/scripts/create_virtual_env.sh
```diff
@@ -7,12 +7,6 @@ requirements_file=$2
...
 venv_pip="${venv_directory}/bin/pip"
-$venv_pip install -r $requirements_file
+$venv_pip install -r $requirements_file || exit 1
 $venv_pip install https://publish.djl.ai/djl_converter/djl_converter-0.31.0-py3-none-any.whl --no-deps
-git clone https://github.com/neuralmagic/AutoFP8.git
-cd AutoFP8
-git reset --hard 4b2092c
-$venv_pip install .
-cd ..
-rm -rf AutoFP8
```
sindhuvahinis, line 17: Do we not need FP8 installation?

siddvenk: not anymore! we're using llm compressor now https://github.com/deepjavalibrary/djl-serving/pull/2701

sindhuvahinis (approved):

PR #2717, opened 2025-02-05
Support Sharding for multinode case

HappyAmazonian:
## Description ##

Support Sharding with PP_Degree > 1
* make tp_degree and pp_degree member variables to `NeoShardingService` class 
* remove pipeline_parallel_degree validation for sharding
* change `save_configs` so it save info to model config obj gradually. It will only output the file to disk when it gets the last pp_rank.
* add  `pp_rank_to_shard ` param to `shard_lmi_dist_model`
  * when pp_degree = 1, the behavior should be the same
* in `run_sharding`, we create a loop of pp_degree, so it calls shard_lmi_dist_model for each pp_rank.

this will break  if the lmi_dist is not  updated

## Type of change

## Checklist:

Tested with 

llama3 405b pp=2 tp=8, sharded on one instance, and tested serving with two p4de instances.

llama3 8b pp=2 tp=4, sharded on one instance, and tested serving with two g6.12xlarge instances.

llama3 8b pp=1 tp=4, sharded on one instance, and tested serving with one g6.12xlarge instance.

llama3 8b pp=1 tp=8, sharded on one instance, and tested serving with two g6.12xlarge instances.


serving/docker/partition/sm_neo_shard.py
```diff
@@ -100,9 +112,13 @@ def generate_properties_file(self):
...
 
+    # By setting pp_rank, only workers in that pp_rank will load the model
+    # i.e. in case pp=2, tp=4, the arg of pp_rank=1 will
+    #   call load model for those workers with local_rank [4,7]
+    #   And those workers with local_rank [0,3] will have empty layers
     def shard_lmi_dist_model(self, input_dir: str, output_dir: str,
-                             pp_degree: int, tp_degree: int,
-                             chunk_mb: int) -> None:
+                             pp_degree: int, tp_degree: int, chunk_mb: int,
+                             pp_rank_to_shard: int) -> None:
```
ethnzhng, line 121: Maybe let's rename `pp_rank_to_shard` to `target_pp_rank` for clarity

serving/docker/partition/sm_neo_shard.py
```diff
@@ -46,30 +48,40 @@ def __init__(self):
...
-            f"SageMaker Fast Model Loader config file saved to {output_dir}")
-        self.copy_non_safetensors_files(input_dir, output_dir)
-        logging.info(f"Other non-Safetensors files copied to {output_dir}")
+        if pp_rank == self.pp_degree - 1:
+            self.shard_config.save(output_dir=output_dir)
+            logging.info(
+                f"SageMaker Fast Model Loader config file saved to {output_dir}"
+            )
+            self.copy_non_safetensors_files(input_dir, output_dir)
+            logging.info(f"Other non-Safetensors files copied to {output_dir}")
```
ethnzhng, line 84: What if we reorganize the logic here to make the intention more clear, maybe something along the lines of

1. instead of passing in pp_rank here, we pass a boolean like `should_save_final_config`, and move the check for setting this value to the caller
2. instead of checking for misaligned pp_rank, we check for an empty config object to determine if we should skip adding a shard to the ModelConfig

serving/docker/partition/sm_neo_shard.py
```diff
@@ -158,42 +174,46 @@ def shard_lmi_dist_model(self, input_dir: str, output_dir: str,
...
-            pp_degree = int(
-                self.properties.get("option.pipeline_parallel_degree", 1))
-            self.shard_lmi_dist_model(
-                input_dir=self.INPUT_MODEL_DIRECTORY,
-                output_dir=self.OUTPUT_MODEL_DIRECTORY,
-                pp_degree=pp_degree,
-                tp_degree=int(
-                    self.properties["option.tensor_parallel_degree"]),
-                chunk_mb=CHUNK_MB)
+            for i in range(self.pp_degree):
```
ethnzhng, line 209: These changes only enable sharding for multi-node when the nodes are split by pipeline parallelism, e.g. tp=8, pp=2; we need to also account for the tensor parallel-only multi-node case, e.g. tp=16, which is rarer but could be required or more performant in some scenarios. 

So we need to check if the tensor parallel degree is greater than what fits on a single node, and if so then partition the ranks accordingly and shard each node’s world sequentially, as we are doing here for pipeline

Lokiiiiii (approved):

serving/docker/partition/sm_neo_shard.py
```diff
@@ -46,26 +46,35 @@ def __init__(self):
...
+            if not entry["config"]:
+                continue
+            self.shard_config.add_shard(
                 pipeline_parallel_degree=int(entry["pp"]),
                 tensor_parallel_degree=int(entry["tp"]),
                 shard_config=entry["config"],
             )
 
-        conf.save(output_dir=output_dir)
+    def save_configs(self, input_dir: str = "", output_dir: str = "") -> None:
```
ethnzhng, line 76: Nit: remove default dir values, since it should be a fatal error if not passed

ethnzhng (approved):

PR #2726, opened 2025-02-06
[fix][lmi] fix issue where streaming response that finishes with empt…

siddvenk:
## Description ##

There was a bug in the RollingBatch.java implementation that was recently triggered by one of the integration tests. The issue was regarding how we decide to send a new chunk of data back to the client. For non-streaming cases, the engine will report back with empty tokens until the request is finished. It also reports the HTTP code on this final chunk, which allows us to provide the correct UX to end-users.

But with streaming, it's possible that the final token provided by the engine is actually empty/null. If that's the case, then the request hangs within the model server until it times out.

This PR updates the logic for determining when to send a new chunk of information back to the client. 

I ran the failing test tests.py::TestVllm1::test_llama3_1_8b_instruct_tool locally and this fixes the issue. I also validated a couple of other test cases locally to ensure no regression.

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
+++ b/engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
@@ -342,7 +342,7 @@ class RollingBatch implements Runnable {
                         break;
                 }
             }
-            if ((nextToken == null || nextToken.isEmpty()) && code == null) {
+            if ((nextToken == null || nextToken.isEmpty()) && !last) {
                 // in non-streaming cases, we do not return content until generation is finished
                 return;
             }


sindhuvahinis (approved):

engines/python/src/main/java/ai/djl/python/engine/RollingBatch.java
```diff
@@ -342,7 +342,7 @@ void addResponse(byte[] json, Map<String, String> properties) {
                         break;
                 }
             }
-            if ((nextToken == null || nextToken.isEmpty()) && code == null) {
+            if ((nextToken == null || nextToken.isEmpty()) && !last) {
```
xyang16, line 345: For tool calling, the last token might be empty as well.

siddvenk: the test you put in caught this issue :)

xyang16 (approved):

PR #2758, opened 2025-03-11
implementation of python async engine, which decouples receiving requ…

siddvenk:
…ests and sending responses in the python engine

## Description ##

This PR implements an async/decoupled Python engine. This will allow for asynchronous code in the python handler, and enable integrations with LLM engines at a higher level than we do today. It's not exclusive to LLM integration, but that is the main purpose.

The goal of this implementation is to more cleanly integrate with LLM inference engines that have largely adopted the async/await paradigm in their public APIs. It will ideally replace the RollingBatch implementation we have today. Initial tests with vLLM 0.7.1 show a 5-7% performance improvement, all of which can be attributed to our current implementation.

When the frontend receives a request, it sends it to the Python backend and does not expect a response. When the Python backend completes a response, it sends it back to the frontend. The frontend maintains a mapping that allow it to associate the response from python with the client request so that it can send the response to the appropriate connection.

This has been tested via a unit test, and sample vLLM handler.

This is the initial PR, and improvements will be tackled in subsequent PRs. At the moment, this supports:
- concurrent streaming and non-streaming requests.

The follow up PRs will address:
- Solid error handling
- Refactoring to support streaming responses on the Python side in a cleaner API
- Improved vLLM handler to support
  - context manager for vLLM engine to support graceful termination on engine failures
  - LMI output schema support
  - Refactor some utilities to be shared with other engines

## Type of change

## Checklist:
- [x] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L72); One example would be `pytest tests.py -k "TestCorrectnessLmiDist"  -m "lmi_dist"`

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/python_async_engine.py
+++ b/engines/python/setup/djl_python/python_async_engine.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python
+#
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+import asyncio
+import logging
+import time
+import traceback
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from threading import Thread
+from queue import Queue
+from asyncio.queues import Queue as AsyncQueue
+
+from djl_python.inputs import Input
+from djl_python.outputs import Output
+from djl_python.python_sync_engine import PythonSyncEngine
+
+REQUEST_TRACKING_ID_KEY = "request_tracking_id"
+
+
+class PythonAsyncEngine(PythonSyncEngine):
+    """
+    Backend engine to run python code in decoupled/async mode.
+    Requests are forwarded from the model server and submitted to the handler.
+    The handler returns responses as they become available, and sends them to the frontend.
+    Requests are tracked/coordinated via the request_tracking_id property.
+    This is an internal property set/managed by the model server.
+    """
+
+    def __init__(self, args, service):
+        super().__init__(args, service)
+        self.output_queue = AsyncQueue()
+        self.exception_queue = Queue()
+        self.loop = None
+        # Todo: for async mode we should maybe consider
+
+    def receive_requests(self):
+        logging.info("starting receive requests thread")
+        while True:
+            inputs, function_name = self._prepare_inputs()
+            logging.debug(
+                f"received new request with tracking_id {inputs.get_property(REQUEST_TRACKING_ID_KEY)}, submitting to handler"
+            )
+            asyncio.run_coroutine_threadsafe(
+                self.invoke_handler(function_name, inputs), self.loop)
+
+    async def invoke_handler(self, function_name: str, inputs: Input):
+        request_tracking_id = inputs.get_property(REQUEST_TRACKING_ID_KEY)
+        try:
+            outputs = await self.service.invoke_handler_async(
+                function_name, inputs, self.cl_socket)
+        except Exception as e:
+            logging.exception("Failed invoke service.invoke_handler_async()")
+            if (type(e).__name__ == "OutOfMemoryError"
+                    or type(e).__name__ == "MemoryError"
+                    or "No available memory for the cache blocks" in str(e)
+                    or "CUDA error: out of memory" in str(e)):
+                logging.exception(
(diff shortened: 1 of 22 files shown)


engines/python/src/main/java/ai/djl/python/engine/Connection.java
```diff
@@ -160,7 +161,7 @@ static String[] getPythonStartCmd(
                 }
                 sb.append(host).append(':').append(localSize);
             }
-            String[] args = new String[50];
+            String[] args = new String[51];
             args[0] = "mpirun";
             args[1] = "-np";
             args[2] = String.valueOf(worldSize);
```
HappyAmazonian, line 167: I wonder if we can also make this a list instead of hard coding the indices from 0 to 51..

siddvenk: yes, I think we should take this up in a follow up PR. I'm trying to keep refactoring in this PR to a minimum to avoid non-essential changes from distracting.

engines/python/setup/djl_python/python_async_engine.py
```diff
@@ -0,0 +1,143 @@
...
+                while True:
+                    if not all(t.is_alive() for t in threads):
+                        return
+                    time.sleep(1)
+
+            with ThreadPoolExecutor(1) as executor:
+                await asyncio.get_event_loop().run_in_executor(
+                    executor, check_threads)
+
+        asyncio.get_event_loop().run_until_complete(main())
```
ethnzhng, line 138: I think we can just do `asyncio.run()` since there is no existing event loop at this point

engines/python/setup/djl_python/python_async_engine.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    def __init__(self, args, service):
+        super().__init__(args, service)
+        self.output_queue = AsyncQueue()
+        self.exception_queue = Queue()
+        self.loop = None
+        # Todo: for async mode we should maybe consider
+
+    def receive_requests(self):
```
ethnzhng, line 47: I'm wondering if we should add a graceful shutdown mechanism to the `receive_requests` and `send_responses` threads, right now it is unclear how these threads would behave if the main proc is abruptly shut down

siddvenk: yes, definitely. We need graceful shutdown in a few places:
1. Python Engine - I think if the python process crashes or is terminated, the on the front-end we should receive that signal and send a response for all incomplete outstanding requests. I'd probably put this logic in AsyncRequestManager, similar to how we have something for RollingBatch called shutdown
2. The vllm_handler - in local testing, i've triggered a few unrecoverable engine errors that at least trigger a python restart, but don't clean up the memory/resources used by vllm. Would be great too if we have some sort of internal health check. 

I think I can tackle 1 in this PR, but 2 may need some more thought.

engines/python/setup/djl_python/python_async_engine.py
siddvenk: good call i'll update

engines/python/setup/djl_python/vllm_async_service.py
```diff
@@ -0,0 +1,194 @@
...
+        if request.stream:
+            return await handle_streaming_response(response, properties,
+                                                   cl_socket)
+
+        output = Output()
+        response_dict = {"data": response.model_dump_json(), "last": True}
+        if isinstance(response, ErrorResponse):
+            response_dict["code"] = response.code
+            response_dict["error"] = response.message
+        for k, v in inputs.get_properties().items():
```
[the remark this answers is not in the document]
siddvenk: nice catch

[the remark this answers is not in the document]
siddvenk: fixed

engines/python/setup/djl_python/python_async_engine.py
siddvenk: updated

engines/python/setup/djl_python/python_async_engine.py
siddvenk: I need to do a bit more here, but for now what I've done is added a terminateInFlightRequests method that gets invoked when the python engine crashes. If there are requests being handled by the engine, and the engine crashes, then this ensures we send a response to the client.

The next biggest area I want to figure out is, how do we know when the python engine crashes? Right now we're only handling cases where the actual Connection dies, but if vLLM starts to hang, we don't have a timeout mechanism now.

I probably want to figure that out in a follow on PR, since the basic use-case here is working and the PR is already quite large. This feature is also currently guarded by a feature flag that is off by default. Thoughts @ethnzhng ?

ethnzhng: Sounds good. Some potential ideas for when we get to it:
- we can have a health state for the service, and base this off a combo of:
  - timeouts for individual handler invocations
  - how the overall queue is moving
 - then we can have a graceful shutdown / cleanup path for the vLLM service, which we could also maybe use in the fatal error case

ethnzhng (approved):

PR #2785, opened 2025-03-31
[python][async] adding async tensorrt-llm handler

siddvenk:
## Description ##

This PR updates TRTLLM to 0.20.0rc0 by leveraging the LLM API and async mode. It handles both AOT/compilation/quantization and inference implementations.

For AOT, we implement
- Build Config parsing
- Quant Config parsing
- Calib Config parsing

This allows full model conversion + quantization in a single step. Note that PyTorch backend of TRTLLM does not seem to support saving a compiled model. If the pytorch backend is specified, we then skip AOT and let the entrypoint handle the conversion/pytorch setup.

For inference, the default handler expects the model to be compiled ahead of time and does not expose Build/Quant/Config settings since those are expected to be handled by the AOT script, or by the user if bringing a compiled TRT model.

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/lmi_trtllm/trtllm_async_service.py
+++ b/engines/python/setup/djl_python/lmi_trtllm/trtllm_async_service.py
@@ -1,0 +1,373 @@
+#!/usr/bin/env python
+#
+# Copyright 2025 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file
+# except in compliance with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
+# BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
+# the specific language governing permissions and limitations under the License.
+
+# Heavily inspired by https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/serve/openai_server.py
+import asyncio
+import logging
+import signal
+import types
+from http import HTTPStatus
+from typing import AsyncGenerator, AsyncIterator, List, Tuple, TypedDict, Union, Optional
+
+from openai.types.chat import ChatCompletionMessageParam
+
+# yapf: disable
+from tensorrt_llm.executor import CppExecutorError
+from tensorrt_llm.executor.postproc_worker import PostprocParams
+from tensorrt_llm.llmapi import LLM
+from tensorrt_llm.llmapi.llm import RequestOutput
+from tensorrt_llm.serve.openai_protocol import (ChatCompletionRequest,
+                                                ChatCompletionResponse,
+                                                CompletionRequest,
+                                                CompletionResponse,
+                                                CompletionResponseChoice,
+                                                ErrorResponse, ModelCard,
+                                                ModelList, UsageInfo,
+                                                to_llm_disaggregated_params)
+from tensorrt_llm.serve.postprocess_handlers import (
+    ChatPostprocArgs, CompletionPostprocArgs, chat_response_post_processor,
+    chat_stream_post_processor, completion_response_post_processor,
+    completion_stream_post_processor)
+
+from djl_python.async_utils import create_non_stream_output, handle_streaming_response, ProcessedRequest
+from djl_python.inputs import Input
+from djl_python.outputs import Output
+from djl_python.properties_manager.trt_properties import TensorRtLlmProperties
+from djl_python.encode_decode import decode
+
+from .request_response_utils import trtllm_non_stream_output_formatter, trtllm_stream_output_formatter
+
+
+logger = logging.getLogger(__name__)
+
+
+class ConversationMessage(TypedDict):
+    role: str
+    content: str
+
+
+def parse_chat_message_content(
+        message: ChatCompletionMessageParam, ) -> List[ConversationMessage]:
+    role = message["role"]
+    content = message.get("content")
+
+    if content is None:
+        return []
+    if isinstance(content, str):
+        return [ConversationMessage(role=role, content=content)]
+
+    # for Iterable[ChatCompletionContentPartTextParam]
+    texts: List[str] = []
+    for part in content:
+        part_type = part["type"]
+        if part_type == "text":
+            text = part["text"]
+            texts.append(text)
+        else:
+            raise NotImplementedError(f"{part_type} is not supported")
+
+    text_prompt = "\n".join(texts)
+    return [ConversationMessage(role=role, content=text_prompt)]
+
+
+class TensorRTLlmAsyncService:
+
+    def __init__(self):
+        self.trt_configs = None
+        self.llm = None
+        self.tokenizer = None
+        self.model_name = None
+        self.postproc_worker_enabled = False
+        self.initialized = False
+
+    def initialize(self, properties: dict):
+        self.trt_configs = TensorRtLlmProperties(**properties)
+        # In this handler, we expect the front-end to pre-compile the model into a TRTLLM engine.
+        # That means we do not expose build config customization here, and only set runtime configs
+        self.llm = LLM(
+            self.trt_configs.model_id_or_path,
+            **self.trt_configs.get_llm_kwargs(),
+        )
+        self.tokenizer = self.llm.tokenizer
+        self.model_name = self.trt_configs.model_id_or_path
+        self.postproc_worker_enabled = self.llm.args.num_postprocess_workers > 0
+        self.initialized = True
+
+
+    def preprocess_requests(self, inputs: Input) -> ProcessedRequest:
+        batch = inputs.get_batches()
+        assert len(batch) == 1, "only one request per batch allowed"
+        raw_request = batch[0]
+        content_type = raw_request.get_property("Content-Type")
+        decoded_payload = decode(raw_request, content_type)
+        if "model" not in decoded_payload:
+            decoded_payload["model"] = self.model_name
+        if "prompt" in decoded_payload:
+            request = CompletionRequest(**decoded_payload)
+            invoke_call = self.openai_completion
+        elif "messages" in decoded_payload:
+            request = ChatCompletionRequest(**decoded_payload)
+            invoke_call = self.openai_chat
+        else:
+            raise RuntimeError("invalid payload. must contain prompt, inputs, or messages")
+        processed_request = ProcessedRequest(
+            request,
+            invoke_call,
+            trtllm_non_stream_output_formatter,
+            trtllm_stream_output_formatter,
+            False,
+            False,
+        )
+        return processed_request
+
+    async def inference(self, inputs: Input) -> Union[Output, AsyncGenerator[Output, None]]:
+        try:
+            processed_request = self.preprocess_requests(inputs)
+        except Exception as e:
+            logger.exception("Input parsing failed")
+            output = create_non_stream_output("", error=f"Input parsing failed: {e}", code=424)
+            return output
+
+        response = await processed_request.inference_invoker(processed_request.request)
+        if isinstance(response, types.AsyncGeneratorType):
+            return handle_streaming_response(
+                response,
+                processed_request.stream_output_formatter,
+                request=processed_request.request,
+            )
+
+        return processed_request.non_stream_output_formatter(
+            response,
+            request=processed_request.request,
+        )
+
+    @staticmethod
+    def create_error_response(
+            message: str,
+            err_type: str = "BadRequestError",
+            status_code: HTTPStatus = HTTPStatus.BAD_REQUEST) -> ErrorResponse:
+        error_response = ErrorResponse(message=message,
+                                       type=err_type,
+                                       code=status_code.value)
+        return error_response
+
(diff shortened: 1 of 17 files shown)


serving/docker/partition/trt_llm_partition.py
```diff
@@ -15,18 +15,35 @@
...
-    create_model_repo(model_id_or_path, **kwargs)
+
+    # https://nvidia.github.io/TensorRT-LLM/llm-api/reference.html
+    build_config = BuildConfig(
+        max_batch_size=int(kwargs.get("max_rolling_batch_size", 256)),
+        max_input_len=int(kwargs.get("max_input_len", 1024)),
+        max_num_tokens=int(kwargs.get("max_num_tokens", 8192)),
+        max_seq_len=int(kwargs.get("max_seq_len", 256)),
+    )
+    llm = LLM(
```
frankfliu, line 34: I think we should support all configurations that `trtllm-build` support: https://nvidia.github.io/TensorRT-LLM/commands/trtllm-build.html

The easiest way is to copy `build.py.main()`: https://github.com/NVIDIA/TensorRT-LLM/blob/main/tensorrt_llm/commands/build.py#L460

frankfliu: The `build()` sets extra default configurations based on model architecture which `LLM` API didn't provide. We might want to port those logic into `trtllm_async_service.py` when before calling LLM api.

siddvenk: Agreed on supporting all the configs that trtllm-build supports. This is just minimal POC for the draft stage. 

I think we can just import build.py.main instead of copying. Regardless, I agree on this entrypoint for TRTLLM as it exposes both args list for BuildConfig, and from_json_file for more customization.

siddvenk: hoping to get this updated by Wednesday - will ping you on this thread then @frankfliu

siddvenk: The build.py from trtllm looks like a good approach here. The one issue I see so far is that prior to calling trtllm-build (either through cli or the script here) we need to convert the HF model into a trtllm checkpoint. I don't see a great way of doing that at the moment. Each model architecture maps to its own convert script, and it doesn't seem as simple as getting the model arch from the config.json and mapping to the appropriate script.

I think leveraging the BuildConfig + defaults from the build command script + leveraging that in the LLM API might work? LLM API handles the conversion from HF -> TRTLLM checkpoint. I'll give that a shot

serving/docker/trtllm-container-requirements.txt
```diff
@@ -0,0 +1,2 @@
+tensorrt_llm==0.17.0.post1
```
frankfliu, line 1: remove `tensorrt_llm`
1. this is duplicate with the one in dockerfile
2. this requires `--extra-index-url https://pypi.nvidia.com`

engines/python/setup/djl_python/lmi_trtllm/trtllm_async_service.py
```diff
@@ -0,0 +1,500 @@
...
+        self.llm = LLM(
+            self.hf_configs.model_id_or_path,
+            tokenizer=self.hf_configs.model_id_or_path,
+            tensor_parallel_size=self.hf_configs.tensor_parallel_degree,
+            trust_remote_code=self.hf_configs.trust_remote_code,
+            build_config=BuildConfig(
+                max_batch_size=self.hf_configs.max_rolling_batch_size,
+                output_timing_cache="/tmp/model.cache",
+            ),
+            **self.hf_configs.get_extra_kwargs(),
```
frankfliu, line 99: 1. We need to convert datatype, properties are all str value, need to convert to int and bool, we might need to declare all those fields
2. `get_extra_kwargs()` is not a `dict` this doesn't work.

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -9,31 +9,12 @@
 # or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS"
 # BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, express or implied. See the License for
 # the specific language governing permissions and limitations under the License.
-ARG version=12.5.1-devel-ubuntu24.04
+ARG version=12.8.1-cudnn-runtime-ubuntu24.04
 FROM nvidia/cuda:$version
-ARG cuda_version=cu125
+ARG cuda_version=cu128
 ARG python_version=3.10
```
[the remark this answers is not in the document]
siddvenk: I've had success running trtllm 0.19 with python3.10, but updating to 3.12 is preferred.

Created separate PR here for py312 support in docker https://github.com/deepjavalibrary/djl-serving/pull/2818/files.

Will update here once that one is merged.

engines/python/setup/djl_python/lmi_trtllm/trtllm_async_service.py
siddvenk: yea, that's a good point. I was hoping we can avoid the declaration of all the fields here, but seems like that's the only way.

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
```diff
@@ -27,28 +27,6 @@
 from djl_python.async_utils import create_non_stream_output, create_stream_chunk_output
 
 
-class ProcessedRequest:
```
frankfliu, line 30: Need to update `recommendedEntryPoint` in `PyModel` for both vllm and trtllm for async mode

serving/docker/partition/trt_llm_partition.py
```diff
@@ -1,39 +1,190 @@
...
+                },
+                'dry_run': args.dry_run,
+                'visualize_network': args.visualize_network,
+                'max_encoder_input_len': args.max_encoder_input_len,
+                'weight_streaming': args.weight_streaming,
+                'monitor_memory': args.monitor_memory,
+            },
+            plugin_config=plugin_config)
 
-def main():
```
frankfliu, line 32: Better to keep `main()`, variables defined in `if __name__ == "__main__":` considered as global variables, they might conflict with parameters passed into `build_engine()`, PyCharm will give warnings.

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -67,39 +49,20 @@ RUN mkdir -p /opt/djl/conf && \
...
-    pip3 cache purge
-
-# Install TensorRT and TRT-LLM Deps
-RUN pip install --no-cache-dir --extra-index-url https://pypi.nvidia.com tensorrt==${tensorrtlibs_version} janus==${janus_version} nvidia-modelopt==${modelopt_version} && \
-    pip install --no-deps ${trtllm_wheel} && \
-    pyver=$(echo $python_version | awk -F. '{print $1$2}') && \
-    pip3 cache purge
+# We install TRTLLM separately because it is hosted in a different pypi index
+COPY trtllm-container-requirements.txt requirements.txt
+RUN pip install tensorrt_llm==${trtllm_version} --extra-index-url https://pypi.nvidia.com && \
```
frankfliu, line 61: `trtllm_version` might be different from the one in `requirements.txt`, If we have to hardcode in `requirements.txt`, why not move it completely to `requirements.txt` file?

Why we need to keep this file in `/opt/djl` folder?

serving/docker/partition/trt_llm_partition.py
```diff
@@ -1,39 +1,190 @@
...
+                'gather_generation_logits': args.gather_generation_logits,
+                'strongly_typed': True,
+                'force_num_profiles': force_num_profiles_from_env,
+                'weight_sparsity': args.weight_sparsity,
+                'profiling_verbosity': args.profiling_verbosity,
+                'enable_debug_output': args.enable_debug_output,
+                'max_draft_len': args.max_draft_len,
+                'speculative_decoding_mode': speculative_decoding_mode,
+                'input_timing_cache': args.input_timing_cache,
+                'output_timing_cache': args.output_timing_cache,
```
frankfliu, line 116: Let's provide a default value for `output_timing_cache` or just hardcode it as `/tmp/cache.engine`

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -438,47 +422,14 @@ private static Path buildTrtLlmArtifacts(
...
-            logger.warn("Could not identify GPU arch {}", computeCapability);
-            return null;
-        }
-    }
-
     static boolean isValidTrtLlmModelRepo(Path modelPath) throws IOException {
-        // TODO: match model name
         AtomicBoolean isValid = new AtomicBoolean();
         try (Stream<Path> walk = Files.list(modelPath)) {
             walk.filter(Files::isDirectory)
```
frankfliu, line 428:
```suggestion
            walk.filter(Files::isRegularFile)
```

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -438,47 +422,14 @@ private static Path buildTrtLlmArtifacts(
...
                     .forEach(
                             p -> {
-                                Path confFile = p.resolve("config.pbtxt");
-                                // TODO: add stricter check for tokenizer
-                                Path tokenizer = p.resolve("tokenizer_config.json");
-                                if (Files.isRegularFile(confFile)
-                                        && Files.isRegularFile(tokenizer)) {
-                                    logger.info("Found triton model: {}", p);
+                                // TODO: is presence of .engine files sufficient?
+                                if (p.endsWith(".engine")) {
```
frankfliu, line 432:
```suggestion
                                if (p.toString().endsWith(".engine")) {
```

frankfliu: Better use the same logic as python `LLM.get_model_format()`:

```
        Path configFile = modelPath.resolve("config.json");
        if (Files.exists(configFile) && Files.isRegularFile(configFile)) {
            String config = Files.readString(configFile);
            JsonObject json = JsonUtils.GSON.fromJson(config, JsonObject.class);
            return json.has("build_config");
        }
        return false;
```

serving/docker/tensorrt-llm.Dockerfile
siddvenk: I think for the trtllm container, the set of requirements is so small I'll just install them in-line in the docker file. 

The main reason for separation is that --extra-index-url applies globally to all requirements, and we have no control over where a requirement is sourced from and it could be tampered with if the index is compromised (appsec).

will change this to:
```
RUN pip install tensorrt_llm==${trtllm_version} --index-url https://pypi.nvidia.com && \
    pip install tensorrt_llm==${trtllm_version} uvloop ninja
```

serving/docker/tensorrt-llm.Dockerfile
```diff
@@ -67,39 +48,20 @@ RUN mkdir -p /opt/djl/conf && \
...
 COPY partition /opt/djl/partition
 
-
-COPY distribution[s]/ ./
-RUN mv *.deb djl-serving_all.deb || true
-
 # Install OpenMPI and other deps
 ARG DEBIAN_FRONTEND=noninteractive
-RUN apt-get update && apt-get install -y g++ wget unzip openmpi-bin libopenmpi-dev libffi-dev git-lfs rapidjson-dev graphviz && \
+RUN apt-get update && apt-get upgrade -y && apt-get install -y g++ wget unzip openmpi-bin libopenmpi-dev libffi-dev git-lfs rapidjson-dev graphviz cuda-compat-12-8 && \
```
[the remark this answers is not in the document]
siddvenk: fixed

serving/docker/partition/trt_llm_partition.py
siddvenk: I have updated this script pretty significantly @frankfliu. I have exposed BuildConfig, QuantConfig, and CalibConfig here in the AOT script. Runtime Configs will be handled in the entryPoint script.

engines/python/setup/djl_python/lmi_trtllm/trtllm_async_service.py
siddvenk: refactored this a bit so that the properties objects returns a dict of LLMArgs that will be passed here

serving/docker/partition/trt_llm_partition.py
siddvenk: updated

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
siddvenk: went with this

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
siddvenk: went with the other approach you mentioned later

serving/docker/partition/trt_llm_partition.py
siddvenk: went with hardcode of `/tmp/model.cache`

frankfliu:
The only thing missing is: https://github.com/deepjavalibrary/djl-serving/blob/master/engines/python/src/main/java/ai/djl/python/engine/PyModel.java#L185
We need the following:

```
+                    recommendedEntryPoint = "djl_python.lmi_trtllm.trtllm_async_service";
+                    pyEnv.setAsyncMode(true);
```

Everything else looks good.

frankfliu (approved):

ethnzhng (approved):

PR #2884, opened 2025-09-24
[lmiv16-vllm] Apply custom formatters to vLLM Async Handlers

ksuma2109:
## Description ##

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
+++ b/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
@@ -70,33 +70,29 @@ class VLLMHandler:
         self.initialized = False
         self.output_formatter = None
         self.input_formatter = None
-    
+
     def load_formatters(self, model_dir: str):
         """Load custom formatters from model.py"""
         self.input_formatter = get_annotated_function(model_dir, "is_input_formatter")
         self.output_formatter = get_annotated_function(model_dir, "is_output_formatter")
         logger.info(f"Loaded formatters - input: {self.input_formatter}, output: {self.output_formatter}")
-    
+
     def apply_input_formatter(self, decoded_payload, **kwargs):
         """Apply input formatter if available"""
         if self.input_formatter:
             return self.input_formatter(decoded_payload, **kwargs)
         return decoded_payload
-    
+
     def apply_output_formatter(self, output):
         """Apply output formatter if available"""
         if self.output_formatter:
             return self.output_formatter(output)
         return output
-    
-    async def apply_output_formatter_streaming(self, stream_generator):
-        """Apply output formatter to streaming responses"""
-        if not self.output_formatter:
-            async for output in stream_generator:
-                yield output
-        else:
-            async for output in stream_generator:
-                yield self.apply_output_formatter(output)
+
+    async def apply_output_formatter_streaming_raw(self, stream_generator):
+        """Apply output formatter to raw streaming responses"""
+        async for response in stream_generator:
+            yield self.apply_output_formatter(response)
 
     async def initialize(self, properties: dict):
         self.hf_configs = HuggingFaceProperties(**properties)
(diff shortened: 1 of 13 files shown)


engines/python/setup/djl_python/service_loader.py
```diff
@@ -109,6 +112,9 @@ def get_annotated_function(model_dir: str,
         annotated_function = find_decorated_function(service.module,
                                                      decorator_attribute)
         if annotated_function:
+            logging.info(
```
Lokiiiiii, line 115: It looks like a repeated log line ? Can we switch one of them to logging.debug ?

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -61,10 +62,44 @@ def __init__(self):
         self.vllm_properties = None
         self.model_name = None
         self.initialized = False
+        self.output_formatter = None
+        self.input_formatter = None
+
+    def load_formatters(self, model_dir: str):
```
Lokiiiiii: All the new code here is not really specific to vLLM. Let's create a base handler with these generic functions and then have vLLM inherit from the base handler. That way we don't have to make these changes separately for vLLM and TRT-LLM.

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -61,10 +62,44 @@ def __init__(self):
...
+        """Load custom formatters from model.py"""
+        self.input_formatter = get_annotated_function(model_dir,
+                                                      "is_input_formatter")
+        self.output_formatter = get_annotated_function(model_dir,
+                                                       "is_output_formatter")
+        logger.info(
+            f"Loaded formatters - input: {self.input_formatter}, output: {self.output_formatter}"
+        )
+
+    def apply_input_formatter(self, decoded_payload, **kwargs):
```
Lokiiiiii: We need better error handling around both loading and applying formatters. 

Let's create a new Exception class that makes it obvious that the errors are stemming from code extension. We will catch and re-throw all errors from interacting with custom code. We will need to make sure we preserve the stack and the original error message.

Lokiiiiii (changes requested):

tests/integration/llm/client.py
```diff
@@ -1637,6 +1645,45 @@ def test_handler_rolling_batch(model, model_spec):
...
+    if "adapters" in spec:
+        req["adapters"] = spec.get("adapters")[0]
+
+    for stream in stream_values:
+        req["stream"] = stream
+        LOGGER.info(f"req {req}")
+        res = send_json(req)
+        message = res.content.decode("utf-8")
+        LOGGER.info(f"res: {message}")
+        response_checker(res, message)
```
Lokiiiiii, line 1670: We need to check if the input/output formatters actually took effect. Can we do an explicit check here ?

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
Lokiiiiii: Remember, to add a new negative test for this.

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -65,6 +66,16 @@ def __init__(self):
     async def initialize(self, properties: dict):
         self.hf_configs = HuggingFaceProperties(**properties)
         self.vllm_properties = VllmRbProperties(**properties)
+
+        # Load formatters
+        model_dir = properties.get("model_dir", ".")
+        try:
+            self.load_formatters(model_dir)
+        except CustomFormatterError as e:
+            logger.error(
```
Lokiiiiii, line 75: Isn't this repetitive ? I believe we already have error handling in `self.load_formatters`

PR #2906, opened 2025-09-30
Add XGB & SKL Py handlers with CSV/json support

smouaa:
- Add sklearn_handler.py with support for multiple model formats (pickle, joblib, skops, cloudpickle)
- Add xgboost_handler.py with support for multiple model formats (pickle, json/ubj, xgb)
- Modified decode_csv to support handler use cases
- Support both JSON and CSV input/output formats
- Added find_model_file util function
- Added import_utils

## Description ##

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/.github/workflows/close_inactive_issues.yaml
+++ b/.github/workflows/close_inactive_issues.yaml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "0 19 * * *"
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: actions/stale@v10
+        with:
+          days-before-issue-stale: 30
+          days-before-issue-close: 14
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue is stale because it has been open for 30 days with no activity."
+          close-issue-message: "This issue was closed because it has been inactive for 14 days since being marked as stale."
+          days-before-pr-stale: -1
+          days-before-pr-close: -1
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -71,9 +71,70 @@ jobs:
                   json_images=$(jq --compact-output --null-input '$ARGS.positional' --args -- "${images[@]}")
                   echo "images are ${json_images}"
                   echo "images=${json_images}" >> "$GITHUB_OUTPUT"
+    ecr-scan:
+        if: always()
+        needs: [determine_images_to_publish, build]
+        runs-on: ubuntu-latest
+        strategy:
+            fail-fast: false
+            matrix:
+                arch: ${{ fromJson(needs.determine_images_to_publish.outputs.images) }}
+        steps:
+            - name: Configure AWS Credentials
+              uses: aws-actions/configure-aws-credentials@v4
+              with:
+                  role-to-assume: arn:aws:iam::185921645874:role/github-actions-djl-serving
+                  aws-region: us-east-1
+            - name: Get image tag
+              id: get-tag
+              run: |
+                  SERVING_VERSION=$(echo "${{ needs.build.outputs.djl_version }}")
+                  mode=${{ inputs.mode }}
+                  if [[ "$mode" != "release" ]]; then
+                    NIGHTLY="-nightly"
+                  fi
+                  IMAGE_TAG="$SERVING_VERSION-${{ matrix.arch }}${NIGHTLY}"
+                  echo "IMAGE_TAG=$IMAGE_TAG" >> $GITHUB_OUTPUT
+            - name: Get image digest
+              id: get-digest
+              run: |
+                  DIGEST=$(aws ecr describe-images \
+                    --repository-name djl-ci-temp \
+                    --image-ids imageTag=${{ steps.get-tag.outputs.IMAGE_TAG }} \
+                    --region us-east-1 \
+                    --query 'imageDetails[0].imageDigest' \
+                    --output text)
+                  echo "DIGEST=$DIGEST" >> $GITHUB_OUTPUT
+            - name: Check Inspector findings
+              run: |
+                  REPO_URI="185921645874.dkr.ecr.us-east-1.amazonaws.com/djl-ci-temp"
+                  RESOURCE_ID="${REPO_URI}@${{ steps.get-digest.outputs.DIGEST }}"
+                  
+                  echo "Checking vulnerabilities for: $RESOURCE_ID"
+                  sleep 30
+                  
+                  FINDINGS=$(aws inspector2 list-findings \
+                    --filter-criteria '{"ecrImageHash":[{"comparison":"EQUALS","value":"${{ steps.get-digest.outputs.DIGEST }}"}]}' \
+                    --region us-east-1 \
+                    --output json)
+                  
+                  HIGH=$(echo "$FINDINGS" | jq '[.findings[] | select(.severity=="HIGH")] | length')
+                  CRITICAL=$(echo "$FINDINGS" | jq '[.findings[] | select(.severity=="CRITICAL")] | length')
+                  
+                  echo "Scan Results for ${{ matrix.arch }}:"
+                  echo "HIGH: $HIGH"
+                  echo "CRITICAL: $CRITICAL"
+                  
+                  if [ "$HIGH" -gt 0 ] || [ "$CRITICAL" -gt 0 ]; then
+                    echo "ERROR: Found HIGH or CRITICAL vulnerabilities"
+                    echo "$FINDINGS" | jq '.findings[] | select(.severity=="HIGH" or .severity=="CRITICAL") | {title, severity, description}'
+                    exit 1
+                  fi
+                  
+                  echo "No HIGH or CRITICAL vulnerabilities found"
     publish:
         if: always()
-        needs: [determine_images_to_publish]
+        needs: [determine_images_to_publish, ecr-scan]
         uses: ./.github/workflows/docker_publish.yml
         secrets: inherit
         with:
--- a/engines/python/setup/djl_python/tests/test_encode_decode.py
+++ b/engines/python/setup/djl_python/tests/test_encode_decode.py
@@ -1,0 +1,268 @@
+import unittest
+import csv
+import numpy as np
+from io import StringIO
+from unittest.mock import Mock, patch
+from djl_python.inputs import Input
+from djl_python.outputs import Output
+from djl_python.encode_decode import decode, encode, decode_csv, encode_csv
+
+
+class TestEncodeDecode(unittest.TestCase):
+
+    def setUp(self):
+        self.mock_input = Mock(spec=Input)
+        self.mock_output = Mock(spec=Output)
+        self.mock_output.add_as_string = Mock()
+        self.mock_output.add_as_json = Mock()
+        self.mock_output.add_as_numpy = Mock()
+        self.mock_output.add_as_npz = Mock()
+        self.mock_output.add_property = Mock()
+
+    def test_decode_csv_valid_inputs_header(self):
+        csv_content = "inputs,other\ntest input 1,value1\ntest input 2,value2"
+        self.mock_input.get_as_string.return_value = csv_content
+
+        result = decode_csv(self.mock_input)
+
+        expected = {"inputs": ["test input 1", "test input 2"]}
+        self.assertEqual(result, expected)
+
+    def test_decode_csv_valid_question_header(self):
+        csv_content = "question,context\nWhat is AI?,Technology context\nWhat is ML?,Machine learning context"
+        self.mock_input.get_as_string.return_value = csv_content
+
+        result = decode_csv(self.mock_input)
+
+        expected = {
+            "inputs": [{
+                "question": "What is AI?",
+                "context": "Technology context"
+            }, {
+                "question": "What is ML?",
+                "context": "Machine learning context"
+            }]
+        }
+        self.assertEqual(result, expected)
+
+    def test_decode_csv_invalid_header(self):
+        csv_content = "invalid,header\nvalue1,value2"
+        self.mock_input.get_as_string.return_value = csv_content
+
+        with self.assertRaises(ValueError) as context:
+            decode_csv(self.mock_input)
+
+        self.assertIn("correct CSV with Header columns",
+                      str(context.exception))
+
+    def test_encode_csv_list_of_dicts(self):
+        content = [{
+            "name": "Axl",
+            "age": "30"
+        }, {
+            "name": "Fiona",
+            "age": "25"
+        }]
+
+        result = encode_csv(content)
+
+        reader = csv.DictReader(StringIO(result))
+        rows = list(reader)
+        self.assertEqual(len(rows), 2)
+        self.assertEqual(rows[0]["name"], "Axl")
+        self.assertEqual(rows[1]["name"], "Fiona")
+
+    def test_encode_csv_single_dict(self):
+        content = [{"name": "Axl", "age": "30"}]
+
+        result = encode_csv(content)
+
+        reader = csv.DictReader(StringIO(result))
+        rows = list(reader)
+        self.assertEqual(len(rows), 1)
+        self.assertEqual(rows[0]["name"], "Axl")
+
+    def test_decode_no_content_type(self):
+        self.mock_input.get_as_bytes.return_value = None
+
+        result = decode(self.mock_input, None)
+
+        expected = {"inputs": ""}
+        self.assertEqual(result, expected)
+
+    def test_decode_no_content_type_with_json(self):
+        test_data = {"test": "data"}
+        self.mock_input.get_as_bytes.return_value = b'{"test": "data"}'
+        self.mock_input.get_as_json.return_value = test_data
+
+        result = decode(self.mock_input, None)
+
+        self.assertEqual(result, test_data)
+
+    def test_decode_application_json(self):
+        test_data = {"message": "hello"}
+        self.mock_input.get_as_json.return_value = test_data
+
+        result = decode(self.mock_input, "application/json")
+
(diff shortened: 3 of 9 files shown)


engines/python/setup/djl_python/encode_decode.py
```diff
@@ -22,9 +22,10 @@
 
 
 def decode_csv(inputs: Input):  # type: (str) -> np.array
-    stream = StringIO(inputs.get_as_string())
+    csv_string = inputs.get_as_string()
```
Lokiiiiii: Could you explain with this is necessary ?

engines/python/setup/djl_python/encode_decode.py
```diff
@@ -37,6 +38,13 @@ def decode_csv(inputs: Input):  # type: (str) -> np.array
         return {"inputs": request_list}
 
 
+def decode_csv_numeric(inputs: Input):  # type: (Input) -> np.array
+    csv_string = inputs.get_as_string()
+    reader = csv.reader(StringIO(csv_string))
+    data = [[float(cell) for cell in row] for row in reader if row]
```
Lokiiiiii: Python list comprehension is not as fast as C, can we try some alternatives like - 

`pandas.read_csv(StringIO(csv_string), header=None).to_numpy()`
or 
`numpy.genfromtxt(StringIO(csv_string), delimiter=",") `

?

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+
+import pickle
+import numpy as np
+import os
+import glob
+from typing import Optional
+from djl_python import Input, Output
+from djl_python.encode_decode import decode, decode_csv_numeric, encode_csv
+
+try:
```
Lokiiiiii: Prefer encapsulating this akin to https://github.com/huggingface/transformers/blob/41e5abac5cb49983a08ddef3e8645d6efd23c8f3/src/transformers/utils/import_utils.py

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+    cloudpickle = None
+
+
+class SklearnHandler:
+
+    def __init__(self):
+        self.model = None
+        self.initialized = False
+
+    def _find_model_file(self, model_dir: str, extensions: list) -> str:
```
Lokiiiiii: Let's make a helper function outside the scope of sklearn so others can use this.

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+            matches = glob.glob(pattern)
+            if matches:
+                return matches[0]  # Return first match
+        return None
+
+    def initialize(self, properties: dict):
+        model_dir = properties.get("model_dir")
+
+        # Check for preferred model format from environment variable
+        preferred_format = os.environ.get("SKLEARN_MODEL_FORMAT", "auto")
```
Lokiiiiii: By default, we should prefer the safest option which is `skops`

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+            model_file = self._find_model_file(model_dir, [preferred_format])
+            if not model_file:
+                raise FileNotFoundError(
+                    f"No model file found with format '{preferred_format}' in {model_dir}"
+                )
+
+            if preferred_format == "skops" and sio:
+                self.model = sio.load(
+                    model_file,
+                    trusted=sio.get_untrusted_types(file=model_file))
```
Lokiiiiii: We don't want to implicitly trust the model.

The user needs to tell us through an ENV variable what are the trusted types. `SKLEARN_SKOPS_TRUSTED_TYPES`

Lokiiiiii: We only want to use pickle based model loaders when - 

`SKLEARN_MODEL_FORMAT=xyz`
And 
`TRUST_INSECURE_PICKLE_FILES=true`

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+            elif preferred_format == "joblib" and joblib:
+                self.model = joblib.load(model_file)
+            elif preferred_format == "pkl":
+                with open(model_file, 'rb') as f:
+                    self.model = pickle.load(f)
+            elif preferred_format == "cloudpkl" and cloudpickle:
+                with open(model_file, 'rb') as f:
+                    self.model = cloudpickle.load(f)
+            else:
+                raise ImportError(
```
Lokiiiiii: Can we also add all these libs to the Dockerfile ? Build a new container and scan it for CVEs - https://docs.aws.amazon.com/AmazonECR/latest/userguide/image-scanning.html

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+            # Try skops first (most secure)
+            if sio:
+                model_file = self._find_model_file(model_dir, ["skops"])
+                if model_file:
+                    self.model = sio.load(
+                        model_file,
+                        trusted=sio.get_untrusted_types(file=model_file))
+
+            # Try joblib
+            if not model_file and joblib:
```
Lokiiiiii: Condition the rest on `TRUST_INSECURE_PICKLE_FILES`

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,157 @@
...
+            input_map = decode(inputs, content_type)
+            data = input_map.get("inputs") if isinstance(input_map,
+                                                         dict) else input_map
+            X = np.array(data)
+
+        predictions = self.model.predict(X)
+
+        outputs = Output()
+        if "text/csv" in accept:
+            csv_data = "\n".join(str(pred) for pred in predictions)
```
Lokiiiiii: is predictions an np.array ? I am looking for a method more efficient that python's join

Lokiiiiii (changes requested):

engines/python/setup/djl_python/encode_decode.py
smouaa: I initially tried using this function to parse CSV inputs since it was already defined and implemented but noticed that it tried to use the splitlines method on an undefined variable (string_like). The original function is shown below:

`def decode_csv(inputs: Input):  # type: (str) -> np.array
    stream = StringIO(inputs.get_as_string())
    # detects if the incoming csv has headers
    if not any(header in string_like.splitlines()[0].lower()
               for header in ["question", "context", "inputs"]):
        raise ValueError(
            "You need to provide the correct CSV with Header columns to use it with the inference toolkit default handler.",
        )
    # reads csv as io
    request_list = list(csv.DictReader(stream))
    if "inputs" in request_list[0].keys():
        return {"inputs": [entry["inputs"] for entry in request_list]}
    else:
        return {"inputs": request_list}`

I then realized that it only accepted CSV with headers because the original function was meant for huggingface transformer models (according to past PRs) so I wrote the decode_csv_numeric function rather than modify this function to suit the SKLearn handler use case in case it's being used elsewhere.

engines/python/setup/djl_python/encode_decode.py
smouaa: Yes, will make this change.

Lokiiiiii:
Note: Add tests, benchmarks

Lokiiiiii:
Note: Add documentation and demo notebook

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+from djl_python.import_utils import joblib, cloudpickle, skops_io as sio
+
+
+class SklearnHandler:
+
+    def __init__(self):
+        self.model = None
+        self.initialized = False
+
+    def _get_trusted_types(self):
```
Lokiiiiii: Let's log the value for posterity

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+import os
+from io import StringIO
+from typing import Optional
+from djl_python import Input, Output
+from djl_python.encode_decode import decode
+from djl_python.utils import find_model_file
+from djl_python.import_utils import joblib, cloudpickle, skops_io as sio
+
+
+class SklearnHandler:
```
Lokiiiiii, line 26: Add Custom Input/Output formatters

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+        if not trusted_types_str:
+            raise ValueError(
+                "SKLEARN_SKOPS_TRUSTED_TYPES environment variable must be set to load skops models. "
+                "Example: SKLEARN_SKOPS_TRUSTED_TYPES='sklearn.ensemble._forest.RandomForestClassifier,numpy.ndarray'"
+            )
+        return [t.strip() for t in trusted_types_str.split(",") if t.strip()]
+
+    def initialize(self, properties: dict):
+        model_dir = properties.get("model_dir")
+        model_format = os.environ.get("SKLEARN_MODEL_FORMAT", "skops")
```
Lokiiiiii: Let's model this as a djl-serving property. Properties can be set through ENV variables or serving.properties file. We want to maintain the same UI.

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class SklearnHandler:
+
+    def __init__(self):
+        self.model = None
+        self.initialized = False
+
+    def _get_trusted_types(self):
+        trusted_types_str = os.environ.get("SKLEARN_SKOPS_TRUSTED_TYPES", "")
```
Lokiiiiii: Let's model this as a djl-serving property. Properties can be set through ENV variables or serving.properties file. We want to maintain the same UI.

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+        extensions = format_extensions.get(model_format)
+        if not extensions:
+            raise ValueError(
+                f"Unsupported model format: {model_format}. Supported formats: skops, joblib, pickle, cloudpickle"
+            )
+
+        model_file = find_model_file(model_dir, extensions)
+        if not model_file:
+            raise FileNotFoundError(
+                f"No model file found with format '{model_format}' in {model_dir}"
```
Lokiiiiii, line 68: Let's log the expected extension types in the error msg

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+            )
+
+        if model_format == "skops":
+            trusted_types = self._get_trusted_types()
+            self.model = sio.load(model_file, trusted=trusted_types)
+        else:
+            if os.environ.get("TRUST_INSECURE_PICKLE_FILES",
```
Lokiiiiii: Let's model this as a djl-serving property. Properties can be set through ENV variables or serving.properties file. We want to maintain the same UI.

engines/python/setup/djl_python/utils.py
```diff
@@ -161,3 +164,21 @@ def get_input_details(requests, errors, batch):
...
+        extensions: List of file extensions to search for (without dots)
+        
+    Returns:
+        Path to first matching model file, or None if not found
+    """
+    for ext in extensions:
+        pattern = os.path.join(model_dir, f"*.{ext}")
+        matches = glob.glob(pattern)
+        if matches:
+            return matches[0]  # Return first match
```
Lokiiiiii: "first match" is not deterministic, it is dependent on the parsing order. You might end up loading 2 different models in 2 different runs of the exact same artifacts. 

Throw an error on multiple matches instead to keep it simple - Only 1 model artifact supported per model directory.

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -0,0 +1,122 @@
...
+                                                         dict) else input_map
+            X = np.array(data)
+
+        if X.ndim == 1:
+            X = X.reshape(1, -1)
+
+        predictions = self.model.predict(X)
+
+        outputs = Output()
+        if "text/csv" in accept:
```
Lokiiiiii, line 142: Do validation of the user provided accept types. if the user requests something like ndarray, please throw a validation error.

Lokiiiiii (changes requested):
Let's add some tests for - 
1. CSV inputs - with & without headers
2. multiple model artifacts
3. Improper ENV variable setup
4. garbage string in `SKLEARN_SKOPS_TRUSTED_TYPES`
5. unsupported accept types
6. Different input shapes
7. Ragged arrays in input

engines/python/setup/djl_python/xgboost_handler.py
```diff
@@ -0,0 +1,112 @@
...
+        extensions = ["pkl", "model", "xgb", "bst", "json", "ubj"]
+
+        model_file = find_model_file(model_dir, extensions)
+        if not model_file:
+            raise FileNotFoundError(
+                f"No XGBoost model file found in {model_dir}")
+
+        try:
+            with open(model_file, 'rb') as f:
+                self.model = pkl.load(f)
```
Lokiiiiii: Similar to skl, we should categorize the file formats as safe and unsafe, create the appropriate serving properties. We should always prefer the safe formats - json and ubjson unless the user tells us otherwise.

tests/integration/tests.py
```diff
@@ -142,6 +144,316 @@ def test_python_dynamic_batch(self):
...
+                cmd="serve -m sklearn_test::Python=file:/opt/ml/model/sklearn_model_v2.zip"
+            )
+            # Test with sample data and verify response
+            test_data = {"inputs": [[1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0]]}
+            response = requests.post(
+                "http://localhost:8080/predictions/sklearn_test",
+                json=test_data,
+                headers={"Content-Type": "application/json", "Accept": "application/json"}
+            )
+            assert response.status_code == 200
```
Lokiiiiii: Why would this succeed without TRUST variable enabled ?

tests/integration/tests.py
```diff
@@ -142,6 +144,316 @@ def test_python_dynamic_batch(self):
...
+            # Wrong number of features (model expects 10, sending 5)
+            test_data = {"inputs": [[1.0, 2.0, 3.0, 4.0, 5.0]]}
+            response = requests.post(
+                "http://localhost:8080/predictions/sklearn_test",
+                json=test_data,
+                headers={"Content-Type": "application/json"}
+            )
+            assert response.status_code == 424  # Should fail with wrong input shape
+
+    def test_sklearn_ragged_arrays(self):
```
Lokiiiiii: Let's test ragged csv as well. 

And add the following test cases - 
1. Empty rows
1. non-numeric values

tests/integration/tests.py
```diff
@@ -142,6 +144,316 @@ def test_python_dynamic_batch(self):
             )
             os.system("./test_client.sh image/jpg models/kitten.jpg")
 
+    def test_sklearn_model(self):
```
Lokiiiiii: Let's split this into a new file. This is getting too big to read.

tests/integration/tests.py
```diff
@@ -142,6 +144,316 @@ def test_python_dynamic_batch(self):
...
+            # Check custom output format
+            assert "probability" in result
+            assert "prediction" in result
+            assert "model_version" in result
+            assert "processed_by" in result
+            assert result["model_version"] == "1.0"
+            assert result["processed_by"] == "xgboost_custom"
+            assert isinstance(result["probability"], float)
+            assert result["prediction"] in [0, 1]
+
```
Lokiiiiii: Add more tests for - 
1. invalid content type
2. invalid accept type
3. X-types. JSON inputs to CSV outputs and vice versa.

Lokiiiiii (changes requested):

serving/docker/Dockerfile
```diff
@@ -64,21 +64,26 @@ LABEL djl-serving-version=$djl_serving_version
 
 FROM base AS cpu-full
 
-ARG python_version=3.10
+ARG python_version=3.11
```
Lokiiiiii, line 67: What is the numpy version in this container ? Can we upgrade it to the latest as well ?

tests/integration/tests.py
smouaa: There is a serving.properties file in the zip file that sets the trust variable:
```
engine=Python
option.entryPoint=djl_python.sklearn_handler
option.model_format=pickle
option.trust_insecure_model_files=true
```

serving/docker/Dockerfile
smouaa: Just checked and it's 1.26.4--I'll update this

engines/python/setup/djl_python/sklearn_handler.py
```diff
@@ -33,11 +33,12 @@ def __init__(self):
         self.custom_predict_formatter = None
 
     def _get_trusted_types(self, properties: dict):
-        trusted_types_str = properties.get("skops_trusted_types", "")
+        trusted_types_str = (properties.get("skops_trusted_types", "")
+                             or os.environ.get("SKOPS_TRUSTED_TYPES", ""))
```
Lokiiiiii: Why do we need both properties and env variables ? properties are automatically derived from env variables. So we should align on properties.

smouaa: When I ran test_sklearn_skops_with_valid_trusted_types, the model server couldn't find skops_trusted_types unless I used os.environ.get("SKOPS_TRUSTED_TYPES", "") directly. I thought the properties method wasn't finding all environment variables, so I changed it. However I just took another look at the properties class and you're right, properties should be automatically derived from the env variables--I'll continue investigating why it's failing to detect SKOPS_TRUSTED_TYPES

PR #2918, opened 2025-10-22
Upgrade vllm to 0.11.0 and set vllm_aysnc_service.py as the default handler

ksuma2109:
## Description ##

Brief description of what this PR is about

- If this change is a backward incompatible change, why must this change be made?
- Interesting edge cases to note here

## Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] This change requires a documentation update

## Checklist:
- [ ] Please add the link of [**Integration Tests Executor** run](https://github.com/deepjavalibrary/djl-serving/actions/workflows/integration_execute.yml) with related tests.
- [ ] Have you [manually built the docker image](https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docker/README.md#build-docker-image) and verify the change?
- [ ] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L72); One example would be `pytest tests.py -k "TestCorrectnessLmiDist"  -m "lmi_dist"`
- [ ] Have you added tests that prove your fix is effective or that this feature works?
- [ ] Has code been commented, particularly in hard-to-understand areas?
- [ ] Have you made corresponding changes to the documentation?

## Feature/Issue validation/testing

Please describe the Unit or Integration tests that you ran to verify your changes and relevant result summary. Provide instructions so it can be reproduced.
Please also list any relevant details for your test configuration.

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

PR #2931, opened 2025-10-30
[fix] Fix setting default rolling batch and extract lora adapter

xyang16:
## Description ##

- Do not set async mode for default case, because async mode might not be supported
- pop adapter from payload, to keep it consistent with other payload behavior, see https://github.com/deepjavalibrary/djl-serving/blob/89403fe24e0813c3afd55107075caa6bb1040da3/engines/python/setup/djl_python/input_parser.py#L171

## Type of change

## Checklist:

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/engines/python/setup/djl_python/async_utils.py
+++ b/engines/python/setup/djl_python/async_utils.py
@@ -128,7 +128,7 @@ def _extract_lora_adapter(raw_request, decoded_payload):
             SAGEMAKER_ADAPTER_IDENTIFIER_HEADER)
         logging.debug(f"Found adapter in headers: {adapter_name}")
     elif "adapters" in decoded_payload:
-        adapter_name = decoded_payload.get("adapters")
+        adapter_name = decoded_payload.pop("adapters")
         logging.debug(f"Found adapter in payload: {adapter_name}")
 
     return adapter_name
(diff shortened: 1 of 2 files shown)


Lokiiiiii:
>Do not set async mode for default case, because async mode might not be supported

Can we quantify where async mode might not be supported ? Maybe we can split the use-cases and set the adequate ENV variables in the Dockerfile ?

xyang16:
Before it was not set, see https://github.com/deepjavalibrary/djl-serving/blob/0.34.0-dlc/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java#L83

I added recently in https://github.com/deepjavalibrary/djl-serving/pull/2918/commits/be55740597f79b28cbdb9d0e03048de7b8511db1, but now I think better to revert it. Because tensorrtllm and cpu containers doesn't support async now. So it's safer not to enable async mode for the default case.

Lokiiiiii:
Folks, async handler is the recommended pathway in vLLM LMI container. Let's consider setting the required ENV variables in the LMI Dockerfile to provide a better experience for LMI customers.

xyang16:
For vllm, it's handled via env variables here https://github.com/deepjavalibrary/djl-serving/blob/master/wlm/src/main/java/ai/djl/serving/wlm/LmiConfigRecommender.java#L78-L80.

The default case is for containers like cpu or aarch64.

PR #2934, opened 2025-10-31
Changed to use custom s5cmd build w go 1.24.2 to address CVEs

smouaa:
Also added new integration tests for DJL CPU container to workflow

## Description ##

## Type of change

## Checklist:
- [ ] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L98); One example would be `pytest tests.py -k "TestVllm1" -m "vllm"`

- [ ] Test A
Logs for Test A

- [ ] Test B
Logs for Test B

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -184,6 +184,15 @@ jobs:
           - test: TestStatefulModel
             instance: g6
             failure-prefix: lmi
+          - test: TestSageMakerCompatibility
+            instance: cpu
+            failure-prefix: cpu
+          - test: TestCustomFormatters
+            instance: cpu
+            failure-prefix: cpu
+          - test: TestXgbSkl
+            instance: cpu
+            failure-prefix: cpu
     outputs:
       failure_cpu: ${{ steps.test-failure.outputs.failure_cpu }}
       failure_gpu: ${{ steps.test-failure.outputs.failure_gpu }}
--- a/serving/docker/scripts/install_s5cmd.sh
+++ b/serving/docker/scripts/install_s5cmd.sh
@@ -4,17 +4,18 @@ set -ex
 
 ARCH=$1
 
+# Download custom s5cmd binary built with Go 1.24.2
 if [[ $ARCH == "aarch64" ]]; then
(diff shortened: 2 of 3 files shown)


serving/docker/scripts/install_s5cmd.sh
```diff
@@ -4,17 +4,18 @@ set -ex
 
 ARCH=$1
 
+# Download custom s5cmd binary built with Go 1.24.2
 if [[ $ARCH == "aarch64" ]]; then
-  curl https://github.com/peak/s5cmd/releases/download/v2.3.0/s5cmd_2.3.0_Linux-arm64.tar.gz -L -o s5cmd.tar.gz
+  curl -f https://publish.djl.ai/s5cmd/go1.24.2/s5cmd-linux-arm64 -L -o s5cmd
```
xyang16, line 9: qq, this is not available publicly?

xyang16 (approved):

.github/workflows/integration.yml
```diff
@@ -184,6 +184,15 @@ jobs:
           - test: TestStatefulModel
             instance: g6
             failure-prefix: lmi
+          - test: TestSagemakerCompatibility
```
xyang16: You need to add TestSagemakerCompatibility to https://github.com/deepjavalibrary/djl-serving/blob/master/tests/integration/tests.py

serving/docker/scripts/install_s5cmd.sh
smouaa: No, s5cmd does not have a public release with go 1.24.2 which has fixes for the security concerns causing the CVEs

.github/workflows/integration.yml
smouaa: Should I also add TestXgbSkl and TestCustomFormatters to tests.py? I was told to separate them into different files because it made tests.py very long

xyang16: I see, it doesn't work if it's not in tests.py though. But please double check.

xyang16: One thing you can do is to import TestSagemakerCompatibility, TestXgbSkl and TestCustomFormatters in tests.py if you want to keep them separate.

PR #2953, opened 2025-11-19
[ci] Remove --runtime=nvidia from launch_container.sh

ethnzhng:
## Description ##

The `--runtime=nvidia` docker run flag is being deprecated and doesn't work in all environments. It is not needed as long as the `--gpus` flag is set

### DIFF
--- a/tests/integration/launch_container.sh
+++ b/tests/integration/launch_container.sh
@@ -85,9 +85,9 @@ fi
 
 is_llm=false
 if [[ "$platform" == *"-gpu"* ]]; then # if the platform has cuda capabilities
-  runtime="nvidia"
+  gpus="all"
 elif [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"* ]]; then # Runs multi-gpu
-  runtime="nvidia"
+  gpus="all"
   is_llm=true
   if [[ "$(is_p4d_or_p5)" == *"true"* || $is_multi_node ]]; then
     shm="20gb"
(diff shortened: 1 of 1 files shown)


tests/integration/launch_container.sh
```diff
@@ -157,7 +157,7 @@ if $is_multi_node; then
     -e DJL_LEADER_ADDR=${leader_hostname} \
     -e DJL_WORKER_ADDR_FORMAT="${LWS_NAME}-${GROUP_INDEX}-%d.${LWS_NAME}.${NAMESPACE}" \
     ${env_file} \
-    ${runtime:+--runtime="${runtime}"} \
+    ${runtime:+--gpus=all} \
```
xyang16: It's already setting `--gpus '"device=0,1"'` in line 140. So just need to remove this line.

xyang16: Also remove `runtime="nvidia"` in line 90

tests/integration/launch_container.sh
```diff
@@ -84,10 +84,7 @@ if [[ "$(support_nvme)" == *"true"* ]]; then
 fi
 
 is_llm=false
-if [[ "$platform" == *"-gpu"* ]]; then # if the platform has cuda capabilities
-  runtime="nvidia"
 elif [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"* ]]; then # Runs multi-gpu
```
xyang16:
```suggestion
if [[ "$platform" == *"lmi"* || "$platform" == *"trtllm"* || "$platform" == *"tensorrt-llm"* ]]; then # Runs multi-gpu
```

tests/integration/launch_container.sh
```diff
@@ -272,7 +264,6 @@ else
     -e AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY \
     -e AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN \
     $uid_mapping \
```
xyang16, line 273: Just realized this docker run doesn't have --gpus, we can add in line 253.

```
    --gpus=all \
```

xyang16: also line 184, 213, 234

xyang16 (approved):

PR #2954, opened 2025-11-19
Fix race condition in LoRA streaming requests

Lokiiiiii:
## The Bug

When using vLLM async service with LoRA adapters, streaming requests were not receiving the `lora_request` parameter. This was due to an async generator timing issue in the original implementation.

### Original Implementation (Buggy)

```python
if processed_request.lora_request:
    original_add_request = self.vllm_engine.add_request
    
    async def add_request_with_lora(*args, **kwargs):
        kwargs['lora_request'] = processed_request.lora_request
        return await original_add_request(*args, **kwargs)
    
    self.vllm_engine.add_request = add_request_with_lora
    try:
        response = await processed_request.inference_invoker(...)
    finally:
        self.vllm_engine.add_request = original_add_request
```

**Problem**: For streaming requests, `inference_invoker` returns an `AsyncGenerator` that hasn't started executing yet. The wrapper is removed in the `finally` block before the generator actually runs and calls `add_request`.

## The Fix

The fix passes `adapter` directly as a parameter to the OpenAI Request object and then send it to inference invoker instead of wrapping the method:

```python
processed_request.vllm_request.model = adapter_name
response = await processed_request.inference_invoker(
            processed_request.vllm_request
```

This ensures the `lora_request` is properly passed to vLLM's OpenAI serving methods for both streaming and non-streaming requests.

## Testing

Manual testing confirms adapter is getting utilized

### DIFF
--- a/engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
+++ b/engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
@@ -55,7 +55,7 @@ def convert_lmi_schema_to_completion_request(
     parameters = payload.get("parameters", {})
 
     completion_dict = {
-        "model": payload.pop("model", None),
+        "model": payload.pop("model"),
         "prompt": payload.pop("inputs"),
         "max_tokens": parameters.pop("max_new_tokens", 30),
         "echo": parameters.pop("return_full_text", False),


HappyAmazonian (dismissed):

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -239,18 +239,9 @@ async def inference(
...
-
-            self.vllm_engine.add_request = add_request_with_lora
-            try:
-                response = await processed_request.inference_invoker(
-                    processed_request.vllm_request)
-            finally:
-                self.vllm_engine.add_request = original_add_request
+            response = await processed_request.inference_invoker(
+                processed_request.vllm_request,
+                lora_request=processed_request.lora_request)
```
HappyAmazonian: does `inference_invoker` take lora_request? I check it refers to the vllm openai chat service.

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -165,6 +165,10 @@ def preprocess_request(self, inputs: Input) -> ProcessedRequest:
             logging.info(
                 f"Using LoRA request: {lora_request.lora_name} (ID: {lora_request.lora_int_id})"
             )
+            # Register the LoRA request with the model registry so vLLM can find it
+            self.model_registry.lora_requests[adapter_name] = lora_request
```
xyang16: I think this should be done in register_adapter. There's no need to register adapter during invocation.

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
```diff
@@ -55,6 +55,7 @@ def convert_lmi_schema_to_completion_request(
     parameters = payload.get("parameters", {})
 
     completion_dict = {
+        "model": payload.pop("model"),
```
xyang16, line 58: DJLServing doesn't have "model" in payload. This will throw KeyError.

xyang16: see https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docs/lmi/user_guides/lmi_input_output_schema.md#request-schema

Lokiiiiii: Added `payload.pop("model", None)`

Lokiiiiii:
Manual testing confirms adapter is getting utilized

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
xyang16: Since DJLServing doesn't have "model" in payload, it will always be None. maybe just remove `"model": payload.pop("model"),` it can be added to decode_payload in vllm_async_service.py

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
```diff
@@ -55,6 +55,7 @@ def convert_lmi_schema_to_completion_request(
     parameters = payload.get("parameters", {})
 
     completion_dict = {
+        "model": payload.pop("model"), None,
```
xyang16:
```
        "model": payload.pop("model", None),
```

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
Lokiiiiii: Removed

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
Lokiiiiii: Fixed

engines/python/setup/djl_python/lmi_vllm/request_response_utils.py
Lokiiiiii: For TextCompletions we have this logic - https://github.com/deepjavalibrary/djl-serving/blob/e8bc912187ea7c12f19633a1bac59e04c7b8a7df/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py#L176-L182

decoded_payload -> convert_lmi_schema_to_completion_request -> vllm_request

Are you suggesting that I add it directly to vllm_request skipping decoded payload ?

xyang16 (approved):

PR #2982, opened 2025-12-19
LCNC lmcache auto config

smouaa:
## Description ##

- Allows user to opt into auto configuration for LMCache based on available CPU/disk space
- **New** Updated to use LMCache configuration env variables rather than writing a yaml file to avoid potential issues with overwriting existing files with the same name
- **New** Added fast fail for maxWorkers >1, option.lmcache_config_file already set by user, etc.
- Tested locally with Qwen2.5-1.5B, , Qwen2.5-7B, and Qwen2.5-72B with varying TP degrees
- Fixed issue where OPTION_LMCACHE_CONFIG_FILE wasn't getting parsed correctly to be read by LMCache
- Removed redundant LMCache benchmarking integration test class and cleaned up to reflect fix above
- Tested w/ unit and integration tests

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] This change requires a documentation update

## Checklist:
- [X] Have you [manually built the docker image](https://github.com/deepjavalibrary/djl-serving/blob/master/serving/docker/README.md#build-docker-image) and verify the change?
- [X] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L98); One example would be `pytest tests.py -k "TestVllm1" -m "vllm"`
- [X] Have you added tests that prove your fix is effective or that this feature works?
- [X] Has code been commented, particularly in hard-to-understand areas?
- [X] Have you made corresponding changes to the documentation?

Created unit tests in `djl-serving/engines/python/setup/djl_python/tests/test_lmcache_utils.py`
Ran LMCache related integration tests

### DIFF
--- a/engines/python/setup/djl_python/lmcache_utils.py
+++ b/engines/python/setup/djl_python/lmcache_utils.py
@@ -26,7 +26,7 @@ from huggingface_hub import HfApi
 
 logger = logging.getLogger(__name__)
 
-LMCACHE_DIR = '/tmp/cache'
+LMCACHE_DIR = '/tmp/lmcache'
 
 
 def get_available_cpu_memory() -> float:
@@ -84,7 +84,7 @@ def calculate_cpu_cache_size(properties: Dict[str, str],
 
 
 def calculate_disk_cache_size(properties: Dict[str, str],
-                             model_size_gb: float) -> float:
+                              model_size_gb: float) -> float:
     """
     Calculate disk cache size for LMCache based on available disk space and TP degree.
     
@@ -98,7 +98,7 @@ def calculate_disk_cache_size(properties: Dict[str, str],
     tp_degree = int(properties.get("tensor_parallel_degree", 1))
 
     try:
-        stat = shutil.disk_usage('/tmp')
+        stat = shutil.disk_usage('/tmp')  # assumes /tmp is a NVMe
         total_disk_gb = stat.total / (1024**3)
         logger.info(f"Total disk space: {total_disk_gb:.1f}GB")
     except Exception as e:
@@ -329,7 +329,8 @@ def apply_lmcache_auto_config(model_path: str,
             f"{k}={v}" for k, v in list(lmcache_env_vars.items())[:3])
         logger.warning(
             f"LMCache auto-configuration detected existing LMCACHE environment variables: {env_var_list}. "
-            f"Applicable LMCache configuration variables will be overwritten by auto-configuration.")
+            f"Applicable LMCache configuration variables will be overwritten by auto-configuration."
+        )
 
     updated_properties = properties.copy()
 
--- a/serving/docs/lmcache_performance.md
+++ b/serving/docs/lmcache_performance.md
@@ -79,7 +79,7 @@ max_local_cpu_size: 5.0 # Changes with Model Size
 # 256 Tokens per KV Chunk
 chunk_size: 256
 # Enable Disk backend
-local_disk: "file://tmp/cache/" # Fixed for SM customers
+local_disk: "file://tmp/lmcache/" # Fixed for SM customers
 # 5GB of Disk memory
 max_local_disk_size: 5.0 # Changes with Model Size
 # Disable OS page cache in favor of CPU Pinned Memory
@@ -100,10 +100,10 @@ The auto-configuration feature:
 - Calculates cache sizes based on model size and instance resources
 - Utilizes LMCache's LazyMemoryAllocator to reduce startup time
 - Sets up both CPU and disk backends with appropriate sizes
-- Generates the LMCache configuration file automatically
 - Requires `maxWorkers=1` (automatically enforced)
+- Only supported for Tensor Parallelism
 
-**Note**: Auto-configuration is designed for single model deployments. If serving multiple model copies on the same server, use manual configuration.
+**Note**: Auto-configuration is designed for serving a single model per container instance. For serving multiple models or model copies use SageMaker's Inference Components which ensures resource isolation between models and model copies.
 
 ## Deployment Recommendations
 
@@ -111,7 +111,7 @@ The auto-configuration feature:
 2. **Use NVMe with O_DIRECT enabled** for workloads requiring larger cache capacity
 3. **Implement session-based sticky routing** on SageMaker Classic to maximize cache hit rates
 4. **Consider model architecture**: Models with different KV head configurations (e.g., Llama 3 8B vs Qwen 2.5-7B) will have different offloading thresholds
-5. **For multi-model deployments**: Use manual configuration instead of auto-config
+5. **Use auto-configuration** for rapid single model deployments
 
 ## Performance Validation
 
--- a/tests/integration/tests.py
+++ b/tests/integration/tests.py
@@ -665,25 +665,19 @@ class TestVllmLmcache_g6:
     def test_lmcache_cpu(self):
         with Runner("lmi", "llama3-8b-lmcache-cpu") as r:
             prepare.build_vllm_async_model("llama3-8b-lmcache-cpu")
-            r.launch(env_vars=[
-                "PYTHONHASHSEED=0"
-            ])
+            r.launch(env_vars=["PYTHONHASHSEED=0"])
             client.run("vllm_lmcache llama3-8b-lmcache-cpu".split())
 
     def test_lmcache_local_storage(self):
         with Runner("lmi", "llama3-8b-lmcache-local-storage") as r:
             prepare.build_vllm_async_model("llama3-8b-lmcache-local-storage")
-            r.launch(env_vars=[
-                "PYTHONHASHSEED=0"
-            ])
+            r.launch(env_vars=["PYTHONHASHSEED=0"])
             client.run("vllm_lmcache llama3-8b-lmcache-local-storage".split())
 
     def test_lmcache_s3(self):
         with Runner("lmi", "llama3-8b-lmcache-s3") as r:
             prepare.build_vllm_async_model("llama3-8b-lmcache-s3")
-            r.launch(env_vars=[
-                "PYTHONHASHSEED=0"
-            ])
+            r.launch(env_vars=["PYTHONHASHSEED=0"])
             client.run("vllm_lmcache llama3-8b-lmcache-s3".split())
 
     def test_lmcache_redis(self):
@@ -698,9 +692,7 @@ class TestVllmLmcache_g6:
         try:
             with Runner("lmi", "llama3-8b-lmcache-redis") as r:
                 prepare.build_vllm_async_model("llama3-8b-lmcache-redis")
-                r.launch(env_vars=[
-                    "PYTHONHASHSEED=0"
-                ])
+                r.launch(env_vars=["PYTHONHASHSEED=0"])
                 client.run("vllm_lmcache llama3-8b-lmcache-redis".split())
         finally:
             # Cleanup Redis container
@@ -718,10 +710,17 @@ class TestVllmLmcache_g6:
         with Runner("lmi", "qwen3-8b-no-cache") as r:
             prepare.build_vllm_async_model("qwen3-8b-no-cache")
             r.launch(env_vars=[
-                "PYTHONHASHSEED=0",
-                "OPTION_LMCACHE_AUTO_CONFIG=True"
+                "PYTHONHASHSEED=0", "OPTION_LMCACHE_AUTO_CONFIG=True"
             ])
             client.run("vllm_lmcache qwen3-8b-lmcache-auto".split())
+
+    def test_lmcache_auto_config_larger_model(self):
+        with Runner("lmi", "qwen2.5-72b-no-cache") as r:
+            prepare.build_vllm_async_model("qwen2.5-72b")
+            r.launch(env_vars=[
+                "PYTHONHASHSEED=0", "OPTION_LMCACHE_AUTO_CONFIG=True"
+            ])
+            client.run("vllm_lmcache qwen2.5-72b-lmcache-auto".split())
 
 
 @pytest.mark.vllm
--- a/engines/python/setup/djl_python/tests/test_lmcache_utils.py
+++ b/engines/python/setup/djl_python/tests/test_lmcache_utils.py
@@ -110,7 +110,7 @@ class TestLMCacheUtils(unittest.TestCase):
     def test_fails_only_with_config_file_property(self):
         """Test that auto-config only fails with config file property, not env vars"""
         # Set env vars (should only warn)
-        os.environ["LMCACHE_LOCAL_DISK"] = "/tmp/cache"
+        os.environ["LMCACHE_LOCAL_DISK"] = "/tmp/lmcache"
 
         # But also set property (should fail)
         properties = {
@@ -296,7 +296,7 @@ class TestLMCacheUtils(unittest.TestCase):
         self.assertEqual(os.environ["LMCACHE_MAX_LOCAL_CPU_SIZE"], "50")
         self.assertEqual(os.environ["LMCACHE_MAX_LOCAL_DISK_SIZE"], "200")
         self.assertEqual(os.environ["LMCACHE_LOCAL_DISK"],
-                         "file:///tmp/cache/")
+                         "file:///tmp/lmcache/")
         self.assertEqual(os.environ["LMCACHE_ENABLE_LAZY_MEMORY_ALLOCATOR"],
                          "true")
         self.assertEqual(os.environ["PYTHONHASHSEED"], "0")
--- a/serving/docker/lmi.Dockerfile
+++ b/serving/docker/lmi.Dockerfile
@@ -74,7 +74,7 @@ COPY config.properties /opt/djl/conf/config.properties
 COPY partition /opt/djl/partition
 COPY scripts/telemetry.sh /opt/djl/bin
 
-RUN apt-get update && apt-get install -yq libaio-dev libopenmpi-dev g++ unzip cuda-compat-12-8 numactl \
+RUN apt-get update && apt-get install -yq libaio-dev libopenmpi-dev g++ unzip cuda-compat-12-8 \
     && scripts/install_openssh.sh \
     && scripts/install_python.sh ${python_version} \
     && scripts/install_s5cmd.sh x64 \
--- a/tests/integration/llm/client.py
+++ b/tests/integration/llm/client.py
@@ -284,6 +284,11 @@ vllm_model_spec = {
         "seq_length": [256],
         "tokenizer": "Qwen/Qwen3-8B"
     },
+    "qwen2.5-72b-lmcache-auto": {
+        "batch_size": [1, 4],
+        "seq_length": [256],
+        "tokenizer": "Qwen/Qwen2.5-72B"
+    },
 }
 
 vllm_neo_model_spec = {
--- a/tests/integration/llm/prepare.py
+++ b/tests/integration/llm/prepare.py
@@ -557,51 +557,36 @@ vllm_model_list = {
     },
     "qwen2.5-72b": {
         "option.model_id": "Qwen/Qwen2.5-72B",
-        "option.tensor_parallel_degree": 4,
+        "option.tensor_parallel_degree": 8,
         "option.load_format": "dummy",
         "option.max_new_tokens": 100,
     },
     "qwen2.5-1.5b-lmcache": {
-        "option.model_id":
-        "Qwen/Qwen2.5-1.5B",
-        "option.tensor_parallel_degree":
-        1,
-        "option.load_format":
-        "dummy",
-        "option.max_new_tokens":
-        100,
-        "option.lmcache_config_file":
-        "lmcache_redis.yaml",
+        "option.model_id": "Qwen/Qwen2.5-1.5B",
+        "option.tensor_parallel_degree": 1,
+        "option.load_format": "dummy",
+        "option.max_new_tokens": 100,
+        "option.lmcache_config_file": "lmcache_redis.yaml",
         "option.kv_transfer_config":
         '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}',
         "load_on_devices": 0
     },
     "qwen2.5-7b-lmcache": {
-        "option.model_id":
-        "Qwen/Qwen2.5-7B",
-        "option.tensor_parallel_degree":
-        1,
-        "option.load_format":
-        "dummy",
-        "option.max_new_tokens":
-        100,
-        "option.lmcache_config_file":
-        "lmcache_qwen25_7b.yaml",
+        "option.model_id": "Qwen/Qwen2.5-7B",
+        "option.tensor_parallel_degree": 1,
+        "option.load_format": "dummy",
+        "option.max_new_tokens": 100,
+        "option.lmcache_config_file": "lmcache_qwen25_7b.yaml",
         "option.kv_transfer_config":
         '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}',
         "load_on_devices": 0
     },
     "qwen2.5-72b-lmcache": {
-        "option.model_id":
-        "Qwen/Qwen2.5-72B",
-        "option.tensor_parallel_degree":
-        4,
-        "option.load_format":
-        "dummy",
-        "option.max_new_tokens":
-        100,
-        "option.lmcache_config_file":
-        "lmcache_qwen25_72b.yaml",
+        "option.model_id": "Qwen/Qwen2.5-72B",
+        "option.tensor_parallel_degree": 4,
+        "option.load_format": "dummy",
+        "option.max_new_tokens": 100,
+        "option.lmcache_config_file": "lmcache_qwen25_72b.yaml",
         "option.kv_transfer_config":
         '{"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_both"}',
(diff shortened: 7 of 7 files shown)


wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
```diff
@@ -86,6 +86,13 @@ static void configureLmiModel(ModelInfo<?, ?> modelInfo) throws ModelException {
             return;
         }
 
+        // Configure LMCache auto-config: set maxWorkers to 1 if enabled
```
Lokiiiiii: I assumed this is because our automatic config logic is pre-conditioned on a single model per server. 
But then it occurs to me that a single model can have multiple workers. 

So why do we have this restriction ?

tests/integration/tests.py
```diff
@@ -1174,17 +1174,17 @@ def test_qwen25_1_5b(self):
...
         # Start Redis via Docker
-        redis_proc = subprocess.Popen(
-            ["docker", "run", "-d", "--rm", "-p", "6379:6379", "redis:alpine"],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.DEVNULL)
-        container_id = redis_proc.stdout.read().decode().strip()
-        time.sleep(5)  # Wait for Redis to start
-
-        with Runner("lmi", "qwen2.5-7b-2a") as r:
+        # redis_proc = subprocess.Popen(
```
Lokiiiiii: Why is this being commented out ?

Lokiiiiii:
Can we update usage documentation as well ?

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+def get_total_cpu_memory() -> float:
+    """
+    Get total available CPU memory in GB.
+    
+    Returns:
+        float: Total available CPU memory in GB
+    """
+    try:
+        with open('/proc/meminfo', 'r') as f:
```
Lokiiiiii: I believe the python native way of accessing system resources is through `psutil` ?

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def get_numa_node_memory() -> float:
+    """
+    Get available memory size of a single NUMA node in GB.
+    
+    Returns:
+        float: Available memory size of first NUMA node in GB, or total system memory as fallback
```
Lokiiiiii: How does NUMA awareness help here ? Why not always use system memory ?

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+    except Exception as e:
+        logger.warning(f"NUMA detection failed: {e}")
+
+    # Fallback to total system memory for any failure or non-NUMA systems
+    logger.info("Using total system memory instead of NUMA node memory")
+    return get_total_cpu_memory()
+
+
+def get_total_cpu_memory() -> float:
```
Lokiiiiii: Rename to reflect what it actually returns - `available` CPU memory (not total)

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+        logger.error(f"Failed to get CPU memory: {e}")
+        raise
+
+
+def get_cache_config(properties: Dict[str, str],
+                     model_size_gb: float) -> float:
+    """
+    Calculate cache configuration using appropriate CPU memory based on TP degree.
+    
+    For TP=1: Uses single NUMA node memory to avoid CUDA allocation limits
```
Lokiiiiii: What do you mean by "CUDA allocation limits" ?

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+                            free_gb = free_mb / 1024.0
+                            logger.info(
+                                f"NUMA node 0 available memory: {free_gb:.1f}GB"
+                            )
+                            return free_gb
+
+    except Exception as e:
+        logger.warning(f"NUMA detection failed: {e}")
+
+    # Fallback to total system memory for any failure or non-NUMA systems
```
Lokiiiiii: This fallback logic is confusing. A function named `get_numa_node_memory` should fail on non-NUMA systems. 

Suggest refactoring the fallback logic to the function call.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+def get_disk_space_info() -> float:
+    """
+    Get available disk space in GB for LMCache storage.
+    
+    Returns:
+        float: total_disk_gb
+    """
+    try:
+        stat = shutil.disk_usage('/tmp/lmcache')
```
Lokiiiiii: Consider using a variable that refers to "/tmp/lmcache" instead of a string literal since it is referenced outside this method. Let's avoid hard coding.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+
+        # Load config from model_id or local path
+        config = AutoConfig.from_pretrained(model_id_or_path)
+
+        dtype = getattr(config, 'torch_dtype', 'float32')
+        dtype_str = str(dtype).split('.')[-1] if '.' in str(dtype) else str(
+            dtype)
+
+        with torch.device("meta"):
+            model = AutoModel.from_config(config)
```
Lokiiiiii: Does this allocate a dummy model ? 

I wonder what the runtime latency overhead for this is. Can we simplify this a bit by just measuring the size of the model directory ? It is guaranteed to be a over-estimation and thus safe in my understanding.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    """
+    Apply LMCache auto-configuration using total CPU memory.
+    
+    This function:
+    1. Checks if lmcache_auto_config=true
+    2. Adds kv_transfer_config for LMCache
+    3. Calculates model size
+    4. Determines cache configuration using total available CPU memory
+    5. Creates LMCache config file
+    6. Sets LMCACHE_CONFIG_FILE environment variable
```
Lokiiiiii: is this documentation accurate - 
https://github.com/deepjavalibrary/djl-serving/blob/2d3d62bb49f607319c4f3c91f0efa462cd11cde7/serving/docs/lmcache_performance.md?plain=1#L57

Is it `OPTION_LMCACHE_CONFIG_FILE` or `LMCACHE_CONFIG_FILE` ?

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+
+        logger.info("LMCache auto-configuration completed successfully")
+        logger.info(
+            f"Final configuration: CPU cache={cpu_cache_gb:.1f}GB, Disk cache={disk_cache_gb:.1f}GB"
+        )
+
+    except Exception as e:
+        logger.error(f"LMCache auto-configuration failed: {e}")
+        logger.warning("Continuing without LMCache auto-configuration")
```
Lokiiiiii: Let's ensure deterministic behavior. 
If LMCache configuration fails, let's fast fail with an error message that asks customer to disable auto configuration.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+        cpu_cache_gb = get_cache_config(properties, model_size_gb)
+
+        disk_cache_gb = get_disk_space_info()
+
+        model_dir = properties.get("model_dir", ".")
+        config_file_path = create_lmcache_config_file(cpu_cache_gb,
+                                                      disk_cache_gb, model_dir)
+
+        # Set environment variable for LMCache to find the config
+        os.environ["LMCACHE_CONFIG_FILE"] = os.path.abspath(config_file_path)
```
Lokiiiiii: What happens if this variable is already set by the customer ? We should fast fail

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    config = {
+        'max_local_cpu_size': int(cpu_cache_gb),
+        'max_local_disk_size': int(disk_cache_gb),
+        'local_disk': 'file:///tmp/lmcache/',
+        'extra_config': {
+            'use_odirect': True
+        }
+    }
+
+    config_path = os.path.join(model_dir, "lmcache_auto_config.yaml")
```
Lokiiiiii: What happens if this file already exists ? 

Test Scenario 1 : Automatic model server restart to recover from some failure. File already exists from previous attempts to run model server. 

Test Scenario 2 : Customer has already created a file with the same name. Need to fast fail.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+        total_cpu_memory = get_numa_node_memory()
+    else:
+        # For multi-process (TP>1), use total system memory since it's distributed
+        total_cpu_memory = get_total_cpu_memory()
+
+    total_reserved = max(2 * model_size_gb, 0.2 * total_cpu_memory)
+    total_cache_memory = max(0, total_cpu_memory - total_reserved)
+
+    # Divide total cache by TP degree to get cache size per LMCache worker
+    cpu_cache_gb_per_gpu = total_cache_memory / tp_degree
```
Lokiiiiii, line 70: Let's double check if LMCache expects global size or per worker size. Link to the documentation in our own LMCache usage guide.

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,310 @@
...
+    For TP>1: Uses total system memory since it's distributed across processes
+    
+    Args:
+        properties: Serving properties
+        model_size_gb: Model size in GB
+        
+    Returns:
+        float: cpu_cache_gb_per_gpu
+    """
+    tp_degree = int(properties.get("tensor_parallel_degree", 1))
```
Lokiiiiii, line 61: What about expert parallel deployments and combinations of TP +EP ? If we don't support it, we should fast fail.

engines/python/setup/djl_python/lmcache_utils.py
smouaa: I noticed that the server would fail to start due to OOM errors if we calculate the total available size as total system memory. When the model is loaded with TP = 1 (1 VLLM worker and 1 LMCache instance) and we attempt to allocate 80% of all system memory, LMCache's memory allocator fails. I'm assuming this is because cudaHostAlloc runs into some sort of error trying to allocate such a large amount of memory at once.

engines/python/setup/djl_python/lmcache_utils.py
smouaa: LMCache crashes the server if max_local_cpu is too large for a singular LMCache instance with RuntimeError: cudaHostAlloc failed

E.g.) With 1.1TB available system memory

max_local_cpu = 876GB (TP=1, Qwen2.5-7B model) fails but 207GB (TP=4, Qwen2.5-72B model) succeeds

engines/python/setup/djl_python/lmcache_utils.py
smouaa: This also was a confusing point for me--LMCache would fail to find the LMCache config file if I set OPTION_LMCACHE_CONFIG_FILE. Not sure if there's logic in DJL-Serving to parse OPTION_LMCACHE_CONFIG_FILE to LMCACHE_CONFIG_FILE (env variable expected by LMCache). Will look into this later

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
smouaa: Yes, the automatic config logic is based on a single model copy per server. Originally I just set load_on_devices = 0 (which achieves the same effect). I think maxWorkers sets the number of DJL worker threads on the Java side, and each worker thread has its own PyProcess. In Connection.java, CUDA_VISIBLE_DEVICES is set on lines 278-284:

`if (worldSize > 0 && device.isGpu()) {
            String cudaDevices = getVisibleDevices(deviceId, worldSize);
            deviceId = 0; // re-map logic device to 0
            pyEnv.addEnv("CUDA_VISIBLE_DEVICES", cudaDevices);
            logger.info("Set CUDA_VISIBLE_DEVICES={}", cudaDevices);
        }`
        
where `int worldSize = tensorParallelDegree * pipelineParallelDegree`

So if max_workers = 1 with TP = 4 and PP = 1 for a model, worldSize = 4 and CUDA_VISIBLE_DEVICES = 0,1,2,3 if there are >=4 available GPUs. VLLM in the PyProcess then creates a number of workers equal to TP based on CUDA_VISIBLE_DEVICES. 

I can modify it to set load_on_devices instead of maxWorkers if you think that's better!

engines/python/setup/djl_python/lmcache_utils.py
smouaa: For a ~70B model it takes 3-5 seconds for the auto config to be generated with this method. 

By model directory, do you mean what's mounted to /opt/ml/model/ in the container? I also considered that method but from my understanding it wouldn't work if the user provides option.model_id instead of the model files since that gets passed to VLLM which then downloads the model to the HuggingFace cache directory. This would occur after the auto config is generated and after the VLLM engine starts.

engines/python/setup/djl_python/lmcache_utils.py
smouaa: I couldn't find any explicit mention of this in any LMCache docs so I manually tested and verified it. One LMCacheEngine instance is created for each VLLM worker (based on TP). So if TP = 4, there are 4 LMCacheEngine instances, which each allocate `max_local_cpu` amount of memory. 

e.g.) TP = 4, max_local_cpu = 210 would be 840 GB total CPU memory allocated for LMCache.

tests/integration/tests.py
smouaa: I was running some tests with a different backend--right now the integration tests aren't properly configured to run for all backends and requires you to swap in different LMCache config files. I'll update the integration tests to properly run for all backends in the next revision

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Will change

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Currently the auto config recommender doesn't consider EP. Will change it to fail if EP Is enabled

wlm/src/main/java/ai/djl/serving/wlm/LmiUtils.java
Lokiiiiii: This is good. I want to add this understanding in comments and in code so other developers can understand why these restrictions exist. 

I also wants to ensure we have the right input validations in place. If auto config mode is enabled customer should not set max_workers > 1. That would be fast fail scenario.

engines/python/setup/djl_python/lmcache_utils.py
Lokiiiiii: I am a bit confused. Let's deep dive here. 

LMcache should be using system memory (RAM) not cuda memory (VRAM). So why would `cudaHostAlloc` be involved in LMCache memory init ? If there is any documentation to the contrary please share. 

Even if we assumed the problem is due to RAM >> VRAM, why would NUMA awareness solve this issue ? We could always set cache size = min(VRAM, RAM*0.8) which does not require NUMA awareness.

Lokiiiiii: I saw your comment down below about the CudaHostAlloc error message. I suspect this is happening from a memory pinning attempt. Could you paste the entire error message so we can get more clarity ?

Lokiiiiii: Could you look into https://docs.lmcache.ai/api_reference/configurations.html#lazy-memory-allocator-configurations ?

smouaa:
```
(EngineCore_DP0 pid=208) ERROR 12-23 18:42:31 [core.py:843] EngineCore failed to start.

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 834, in run_engine_core
    engine_core = EngineCoreProc(*args, **kwargs)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 610, in __init__
    super().__init__(
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 109, in __init__
    num_gpu_blocks, num_cpu_blocks, kv_cache_config = self._initialize_kv_caches(
                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/engine/core.py", line 251, in _initialize_kv_caches
    self.model_executor.initialize_from_config(kv_cache_configs)
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/executor/abstract.py", line 115, in initialize_from_config
    self.collective_rpc("initialize_from_config", args=(kv_cache_configs,))
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/executor/uniproc_executor.py", line 75, in collective_rpc
    result = run_method(self.driver_worker, method, args, kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/serial_utils.py", line 479, in run_method
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/worker/worker_base.py", line 320, in initialize_from_config
    self.worker.initialize_from_config(kv_cache_config)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/v1/worker/gpu_worker.py", line 394, in initialize_from_config
    ensure_kv_transfer_initialized(self.vllm_config, kv_cache_config)
  File "/usr/local/lib/python3.12/dist-packages/vllm/distributed/kv_transfer/kv_transfer_state.py", line 67, in ensure_kv_transfer_initialized
    _KV_CONNECTOR_AGENT = KVConnectorFactory.create_connector(
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/distributed/kv_transfer/kv_connector/factory.py", line 82, in create_connector
    return connector_cls(config, role, kv_cache_config)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/vllm/distributed/kv_transfer/kv_connector/v1/lmcache_connector.py", line 55, in __init__
    self._lmcache_engine = cls(vllm_config, role, self)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/integration/vllm/vllm_v1_adapter.py", line 713, in __init__
    self.lmcache_engine = _init_lmcache_engine(
                          ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/integration/vllm/vllm_v1_adapter.py", line 616, in _init_lmcache_engine
    engine = LMCacheEngineBuilder.get_or_create(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/cache_engine.py", line 1701, in get_or_create
    engine = LMCacheEngine(
             ^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/cache_engine.py", line 178, in __init__
    self.storage_manager = StorageManager(
                           ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/storage_backend/storage_manager.py", line 233, in __init__
    CreateStorageBackends(
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/storage_backend/__init__.py", line 142, in CreateStorageBackends
    local_cpu_backend = LocalCPUBackend(
                        ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/storage_backend/local_cpu_backend.py", line 68, in __init__
    self.initialize_allocator(config, metadata)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/storage_backend/local_cpu_backend.py", line 420, in initialize_allocator
    return MixedMemoryAllocator(
           ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/memory_management.py", line 1644, in __init__
    self.buffer = _allocate_cpu_memory(size, self.numa_mapping)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/lmcache/v1/memory_management.py", line 350, in _allocate_cpu_memory
    ptr = lmc_ops.alloc_pinned_ptr(size, 0)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: cudaHostAlloc failed: 2
```

engines/python/setup/djl_python/lmcache_utils.py
smouaa: Fixed by configuring LMCache with env variables instead to avoid potential issues with overwriting

serving/docs/lmcache_performance.md
```diff
@@ -60,38 +62,56 @@ OPTION_KV_TRANSFER_CONFIG={"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_bo
...
+
+The auto-configuration feature:
+- Automatically detects available CPU RAM and local storage
+- Calculates cache sizes based on model size and instance resources
+- Utilizes LMCache's LazyMemoryAllocator to reduce startup time
+- Sets up both CPU and disk backends with appropriate sizes
+- Generates the LMCache configuration file automatically
+- Requires `maxWorkers=1` (automatically enforced)
+
+**Note**: Auto-configuration is designed for single model deployments. If serving multiple model copies on the same server, use manual configuration.
```
Lokiiiiii: **Note**: Auto-configuration is designed for serving a single model per container instance. For serving multiple models or model copies use SageMaker's Inference Components which ensures resource isolation between models and model copies.

serving/docs/lmcache_performance.md
```diff
@@ -60,38 +62,56 @@ OPTION_KV_TRANSFER_CONFIG={"kv_connector":"LMCacheConnectorV1", "kv_role":"kv_bo
...
 ```
 
+### Option 2: Automatic Configuration
+
+LMI can provide automatic LMCache configuration for CPU and local storage offloading based on your instance resources and model size:
+
+* `option.lmcache_auto_config=True` in serving.properties
+* `OPTION_LMCACHE_AUTO_CONFIG=True` as an environment variable
+
+The auto-configuration feature:
```
Lokiiiiii, line 98: Only supported for Tensor Parallelism

engines/python/setup/djl_python/lmcache_utils.py
```diff
@@ -0,0 +1,368 @@
...
+        properties: Serving properties
+        model_size_gb: Model size in GB
+        
+    Returns:
+        float: disk_cache_gb_per_gpu
+    """
+    tp_degree = int(properties.get("tensor_parallel_degree", 1))
+
+    try:
+        stat = shutil.disk_usage('/tmp')
```
Lokiiiiii: Let's add a comment, doc string, etc. to make it clear that we are assuming /tmp is a NVMe

PR #3016, opened 2026-04-21
Support embedding models with vLLM backend

### LINKED ISSUE #3015: Support embedding tasks with vLLM backend
## Description

I'd like to bring embedding model support to vLLM backend.

DJL Serving currently only supports generation tasks with vLLM backend. However, vLLM has already provided support for a variety of embedding models: https://docs.vllm.ai/en/latest/models/pooling_models/embed

Hosting embedding models with vLLM can drastically improve throughput with its features like dynamic batching and punica kernels.

### Will this change the current api? How?

The API will stay the same as depicted here: https://github.com/yuancu/djl-serving/blob/master/serving/docs/lmi/user_guides/embedding-user-guide.md

### Who will benefit from this enhancement?

Embedding model users who want to adopt vLLM for high throughput, especially those that want to make use of multi-LoRA inference with punica kernels.

## References
- list reference and related literature
- list known implementations

yuancu:
## Description

This PR adds support of embedding models (e.g. `intfloat/e5-small`) to vLLM backend through the vLLM handler

Resolves #3015 

## Type of change

- [x] This change requires a documentation update

## Checklist:
- [x] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L98); One example would be `pytest tests.py -k "TestVllm1" -m "vllm"`
- [x] Have you made corresponding changes to the documentation?

- [x] Unit tests (27 tests in `test_vllm_embedding.py`, 1 in `test_properties_manager.py`)
    - Embedding output formatter: single/batch responses, error handling (non-200 status, missing data field), high-dimensional vectors
    - Task-to-runner/convert mapping: all task values (embed, feature-extraction, generate, classify, reward, pooling, auto, unknown)
    - Runner/convert values propagated to engine args; passthrough override behavior verified
    - Request preprocessing: single/batch text, normalize flag propagation to `use_activation`, model name handling, empty/missing inputs, invalid type rejection
    - Output contract: DJL flat-array shape, content type validation, batch size matching

- [x] Integration tests (e5-small-vllm, bge-base-vllm)
    - Configs added to `prepare.py` and `client.py` with batch sizes [1, 8]
    - Requires Integration Tests Executor run to validate

### DIFF
--- a/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
+++ b/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
@@ -150,7 +150,9 @@ class VLLMHandler(AdapterFormatterMixin):
             base_model_paths,
         )
 
-        self.is_embedding = self.vllm_properties.task in ("text_embedding", "text-embedding", "feature-extraction")
+        self.is_embedding = self.vllm_properties.task in ("text_embedding",
+                                                          "text-embedding",
+                                                          "feature-extraction")
 
         if self.is_embedding:
             self.embedding_service = ServingEmbedding(
@@ -159,7 +161,9 @@ class VLLMHandler(AdapterFormatterMixin):
                 request_logger=None,
             )
             self.normalize_embeddings = self.vllm_properties.normalize
-            logger.info(f"Embedding mode enabled (task={self.vllm_properties.task}, normalize={self.normalize_embeddings})")
+            logger.info(
+                f"Embedding mode enabled (task={self.vllm_properties.task}, normalize={self.normalize_embeddings})"
+            )
         else:
             resolved_chat_template = load_chat_template(
                 self.vllm_properties.chat_template)
--- a/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
+++ b/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
@@ -150,8 +150,7 @@ class VLLMHandler(AdapterFormatterMixin):
             base_model_paths,
         )
 
-        self.is_embedding = self.vllm_properties.task in ("text_embedding",
-                                                          "text-embedding",
+        self.is_embedding = self.vllm_properties.task in ("text-embedding",
                                                           "feature-extraction")
 
         if self.is_embedding:
--- a/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
+++ b/engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
@@ -251,10 +251,15 @@ class VLLMHandler(AdapterFormatterMixin):
             texts = decoded_payload.get("inputs", "")
             if isinstance(texts, str):
                 texts = [texts]
+            if not isinstance(texts, list):
+                raise ValueError(
+                    f"'inputs' must be a string or list of strings, got {type(texts).__name__}"
+                )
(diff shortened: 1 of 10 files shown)


xyang16:
@yuancu Thanks for contributing! Could you run `./gradlew formatPython` to format code? Thanks.

yuancu:
@xyang16 Thanks for pointing out! Fixed `:)`

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
```diff
@@ -192,7 +214,7 @@ def _get_custom_formatter(self,
                 return True
         return self.output_formatter is not None
 
-    def preprocess_request(self, inputs: Input) -> ProcessedRequest:
+    async def preprocess_request(self, inputs: Input) -> ProcessedRequest:
```
xyang16: Why is `async` needed? There's no `await` in the function.

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
```diff
@@ -102,6 +103,38 @@ def validate_task(cls, task):
...
+            },
+            'generate': {
+                'runner': 'generate',
+                'convert': 'auto'
+            },
+            'text_embedding': {
+                'runner': 'auto',
+                'convert': 'embed'
+            },
+            'text-embedding': {
```
xyang16, line 119: Why there are `text-embedding` and `text_embedding`?

engines/python/setup/djl_python/lmi_vllm/vllm_async_service.py
yuancu: Thank you for pointing out! Fixed. It was from a previous experiment where I added awaits in the function. I should have deleted it.

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
yuancu: Because I found that they are used in a mixed way, I added support for both.

For example, in the integration test, `text_embedding` is used

https://github.com/deepjavalibrary/djl-serving/blob/12e9b91e017b4f157127b7d2310e120af2d5bf51/tests/integration/llm/prepare.py#L1378

while in the embedding user guide, `text-embedding` is used

https://github.com/deepjavalibrary/djl-serving/blob/12e9b91e017b4f157127b7d2310e120af2d5bf51/serving/docs/lmi/user_guides/embedding-user-guide.md?plain=1#L90


I guess `text-embedding` should be the canonical one as it aligns with the convention of other task names like `question-answering` and `text-generation`. Should I adopt only `text-embedding` and modify the integration test to use it as well?

xyang16 (approved):

engines/python/setup/djl_python/properties_manager/vllm_rb_properties.py
xyang16: Yes, `text-embedding` should be the canonical one. Otherwise lgtm. Thanks!

yuancu: Thank you! I have removed `text_embedding` as a task name.

Issue #3066, opened 2026-08-05
Any way to get hidden states out of rolling_batch=vllm?


ramkrishs:
# Feature request draft: djl-serving

Target repo: `deepjavalibrary/djl-serving`

---

## Title

Any way to get hidden states out of `rolling_batch=vllm`?

## The problem

We serve a small fine-tuned Llama model (3B, LoRA) with `option.rolling_batch=vllm`, and on top
of the normal generated text we want to return a confidence score for each response. We compute it
with a tiny linear probe (a few thousand parameters) run on the last hidden state of the final
generated token — basically free, since that hidden state already exists as part of the forward
pass, we just need a way to read it out.

The only way we've found to do this is to drop `rolling_batch=vllm` entirely and write a custom
`model.py` that calls `transformers.generate(..., output_hidden_states=True)` directly. That works,
but it means giving up vLLM's batching. We benchmarked both on the same GPU (A10G): the custom
handler does about 30 tok/s no matter how many requests come in at once (no batching), while
`rolling_batch=vllm` does 500-1000+ tok/s at batch sizes 16-128. So getting the confidence score
currently costs us somewhere around 15-30x throughput.

We tried the documented pattern for customizing rolling-batch output (`HuggingFaceService` +
`@output_formatter` in `model.py`), hoping we could keep vLLM batching and just add our own
post-processing on the response. But looking through `request_io.py` and `utils.py`, it looks like
hidden states just aren't kept anywhere in that path — `Token`/`Sequence`/`TextGenerationOutput`
only carry token ids, decoded text, and log-probs. So there's nothing for an output formatter to
read, no matter how it's written.

## What we'd want

Some opt-in way to keep the last token's hidden state around when using `rolling_batch=vllm`, so a
custom `@output_formatter` could read it off the result and do something with it (a confidence
probe like ours, or anything else people might want a hidden state for).

Rough shape of what that could look like:

- A `serving.properties` flag, something like `option.output_last_hidden_state=true`, off by
  default since it's extra memory most people won't want.
- When it's on, `TextGenerationOutput`/`Sequence` gets an extra optional field with the hidden
  state for the last token of that sequence (not the whole sequence — just the last token, to keep
  the cost bounded).
- Then `@output_formatter` code can just read it off the result the same way it already reads
  `.finish_reason` or the token list.

## Why we haven't just done this ourselves

We think the real limitation is actually one level down, in vLLM itself — its normal generation
[...]

PR #3067, opened 2026-08-06
Surface vLLM extract_hidden_states via VLLMRollingBatch

ramkrishs:
## Description

Follow-up to #3066. vLLM's `extract_hidden_states` feature already works and
returns per-request hidden states via
`RequestOutput.kv_transfer_params["hidden_states_path"]`, but djl-serving's
`VLLMRollingBatch` never reads that field, so nothing built on top of
`rolling_batch=vllm` (e.g. a custom `@output_formatter`) can access it today.

This PR closes that gap:

- `request_io.py`: adds an optional `hidden_states_path` field to `Sequence`.
- `rolling_batch_vllm_utils.py`: copies `kv_transfer_params.hidden_states_path`
  onto the `Sequence` when present. No-op when absent (the default case).
- `vllm_rolling_batch.py`: lets a caller pass `kv_transfer_params` per request
  (e.g. `include_output_tokens`), routed through `SamplingParams.extra_args`
  since it isn't itself a `SamplingParams` field.

Nothing here changes default behavior — everything is opt-in and a no-op for
existing `rolling_batch=vllm` deployments that don't configure
`extract_hidden_states`.

## Type of change

## Checklist

- [ ] Integration Tests Executor run
- [x] Manually built the docker image (`serving/docker/lmi.Dockerfile`) and
  verified the change (see Testing below)
- [x] Run related tests
- [x] Added tests that prove the change works
- [x] Code commented where non-obvious
- [ ] Documentation update

## Testing

- Two new unit tests in `test_rb_vllm_utils.py` covering `hidden_states_path`
  set/unset, plus two more covering `translate_vllm_params`'s
  `kv_transfer_params` routing. All existing tests in the file still pass.
  7/7 passing locally (`python -m unittest djl_python.tests.test_rb_vllm_utils`).
- Built `lmi.Dockerfile` locally and validated on our own model (Llama-3.2-3B
  + LoRA) on an A10G:
  - Plain `rolling_batch=vllm` baseline: real inference request, HTTP 200.
  - With `speculative_config={"method": "extract_hidden_states", ...}` +
    `kv_transfer_config` (`ExampleHiddenStatesConnector`): connector
    initializes correctly, requests schedule successfully against the
    modified code.
- Benchmarked the real throughput cost of `extract_hidden_states` (chunked
  prefill must be disabled for it) vs. plain `rolling_batch=vllm`, same
  model/GPU, three concurrency levels:

  | Batch size | Baseline tok/s | Extract tok/s | Cost |
  |---|---|---|---|
  | 32  | 1322.2 | 1179.3 | -10.8% |
  | 64  | 2005.6 | 1543.0 | -23.1% |
  | 128 | 2486.2 | 1724.4 | -30.6% |

  Cost scales with concurrency (chunked prefill disabled means large prefill
  batches block decode for in-flight requests) rather than being flat.

## Note on custom output formatters

Earlier I thought custom `@output_formatter` support was broken for
`rolling_batch=vllm` in general. Turns out that's only true for the legacy
sync engine path — I tested the recommended async path directly
(`option.async_mode=true`, which is the default for vLLM since LMI v17) and a
custom formatter there reads `hidden_states_path` correctly with no changes
needed. So no open gap here for how most people actually run this.

### DIFF
--- a/engines/python/setup/djl_python/request_io.py
+++ b/engines/python/setup/djl_python/request_io.py
@@ -85,6 +85,9 @@ class Sequence:
         cumulative_log_prob: cumulative log probability of the sequence.
         finish_reason: finish reason of the sequence.
         stop_reason: stop reason of the sequence.
+        hidden_states_path: path to a file containing extracted hidden states
+            for this sequence (rolling_batch=vllm with extract_hidden_states),
+            or None if hidden state extraction was not requested/configured.
     """
     tokens: List[Token] = field(default_factory=lambda: [])
     top_tokens: Optional[List[List[Token]]] = field(default_factory=lambda: [])
@@ -92,6 +95,7 @@ class Sequence:
     finish_reason: str = None
     _last_token_index: Optional[int] = None
     stop_reason: Optional[str] = None
+    hidden_states_path: Optional[str] = None
     _tokens_iterator: Optional[Iterator] = field(init=False, default=None)
     _top_tokens_iterator: Optional[Iterator] = field(init=False, default=None)
 
--- a/engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
+++ b/engines/python/setup/djl_python/rolling_batch/rolling_batch_vllm_utils.py
@@ -92,11 +92,23 @@ def update_request_cache_with_output(request_cache: OrderedDict,
 
 
 def update_multiple_sequences(request_output, vllm_request_output):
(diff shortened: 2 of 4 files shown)


ramkrishs:
Hi @lanking520, Could you review this PR Thanks !

ramkrishs:
FYI @a-ys

ramkrishs:
Hi @xyang16  Addressed all review comments, Kindly take a look at it.

xyang16 (approved):

Issue #3071, opened 2026-08-19
High p50 latency under concurrent streaming due to ForkJoinPool.commonPool() and missing TCP_NODELAY


ramkrishs:
### Description

While benchmarking DJL Serving against vLLM directly (same model, same GPU, same prompts, `async_mode=true`), I found DJL's median latency was ~5-7x worse than it should be. Traced it to two separate bottlenecks in the response path:

**1. `InferenceRequestHandler.java` — response completion runs on `ForkJoinPool.commonPool()`**

The streaming response path completes via `.whenCompleteAsync(...)` with no explicit executor, so it defaults to the common pool, which is capped at `availableProcessors() - 1` threads. Under concurrent load this pool is shared with every other unrelated `CompletableFuture` in the JVM, so response dispatch queues up behind whatever else is using it.

Fix: give it a dedicated cached thread pool instead of relying on the default.

**2. `ModelServer.java` — no `TCP_NODELAY` on the server socket**

The Netty `ServerBootstrap` never sets `ChannelOption.TCP_NODELAY`. Streaming responses write small chunks per token, and without `TCP_NODELAY` those can sit behind Nagle's algorithm / delayed ACK, adding latency per chunk.

**3. `python_async_engine.py` — `send_responses()` uses `asyncio.Queue` from a non-async thread**

The response sender thread pushes to an `asyncio.Queue` via `asyncio.run_coroutine_threadsafe(...).result()`, which is a full cross-thread round trip into the event loop for every single token. A plain `queue.Queue` does the same job with a blocking `get()`/`put_nowait()` and no event loop involvement.

### Results

Benchmarked with `meta-llama/Llama-3.2-3B-Instruct`, vLLM `async_mode=true`, 128 concurrent requests, generic public prompt set (not internal data, so this should be reproducible):

| | before | after |
|---|---|---|
| p50 latency | 6.993s | 1.026s |
| throughput | 2559 tok/s | 3743 tok/s |
| p99 latency | 8.025s | 8.698s (no change) |

Same pattern held at 8B scale. p50 and throughput improve a lot; p99 doesn't move, but that turned out to be expected — the remaining tail traces back to vLLM's own scheduler (`max_num_batched_tokens` / `max_num_seqs` / FCFS admission) rather than anything in DJL, since a patched build's p99 already beats vLLM's own standalone p99 at the same concurrency. Mentioning it here just so nobody goes looking for a fourth fix in this codebase for that part.

### How to reproduce

[...]

PR #3072, opened 2026-08-19
PERF Optimization - p50/p99 latency fixes

ramkrishs:
## Description ##

Fixes high p50 latency under concurrent streaming (`async_mode=true`). Two bottlenecks in the response path plus one related simplification:

- **`InferenceRequestHandler.java`**: streaming responses completed via `.whenCompleteAsync(...)` with no explicit executor, so they ran on `ForkJoinPool.commonPool()` (capped at `availableProcessors() - 1` threads, shared with every other unrelated `CompletableFuture` in the JVM). Now uses a dedicated cached thread pool for response dispatch.
- **`ModelServer.java`**: the `ServerBootstrap` never set `ChannelOption.TCP_NODELAY`. Streaming writes small chunks per token, and without it those chunks could sit behind Nagle's algorithm / delayed ACK.
- **`python_async_engine.py`**: `send_responses()` pushed to an `asyncio.Queue` from a non-async thread via `asyncio.run_coroutine_threadsafe(...).result()` — a full cross-thread round trip into the event loop per token. Switched to a plain `queue.Queue` (blocking `get()` / `put_nowait()`), removing the event-loop round trip entirely.

None of these are backward incompatible — internal implementation details only, no public API or config surface changes.

Benchmarked with `meta-llama/Llama-3.2-3B-Instruct`, vLLM `async_mode=true`, 128 concurrent requests:

| | before | after |
|---|---|---|
| p50 latency | 6.993s | 1.026s |
| throughput | 2559 tok/s | 3743 tok/s |
| p99 latency | 8.025s | 8.698s (no change) |

p99 doesn't move — traced that separately and it's coming from vLLM's own scheduler (`max_num_batched_tokens` / `max_num_seqs` / FCFS admission) at this concurrency, not from anything in DJL, so out of scope for this PR. Filed as context in #3071.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [x] Have you run related tests? Check [how to set up the test environment here](https://github.com/deepjavalibrary/djl-serving/blob/master/.github/workflows/integration_execute.yml#L98); One example would be `pytest tests.py -k "TestVllm1" -m "vllm"`

Added unit tests for all three changes, each verified to fail against the pre-fix code:

- [x] `serving/src/test/java/ai/djl/serving/http/InferenceRequestHandlerTest.java` — asserts the response executor is not `ForkJoinPool.commonPool()`, supports more concurrent blocking tasks than common-pool parallelism, and uses daemon threads.
- [x] `serving/src/test/java/ai/djl/serving/ModelServerSocketOptionsTest.java` — asserts `TCP_NODELAY` (and the other socket options) are set on the bootstrap's child options.
- [x] `engines/python/setup/djl_python/tests/test_python_async_engine.py` — covers `invoke_handler` (streaming, non-streaming, OOM, generic error, `None` output) and `send_responses`/`output_queue` cross-thread hand-off.

Load-test comparison (fix vs. no-fix, same `async_mode=true` path, 128 concurrent requests, generic public prompt set, not internal data — reproducible):

```
python3 load_test.py --url http://localhost:8002 --concurrency 128 --duration 90 \
    --config-name djl_async_baseline --container djl --batch-size 128 --dtype auto \
    --gpu-mem-util 0.9 --prompts-file prompts_public.jsonl
```

- [x] Before fix (`djl_llama32_3b_async`): p50 6.993s, p99 8.025s, 2559 tok/s
- [x] After fix (`djl_llama32_3b_patched_v2`): p50 1.026s, p99 8.698s, 3743 tok/s

Same relative improvement (p50 2.8x faster) reproduced at 8B scale (`Llama-3.1-8B-Instruct`).

### DIFF
--- a/engines/python/setup/djl_python/python_async_engine.py
+++ b/engines/python/setup/djl_python/python_async_engine.py
@@ -20,7 +20,6 @@ from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 from threading import Thread
 from queue import Queue
-from asyncio.queues import Queue as AsyncQueue
 
 from djl_python.inputs import Input
 from djl_python.outputs import Output
@@ -40,7 +39,17 @@ class PythonAsyncEngine(PythonSyncEngine):
 
     def __init__(self, args, service):
         super().__init__(args, service)
-        self.output_queue = AsyncQueue()
+        # A plain thread-safe queue.Queue, not asyncio.Queue: outputs are
+        # produced on the event-loop thread (invoke_handler) and consumed on a
+        # dedicated OS thread (send_responses). queue.Queue.put_nowait/get are
+        # safe to call across threads with no event-loop involvement, unlike
+        # asyncio.Queue, which requires scheduling every access back onto the
+        # loop via run_coroutine_threadsafe - that added a cross-thread round
+        # trip (and a wakeup of whichever coroutine loses the GIL race for it)
+        # per streamed token, serializing all concurrent requests' token
+        # delivery through the event loop for no reason: the queue itself
+        # doesn't need the loop, only the blocking socket write does.
+        self.output_queue = Queue()
         self.exception_queue = Queue()
         self.loop = None
         # Todo: for async mode we should maybe consider
@@ -93,21 +102,19 @@ class PythonAsyncEngine(PythonSyncEngine):
             async for output in outputs:
                 output.add_property(REQUEST_TRACKING_ID_KEY,
                                     request_tracking_id)
-                await self.output_queue.put(output)
+                self.output_queue.put_nowait(output)
             return
         # Request tracking ID is needed always for async
         # Do this here so that users in custom handlers do not need to worry about it
         outputs.add_property(REQUEST_TRACKING_ID_KEY, request_tracking_id)
         logging.debug(f"putting result of inference to output queue")
-        await self.output_queue.put(outputs)
+        self.output_queue.put_nowait(outputs)
 
     def send_responses(self):
         logging.info("starting send responses thread")
         while True:
-            future = asyncio.run_coroutine_threadsafe(self.output_queue.get(),
-                                                      self.loop)
             logging.debug("waiting for new inference response")
-            output = future.result()
+            output = self.output_queue.get()
             output.send(self.cl_socket)
 
     def run_server(self):
--- a/engines/python/setup/djl_python/__init__.py
+++ b/engines/python/setup/djl_python/__init__.py
@@ -15,3 +15,5 @@ from .inputs import Input
 from .outputs import Output
 from .pair_list import PairList
 from .session_manager import SessionManager, Session
+
+__version__ = "0.36.0"
--- a/engines/python/setup/djl_python/__init__.py
+++ b/engines/python/setup/djl_python/__init__.py
@@ -15,5 +15,3 @@ from .inputs import Input
 from .outputs import Output
 from .pair_list import PairList
 from .session_manager import SessionManager, Session
-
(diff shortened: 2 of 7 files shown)


krickert:
would love to see this merged

xyang16:
@ramkrishs Thanks for contribution! Could you please run `./gradlew formatJava` to format code and take care of the review comments?

ramkrishs:
Thanks @xyang16 for reviewing it, yes i have addressed all the comments and did lint checks and formated the code as per your request.

ramkrishs:
Not sure why windows build failing looks flaky to me @xyang16  Could you take a look at it ?

xyang16:
> Not sure why windows build failing looks flaky to me @xyang16 Could you take a look at it ?

The windows build succeed on master https://github.com/deepjavalibrary/djl-serving/actions/runs/35145362106.

Looks like the cause of windows build failure is GrpcTest.

```
Gradle suite > Gradle test > ai.djl.serving.grpc.GrpcTest > test FAILED
    java.util.concurrent.CompletionException at CompletableFuture.java:315
        Caused by: ai.djl.engine.EngineException at PyProcess.java:231
            Caused by: java.util.concurrent.TimeoutException at CompletableFuture.java:1960
```

GrpcTest runs 2 models: https://github.com/deepjavalibrary/djl-serving/blob/8e24bd755bd573364b1c4bdbb1db09ee3de05c1a/plugins/grpc/src/test/java/ai/djl/serving/grpc/GrpcTest.java#L47-L51

The log shows it succeeds for one model (echo) and the other model (rolling_batch) connect timeout.

Also on windows the socket is tcp socket.

engines/python/setup/djl_python/python_async_engine.py
xyang16:
```suggestion
                Thread(target=partial(catch_all, self.receive_requests), daemon=True),
```

xyang16:
```suggestion
                Thread(target=partial(catch_all, self.send_responses), daemon=True),
```

xyang16: Need to make receive_requests and send_responses as daemon threads so the async worker process can exit cleanly at shutdown. Otherwise non-daemon thread keeps the process alive indefinitely, leaving the worker orphaned on its socket port. So on Windows then collided with a later worker binding the same 19001 port.

xyang16:
@ramkrishs Could you please run `./gradlew formatPython` to format code?

ramkrishs:
> @ramkrishs Could you please run `./gradlew formatPython` to format code?

Thanks yeah , Just did ran the lint check and pushed it.  Moving forward I am adding a pre commit hook to run it every time before I commit.

engines/python/setup/djl_python/python_async_engine.py
```diff
@@ -124,7 +131,8 @@ def catch_all(func):
                     self.exception_queue.put(str(traceback.format_exc()))
 
             threads = [
-                Thread(target=partial(catch_all, self.receive_requests)),
+                Thread(target=partial(catch_all, self.receive_requests),
+                       daemon=True),
                 Thread(target=partial(catch_all, self.send_responses)),
```
xyang16:
```suggestion
                Thread(target=partial(catch_all, self.send_responses),
                       daemon=True),
```

ramkrishs:
@xyang16 Added all changes based on yours and copilot comments. Let me know if we need anything else.

krickert:
@xyang16 this is going to save some people money :) I looked it over, it looks good.

